Issue #7, opened 2016-06-28
process bounds expressions for parameters with all the parameters available


dtarditi:
Bounds expressions need to processed during clang's semantic checking/processing phase in a scope with all the parameters available.  I am about to submit a pull request where they are being processed in a scope that contains only the parameters seen so far.  

This is a little complicated to implement in clang's semantic checking/processing phase.  You have to delay parsing of the bounds expressions because parsing and semantic processing are intermixed.

dtarditi:
This work is complete.

Issue #6, opened 2016-06-28
test tree transformations of ASTs with new checked types and bounds expressions


dtarditi:

dtarditi:
Putting this in Sprint 10 for now - not sure how much work it is.

dtarditi:
There aren't tests for this in clang because it is difficult to test. Instead, it is tested indirectly by using it within clang. We are using the AST rewriting templates enough we can rely on indirect testing too.

Issue #5, opened 2016-06-28
test traversal of ASTs with new checked types and bounds expressions


dtarditi:

dtarditi:
I'm placing this in Sprint 9 - not sure how much work this is, so I may move it out of Sprint 9.

dtarditi:
There aren't tests for this in clang because it is difficult to test.   Instead, it is tested indirectly by using it within clang.  We are using the AST tree traversal templates enough we can rely on indirect testing too.

Issue #4, opened 2016-06-28
test AST printing of clang IR with new checked C types and bounds expressions


dtarditi:

dtarditi:
This work item has been completed.

PR #8, opened 2016-06-28
Extend clang IR with bounds expressions and parse bounds expressions for parameters.

dtarditi:
This change extends the clang IR to represent Checked C bounds expressions and optional bounds expressions for variable declarations.  It also adds support for parsing bounds expressions and modifies parsing of function parameter lists to parse optional bounds expressions.

Bounds expressions are represented in the IR by adding a new abstract class BoundsExpr and subclassing it for count bounds expressions (count(e1) and byte_count(e1)), range bounds expressions (bounds(e1, e2)), and nullary bounds expressions(bounds(none)).  AST printing, serialization, traversal, and tree transformations are extended handle the new expressions.

Bounds expressions are attached to variable declarations by adding an additional member to VarDecls.  Many VarDecls will not have bounds expressions, so this adds extra space overhead to the representation of VarDecls.  We can revisit this later if it becomes an issue.

To test the new bounds expressions, we add parsing of bounds expressions for function parameter lists and attach the parsed bounds expressions to the VarDecls for the parameters.

Bounds expressions for parameters need to be processed in a scope with all the parameters available.  They are currently being processed in a scope that contains the parameters seen so far.  This is a little complicated to implement in clang.  You have to delay parsing of the bounds expressions. I will come back to this after getting basic parsing of bounds expressions working.  I've opened issue #7  to track this.

Testing:
- This passes the current test baseline for this snapshot of clang:
- Wrote new feature tests of parsing of parameters with bounds declarations.  There will be a separate pull request to the Github CheckedC repo for these tests.
- Passes the existing Checked C tests.
```
  Expected Passes    : 8942
  Expected Failures  : 21
  Unsupported Tests  : 206
  Unexpected Failures: 3
```
- We still need to test AST printing, serialization, traversal, and tree
transformations.  I've opened issues #4 , #3, #5, and #6  to track this.


include/clang/AST/Decl.h
```diff
@@ -1120,6 +1129,19 @@ class VarDecl : public DeclaratorDecl, public Redeclarable<VarDecl> {
     return false;
   }
 
+  /// \brief - Return true if this variable has bounds declared for it.
+  bool hasBoundsExpr() const;
+
+  const Expr *getBoundsExpr() const {
```
reubeno: Does this need a header comment too?

include/clang/AST/Decl.h
```diff
@@ -1120,6 +1129,19 @@ class VarDecl : public DeclaratorDecl, public Redeclarable<VarDecl> {
     return false;
   }
 
+  /// \brief - Return true if this variable has bounds declared for it.
+  bool hasBoundsExpr() const;
```
reubeno: What's your thoughts on making these accessors public?  They seem useful to have around.  At minimum, I'd expect them to be at least as accessible as the underlying BoundsExpr pointer itself (protected).

include/clang/AST/Decl.h
```diff
@@ -800,6 +800,15 @@ class VarDecl : public DeclaratorDecl, public Redeclarable<VarDecl> {
...
   mutable InitType Init;
 
+  // TODO: like the Init member above, it wastes space to have a pointer to a
+  // BoundsExpr in every VarDecl when many of them won't have bounds
+  // declarations. We could move the Init member and the BoundsExpr to an
+  // "extension" object that is allocated on demand, at least not increasing
+  // the space usage of every single VarDecl.
+
+  /// \brief The bounds expression for the variable, if any.
+  Expr *BoundsExpr;
```
reubeno: Should this be a BoundsExpr *?  Or are there cases where you'd need to hang some other kind of expression of it?

include/clang/AST/Expr.h
```diff
@@ -4514,6 +4514,120 @@ class GenericSelectionExpr : public Expr {
   friend class ASTStmtReader;
 };
 
+/// \brief Represents a Checked C bounds expression.
+class BoundsExpr : public Expr {
```
reubeno: Should this derive from Expr?  If it's not a C++ evaluatable expression (which yields a value at compile time or runtime), and if only something that can be attached to a variable or field, then could it be a standalone class with no parents?  I see that Expr derives from Stmt; perhaps it could be a peer of Stmt, then, as a toplevel class in the AST class hierarchy.

I ask not just to make sure that we're appropriately representing it, but also to avoid cases where someone accidentally could stick a bounds expression somewhere they ought not to (e.g. the condition expression of an if statement).  Even if the parser wouldn't do that, that wouldn't stop someone from programmatically inserting one there in an AST.

include/clang/AST/Expr.h
```diff
@@ -4514,6 +4514,120 @@ class GenericSelectionExpr : public Expr {
...
+private:
+  SourceLocation StartLoc, RParenLoc;
```
reubeno: I tend to avoid multiple declarators like this, but I'll keep my mouth shut if this fits in with the style of this code base :)

include/clang/AST/Expr.h
```diff
@@ -4514,6 +4514,120 @@ class GenericSelectionExpr : public Expr {
...
+
+public:
+  BoundsExpr(StmtClass StmtClass, SourceLocation StartLoc, SourceLocation RParenLoc)
+    : Expr(StmtClass, QualType(),  VK_RValue, OK_Ordinary, false,
+           false, false, false), StartLoc(StartLoc), RParenLoc(RParenLoc) {
+  }
+
+  explicit BoundsExpr(StmtClass StmtClass, EmptyShell Empty) :
```
reubeno: Pardon my ignorance, but what does this do?

include/clang/AST/Expr.h
```diff
@@ -4514,6 +4514,120 @@ class GenericSelectionExpr : public Expr {
...
+  SourceLocation getStartLoc() { return StartLoc; }
+  void setStartLoc(SourceLocation Loc) { StartLoc = Loc; }
+  SourceLocation getRParenLoc() { return RParenLoc; }
+  void setRParenLoc(SourceLocation Loc) { RParenLoc = Loc; }
+
+  SourceLocation getLocStart() const LLVM_READONLY { return StartLoc; }
+  SourceLocation getLocEnd() const LLVM_READONLY { return RParenLoc; }
+};
+
+/// \brief Represents a Checked C nullary bounds expression.
```
reubeno: Will we ever use this for something other than 'None' bounds?  If not, should we make this less generic and simply call it NoneBoundsExpr?

include/clang/AST/Expr.h
dtarditi: This is used for deserialization of IR, where you need to create an IR node and then fill it in.

include/clang/AST/Expr.h
```diff
@@ -4514,6 +4514,120 @@ class GenericSelectionExpr : public Expr {
...
+    return child_range(child_iterator(), child_iterator());
+  }
+};
+
+/// \brief Represents a Checked C count bounds expression.
+class CountBoundsExpr : public BoundsExpr {
+public:
+  enum Kind {
+    Count = 0,
+    Byte_Count = 1,
```
reubeno: Style nit: I know the existing code is sometimes inconsistent, but the style seems to be--at least in this source file--one of the two forms:

<allcaps_prefix>_NameWithNoUnderscores
NameWithNoUnderscores

And if used, the prefix seems to be derived from the name of the type.  So I'd recommend going with one of these:

CK_ByteCount
ByteCount

include/clang/AST/Expr.h
dtarditi: I might need to add bounds(Any) back for internal IR processing/checking.   I agree that we can specialize this if we don’t end up needing the generality.

include/clang/AST/Expr.h
```diff
@@ -4514,6 +4514,120 @@ class GenericSelectionExpr : public Expr {
...
+  };
+private:
+  Stmt* CountExpr;
+
+public:
+  CountBoundsExpr(Kind Kind, Expr* Count, SourceLocation StartLoc,
```
reubeno: Style nit: I see asterisks placed both ways for pointer types (i.e. "T *foo" and "T* foo").  But this doesn't seem consistent with the other uses in your addition below.

include/clang/AST/Expr.h
```diff
@@ -4514,6 +4514,120 @@ class GenericSelectionExpr : public Expr {
...
+
+  Kind getKind() const { return (Kind)CountBoundsExprBits.Kind; }
+  void setKind(Kind Kind) { CountBoundsExprBits.Kind = Kind; }
+
+  Expr *getCountExpr() const { return cast<Expr>(CountExpr); }
+  void setCountExpr(Expr *E) { CountExpr = E; }
+
+  // Iterators
+  child_range children() {
+    return child_range(&CountExpr, &CountExpr + 1);
```
reubeno: I see that you've reused the SubExprs pattern below in RangeBoundsExpr.  Should we use that in this class too for consistency?  Also, with that pattern, there's a guarantee of a NULL Stmt pointer after the last valid subexpression; is that an invariant we need to maintain here, or is that not required of the child_range returned by children()?

include/clang/AST/Expr.h
dtarditi: I agree, we should be consistent about underscore and make this change.   I’ll need to look at the code a bit to understand if we need the prefixes.

include/clang/AST/Expr.h
```diff
@@ -4514,6 +4514,120 @@ class GenericSelectionExpr : public Expr {
...
+  child_range children() {
+    return child_range(child_iterator(), child_iterator());
+  }
+};
+
+/// \brief Represents a Checked C count bounds expression.
+class CountBoundsExpr : public BoundsExpr {
+public:
+  enum Kind {
+    Count = 0,
```
reubeno: This is an object/element count; if we don't want to give it a more precise name, then maybe a brief comment distinguishing it from ByteCount would be helpful?

include/clang/Parse/Parser.h
```diff
@@ -144,6 +144,25 @@ class Parser : public CodeCompletionHandler {
   mutable IdentifierInfo *Ident_final;
   mutable IdentifierInfo *Ident_override;
 
+  /// Checked C contextual keywords
+
+  /// These keywords are for bounds expressions.  They are contextual to avoid
+  /// collisions with existing identifiers in programs.  Some keywords like count
```
reubeno: Nit: I know this is being picky, but I'd recommend placing keywords like this in quotes or double-quotes when using them in comments; it just makes it easier for the reader to parse fragments like "and any are likely to collide".

include/clang/Parse/Parser.h
```diff
@@ -1611,6 +1630,18 @@ class Parser : public CodeCompletionHandler {
   ExprResult ParseInitializerWithPotentialDesignator();
 
   //===--------------------------------------------------------------------===//
+  // Checked C Expressions
+
+  ExprResult ParseBoundsExpression();
+
+  private:
```
reubeno: Style nit: access specifiers should be at a previous indent.  Ditto below.

lib/AST/Decl.cpp
```diff
@@ -2009,6 +2010,20 @@ VarDecl *VarDecl::getDefinition(ASTContext &C) {
...
+
+bool VarDecl::hasBoundsExpr() const {
+  return BoundsExpr != nullptr;
+}
+
+Expr *VarDecl::getBoundsExpr() {
+  return BoundsExpr;
+}
+
+void VarDecl::setBoundsExpr(Expr *E) {
```
reubeno: I'm not sure if it's a good example to follow, but the code below for setInit() first tries to deallocate any existing stored expression ('BoundsExpr') in this case before storing the pointer.  Do we need to do anything to make sure that the lifetime of 'E' is appropriate for the lifetime of 'this'?

lib/AST/StmtPrinter.cpp
```diff
@@ -1735,6 +1735,46 @@ void StmtPrinter::VisitAtomicExpr(AtomicExpr *Node) {
...
+
+void StmtPrinter::VisitCountBoundsExpr(CountBoundsExpr *Node) {
+  switch (Node->getKind()) {
+    case CountBoundsExpr::Kind::Byte_Count:
+      OS << "byte_count(";
+    case CountBoundsExpr::Kind::Count:
+      OS << "count(";
+    default:
+      llvm_unreachable("count bounds kind not covered by switch");
+      break;
```
reubeno: Are these breaks under llvm_unreachable calls redundant?  If llvm_unreachable is marked as noreturn, then it's probably safe to elide the breaks.

lib/AST/StmtPrinter.cpp
```diff
@@ -1735,6 +1735,46 @@ void StmtPrinter::VisitAtomicExpr(AtomicExpr *Node) {
...
+  }
+  PrintExpr(Node->getCountExpr());
+  OS << ")";
+}
+
+void StmtPrinter::VisitNullaryBoundsExpr(NullaryBoundsExpr *Node) {
+  switch (Node->getKind()) {
+  case NullaryBoundsExpr::Invalid: {
+    OS << "bounds(invalid)";
```
reubeno: Is this actually a valid syntax?  If not, should we be reporting an error instead of printing this out?

lib/AST/StmtPrinter.cpp
```diff
@@ -1735,6 +1735,46 @@ void StmtPrinter::VisitAtomicExpr(AtomicExpr *Node) {
...
+  default:
+    llvm_unreachable("count bounds kind not covered by switch");
+    break;
+  }
+}
+
+void StmtPrinter::VisitRangeBoundsExpr(RangeBoundsExpr *Node) {
+  OS << "bounds(";
+  PrintExpr(Node->getLowerExpr());
+  OS << ",";
```
reubeno: The printer seems to use whitespace after commas -- we should follow that here and make this ", ".

lib/Sema/SemaDecl.cpp
```diff
@@ -10868,6 +10868,30 @@ void Sema::ActOnFinishKNRParamDeclarations(Scope *S, Declarator &D,
   }
 }
 
+/// ActOnBoundsExpr: attach a bounds expression to a parameter declaration.
+void Sema::ActOnBoundsExpr(Decl *D, Expr *BoundsExpr) {
+  if (!D || !BoundsExpr)
+    return;
+
+  VarDecl *Var = cast<VarDecl>(D);
```
reubeno: Does this throw or otherwise fail in a noticeable way if the Decl isn't a VarDecl?  It would be nice to know as soon as that happens rather than end up corrupting memory when we call setBoundsExpr below.

lib/Parse/ParseExpr.cpp
```diff
@@ -2744,6 +2744,79 @@ void Parser::ParseBlockId(SourceLocation CaretLoc) {
...
+  if (Tok.getKind() == tok::identifier) {
+    IdentifierInfo *Ident = Tok.getIdentifierInfo();
+    if (Ident == Ident_bounds || Ident == Ident_byte_count || Ident == Ident_count) {
+      return Ident;
+    }
+  }
+  return nullptr;
+}
+
+ExprResult Parser::ParseBoundsExpression() {
```
reubeno: There's a lot going on in here, and in the future there will be more; some more inline comments might help guide the reader to more quickly parse it.

lib/Parse/ParseExpr.cpp
```diff
@@ -2744,6 +2744,79 @@ void Parser::ParseBlockId(SourceLocation CaretLoc) {
...
+                                           Tok.getLocation());
+    } else {
+      Result = ExprError();
+    }
+  } else if (Ident == Ident_bounds) {
+    bool FoundNullaryOperator = false;
+
+    if (Tok.getKind() == tok::identifier) {
+      IdentifierInfo *NextIdent = Tok.getIdentifierInfo();
+      if (NextIdent == Ident_none) {
```
reubeno: none is a reserved keyword inside bounds expressions, right?  So we don't have to worry about a range-based bounds expression referring to a variable named 'none'?

lib/Parse/ParseExpr.cpp
```diff
@@ -2744,6 +2744,79 @@ void Parser::ParseBlockId(SourceLocation CaretLoc) {
...
+      }
+    }
+
+    if (!FoundNullaryOperator) {
+      ExprResult LowerBound = Actions.CorrectDelayedTyposInExpr(ParseAssignmentExpression());
+      if (LowerBound.isInvalid()) {
+         Result = ExprError();
+      }
+      if (ExpectAndConsume(tok::comma)) {
+          return ExprError();
```
reubeno: Why is it we're elsewhere setting 'Result' to ExprError() and continuing on, but with this case we need to bail immediately?

lib/Parse/ParseExpr.cpp
```diff
@@ -2744,6 +2744,79 @@ void Parser::ParseBlockId(SourceLocation CaretLoc) {
...
+    }
+  }
+  return nullptr;
+}
+
+ExprResult Parser::ParseBoundsExpression() {
+  IdentifierInfo *Ident = getBoundsExpressionStart();
+  if (!Ident) {
+    Diag(Tok, diag::err_expected_bounds_expr);
+    return ExprError();
```
reubeno: Is there a standard in clang for when you should provide more details with the expression error and when you shouldn't?  Below, if we hit a parser issue with the inner expressions (i.e. in ParseAssignmentExpression) will we preserve any inner error message or is our logic here stomping over that?

lib/Parse/ParseExpr.cpp
```diff
@@ -2744,6 +2744,79 @@ void Parser::ParseBlockId(SourceLocation CaretLoc) {
...
+      } else {
+         Result = ExprError();
+      }
+    }
+  } else {
+    Diag(Tok, diag::err_expected_bounds_expr);
+    Result = ExprError();
+  }
+
+  if (ExpectAndConsume(tok::r_paren)) {
```
reubeno: If 'Result' is already an error at this point in the flow of the function, would it be preferable to return it rather than to re-call ExprError()?  (Or perhaps this might only be more valuable in the future if we add more context / explicit messages to the errors?

reubeno:
I finished a first pass over the pull request; please pardon any naive questions that are more about clang and less about this change--but I'm very eager to learn the answers to those questions if you have the time to answer them :)

lib/AST/Decl.cpp
dtarditi: clang is using a region-based allocation scheme for AST nodes.  They are allocated into a region of memory associated with an ASTContext (see include\clang\AST\ASTContext.h).    Memory is allocated into the region using a bump-pointer based allocation scheme.   With this sort of approach, memory can only be freed when the entire region is freed.  Deallocation for an individual allocation doesn't doesn't do anything (see deallocate in ASTContext.h).   This means that code doesn't need to free nodes when references to them are dropped.  

I looked at the setInit code.  From what I can tell, there's memory allocated outside the ASTContext for the "Evaluated" meber that holds the result of compile-time evaluating the expression.  The main thing that setInit is doing in this case os running the destructor for EvaluatedStmt.  The destructor is implicitly generated. It runs the destructor for the Evaluated member in EvaluatedStmt, which looks like it frees some memory.

lib/AST/StmtPrinter.cpp
dtarditi: Yes, llvm_unreachable is marked as noreturn.   However, I checked the LLVM coding convention and it says "Don’t use default labels in fully covered switches over enumerations", so I'm going to delete this code.

lib/AST/StmtPrinter.cpp
dtarditi: This is an AST printer, so we should just print out the AST, not give an error, even if the the kind is Invalid.

include/clang/AST/Decl.h
dtarditi: Added a comment

include/clang/AST/Decl.h
dtarditi: Changed the type to BoundsExpr.

include/clang/AST/Expr.h
dtarditi: There's lots of code that operates over Expr that is also useful for BoundsExpr.  We'd lose the re-use that code if we made BoundsExpr a peer of Expr in the class hierarchy.

include/clang/AST/Expr.h
dtarditi: This is what is done elsewhere in the file.

include/clang/AST/Expr.h
dtarditi: I did look at the code some more and the LLVM coding convention guide.  The LLVM coding convention documente says the prefixes on enum names should not be used if they are nested within a class.

LLVM does not generally seem to be using enum class, likely because that causes problems with C interoperation.  For an enum declaration, the enum tags are injected into the scope at which the enum is declared.  That's why the prefixes are not needed.

include/clang/AST/Expr.h
dtarditi: Fixed.

include/clang/AST/Expr.h
dtarditi: Other classes that have singleton children aren't using the SubExpr pattern, so that is why I used an explicit child here.

Note that with that pattern, I don't think we're getting a guarantee of a NULL Stmt pointer after the last valid expression.  If we have enum {A, B, LAST_ELEM}, LAST_ELEM will typically have the value 2, giving us a 2 element array sufficient to hold A & B.

include/clang/AST/Expr.h
dtarditi: I renamed this to ElementCount.

include/clang/Parse/Parser.h
dtarditi: Done.

include/clang/Parse/Parser.h
dtarditi: Fixed.

lib/AST/StmtPrinter.cpp
dtarditi: Fixed.

lib/Sema/SemaDecl.cpp
dtarditi: This cast asserts in debug builds, but that leaves release builds at risk.  We could use dyn_cast instead for release builds.

lib/Parse/ParseExpr.cpp
dtarditi: I rewrote this function to be clearer.

lib/Parse/ParseExpr.cpp
dtarditi: None is a contextual keyword.   You can't write bounds(none, y).  But you could write bounds(y, none), and none will be treated as a variable name.

lib/Parse/ParseExpr.cpp
dtarditi: I rewrote this code to be clearer.

lib/Parse/ParseExpr.cpp
dtarditi: I don't think there is a standard, but there is a convention.  Generally you should issue an error or diagnostic message before returning an ExprError().   Because this is compiler output, we don't have to worry about the message being lost.  In this method, most of the error outputs are buried in the helper methods: ExpectAndConsume, for example, issues an error message if the expected token does not appear next in the input.

dtarditi:
I've addressed your current round of feedback on the pull request.   I've pushed a new commit to the branch in personal repo that addresses the code changes.

One open issue is whether we want to avoid using the cast operator.  As far as I can tell, it only asserts in debug builds.  We could use the dyn_cast operator instead, which returns null if the cast fail.  I'll look at existing code some more to see what it is doing and whether one still is favored or not.

include/clang/AST/Expr.h
reubeno: Okay; that's probably a good candidate for mentioning along with the other high-level implementation design decisions that you were planning to write up.  I assume other folks will ask the same question in the future, and it might be helpful to anticipate that :)

---
In reply to: [69053660](https://github.com/Microsoft/checkedc-clang/pull/8#discussion_r69053660) [](ancestors = 69053660)

include/clang/AST/Expr.h
reubeno: Yes, you're totally right--apparently I was confused about 2 not being equal to 3 :)

---
In reply to: [69054080](https://github.com/Microsoft/checkedc-clang/pull/8#discussion_r69054080) [](ancestors = 69054080)

lib/AST/StmtPrinter.cpp
reubeno: Hmm--seems like no good answer here.  This is not a straightforward AST dumper--it's a pretty printer.  And for a previous project I used it to round-trip code from .C -> AST -> .C (back into parseable language text); anytime I saw it emit invalid C code I had to apply a workaround (and assumed it was a code defect).  And I see that there is a precedent of using llvm_unreachable in various places in this file, but I agree that might not be so great to invoke for someone who really does manually construct an AST with an invalid node like this.

Ultimately, I'll defer to your judgment on this.

---
In reply to: [69020519](https://github.com/Microsoft/checkedc-clang/pull/8#discussion_r69020519) [](ancestors = 69020519)

lib/Parse/ParseDecl.cpp
```diff
@@ -5943,6 +5943,17 @@ void Parser::ParseParameterDeclarationClause(
...
       Decl *Param = Actions.ActOnParamDeclarator(getCurScope(), ParmDeclarator);
+
+      // Parse the bounds expression, if any.
+      if (getLangOpts().CheckedC && Tok.is(tok::colon)) {
+        ConsumeToken();
+        ExprResult Bounds = ParseBoundsExpression();
+        if (Bounds.isInvalid()) {
+          SkipUntil(tok::comma, tok::r_paren, StopAtSemi | StopBeforeMatch);
+          Actions.ActOnInvalidBoundsExpr(Param);
+        } else Actions.ActOnBoundsExpr(Param, cast<BoundsExpr>(Bounds.get()));
```
reubeno: Nit: IMHO, this would be more readable if the body of the else were at least on its own line.

lib/Parse/ParseExpr.cpp
```diff
@@ -2744,6 +2744,81 @@ void Parser::ParseBlockId(SourceLocation CaretLoc) {
...
+  if (Ident == Ident_byte_count || Ident == Ident_count) {
+    // Parse byte_count(e1) or count(e1)
+    Result = Actions.CorrectDelayedTyposInExpr(ParseAssignmentExpression());
+    CountBoundsExpr::Kind CountKind = Ident == Ident_count ?
+      CountBoundsExpr::Kind::ElementCount : CountBoundsExpr::Kind::ByteCount;
+    if (!Result.isInvalid())
+      Result = Actions.ActOnCountBoundsExpr(BoundsKWLoc, CountKind,
+                                            Result.get(),
+                                           Tok.getLocation());
+    else Result = ExprError();
```
reubeno: See previous nit about else bodies (and ditto for other uses in this function).  Also feel free to tell me to stop bothering you about this sort of thing if this conforms to patterns in this code base...

reubeno:
Changes look good to me; I made a couple more style comments, but apart from those I'd say we're good to go with this :)

lib/Parse/ParseDecl.cpp
dtarditi: Agreed.  I looked at the code and else statements have on the same line either nothing, {, or an if-control statement:
```
else
   ...
```
or
```
else { 
  ...
 }
```
or 
```
else if (cond) 
```

lib/Parse/ParseExpr.cpp
dtarditi: Agreed - this should be fixed.

dtarditi:
Thanks!  I updated the formatting and completed the pull request.

Issue #9, opened 2016-06-28
Only allow Checked C extension flag for C in clang


dtarditi:
We should only allow the Checked C extension flag to be used for C in clang.   We should issue an error message for other languages.   We're not currently doing the implementation work to enable all the language features for other C family languages supported by clang (such as C++).  We're also not testing the other C family languages.

dtarditi:
This issue has been addressed by pull request #16.

PR #16, opened 2016-07-01
Only allow the Checked C extension flag for C.

dtarditi:
This change only allows the -fcheckedc-extension flag to be used for C programs
in clang.  The clang driver will reject the use of -fcheckedc-extension for
other C family languages supported by clang, including C++, Objective
C, OpenCL, and CUDA. This addresses issue #9 in the checked-clang Github repo.

We are currently not modifying clang to support these other languages, which
is why need to disallow using the extension with them.

Testing:
-  Add 4 new tests to clang.  They test that use of the extension flag is
rejected for C++, Objective C, CUDA, and OpenCL.
- As recommended by the clang documentation, I placed the tests in with
  other similar tests in the test tree.
- I updated the testing baselines to reflect the new tests.  I also updated
the documentation to reflect the fact that we have Checked C specific tests
in clang. We need to take the new tests into account when updating to new
versions of the clang/LLVM sources.

### DIFF
--- a/docs/checkedc/Test-Baselines.md
+++ b/docs/checkedc/Test-Baselines.md
@@ -1,42 +1,5 @@
 # Test base lines
 
-## Current baseline for the master branch
-
-Here is the current baseline for testing just clang with the x86 target (using the check-clang project)
-
-```
-         Failing Tests (3):
-             Clang :: Index/index-templates.cpp
-             Clang :: Index/usrs.m
-             Clang :: Lexer/eof-conflict-marker.c
-```
-```
-           Expected Passes    : 8945
-           Expected Failures  : 21
-           Unsupported Tests  : 206
-           Unexpected Failures: 3
-```
-
-Here is the current base line for testing LLVM + clang on x86 (check-all):
-```
-         Failing Tests (5):
-             Clang :: Index/index-templates.cpp
-             Clang :: Index/usrs.m
-             Clang :: Lexer/eof-conflict-marker.c
-             LLVM :: MC/AsmParser/macros-gas.s
-             LLVM :: tools/llvm-objdump/malformed-archives.test
-```
-```
-           Expected Passes    : 18787
-           Expected Failures  : 97
-           Unsupported Tests  : 6647
-           Unexpected Failures: 5
-```
-
-
-The current base line for testing LLVM + clang on all targets (check-all) in the master
-branch needs to be updated.
-
 ## Testing baseline for the base line branch
 
 
@@ -89,17 +52,14 @@ Here is the current base line for testing LLVM + clang on all targets (check-all
            Unexpected Failures: 6
 ```
 
-## Delta between the master branch and baseline branch
+## Differences in test results between branches
 
-We have added tests for Checked C to the master branch.  These tests are specific to clang, such as tests of
-compiler internals or the driver.  We currently expect the master branch to pass the following additional tests
-in these configurations.
-
-- For just clang with the x86 target (using the check-clang project), 3 additional `Expected Passes` tests.
-- For testing LLVM + clang on x86 (check-all), 8 additional `Expected Passes` tests
-
+The  master branch in the checkedc-clang repo has additional tests for
+Checked C.  It is expected that the master branch will have more
+`Expected Passes` than the baseline branch.
 
 ## In-progress baseline updates
 
-This section records the test results for an in-progress update to latest sources in the baseline branch.  It is currently empty because
-no update is in-progress.
+This section records the test results for an in-progress update to latest
+sources in the baseline branch.  It is currently empty because no update
+is in-progress.
--- a/docs/checkedc/Test-Baselines.md
+++ b/docs/checkedc/Test-Baselines.md
@@ -1,6 +1,43 @@
 # Test base lines
 
-## Current baseline
+## Current baseline for the master branch
+
+Here is the current baseline for testing just clang with the x86 target (using the check-clang project)
+
+```
+         Failing Tests (3):
+             Clang :: Index/index-templates.cpp
+             Clang :: Index/usrs.m
+             Clang :: Lexer/eof-conflict-marker.c
+```
+```
+           Expected Passes    : 8945
+           Expected Failures  : 21
+           Unsupported Tests  : 206
+           Unexpected Failures: 3
+```
+
+Here is the current base line for testing LLVM + clang on x86 (check-all):
+```
+         Failing Tests (5):
+             Clang :: Index/index-templates.cpp
+             Clang :: Index/usrs.m
+             Clang :: Lexer/eof-conflict-marker.c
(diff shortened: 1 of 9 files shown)


docs/checkedc/Test-Baselines.md
```diff
@@ -52,6 +89,16 @@ Here is the current base line for testing LLVM + clang on all targets (check-all
...
 ```
 
+## Delta between the master branch and baseline branch
+
+We have added tests for Checked C to the master branch.  These tests are specific to clang, such as tests of
+compiler internals or the driver.  We currently expect the master branch to pass the following additional tests
+in these configurations.
+
+- For just clang with the x86 target (using the check-clang project), 3 additional `Expected Passes` tests.
+- For testing LLVM + clang on x86 (check-all), 8 additional `Expected Passes` tests
```
reubeno: I saw your updated instructions for keeping this .md file up to date, but I just get the sense that it's going to get out-of-sync.  (And if it stays in sync, I'm wondering if it's going to be unnecessary developer burden for each change that adds a new test.)

I know it's important to keep in agreement on which failures are expected (because they show up in the base line) and agree we should keep that updated--but do we need to update the counts of tests we've added?  I'm not sure how much we'll get from that...

lib/Frontend/CompilerInvocation.cpp
```diff
@@ -1701,7 +1701,25 @@ static void ParseLangArgs(LangOptions &Opts, ArgList &Args, InputKind IK,
...
+  if (Args.hasArg(OPT_fcheckedc_extension)) {
+    const char *disallowed = nullptr;
+    if (Opts.CUDA)
+      disallowed = "CUDA";
+    else if (Opts.OpenCL)
+      disallowed = "OpenCL";
+    else if (Opts.CPlusPlus)
+      disallowed = "C++";
+    else if (Opts.ObjC1 || Opts.ObjC2)
+      disallowed = "Objective C";
```
reubeno: Are these options exclusive?  If not, should we capture the list of all of the conflicting options rather than a single string?

lib/Frontend/CompilerInvocation.cpp
```diff
@@ -1701,7 +1701,25 @@ static void ParseLangArgs(LangOptions &Opts, ArgList &Args, InputKind IK,
...
+    if (disallowed) {
+      Diags.Report(diag::err_drv_argument_not_allowed_with) <<
```
reubeno: Will this end up being fatal?  Or just a warning?  Or will it just not enable the Checked C Extensions?

lib/Frontend/CompilerInvocation.cpp
```diff
@@ -1701,7 +1701,25 @@ static void ParseLangArgs(LangOptions &Opts, ArgList &Args, InputKind IK,
...
+        "-fcheckedc-extension" << disallowed;
+    } else if (!(Opts.C99 || Opts.C11)) {
+      Diags.Report(diag::err_drv_argument_only_allowed_with) <<
+        "-fcheckedc-extension" << "C";
```
reubeno: Should we be more precise and call out that only C99 and C11 are supported?  I don't remember if clang supports ANSI C under a separate option...

test/Driver/checkedc-notsupported.cpp
```diff
@@ -0,0 +1,9 @@
+// Checked C extension is not supported for C++.   Make sure driver
```
reubeno: It's probably low priority, but we could also add a test case for a .c file that is forced to be compiled as C++ using command-line options.  That's a pattern I've certainly seen before.

reubeno:
Finished taking a pass over this; my main question is around the overhead to maintain the expected test counts for tests we add.  Apart from that, this looks good.  (And thanks for putting in the effort to add those test cases!)

docs/checkedc/Test-Baselines.md
dtarditi: Agreed.  Updating precise counts will likely become a developer burden.  We only need to track 'Unexpected failures' in the baseline branch.  We don't particularly need to track the counts. I'll update the md files in the pull request to reflect that. 

I did have some concerns that people will add tests incorrectly and not notice they aren't running.   However, this seems pretty unlikely given the way that the LLVM test infrastructure  works.  It will produce an error for tests that are lacking RUN lines.  The only requirement during code reviews to make sure that tests will run is to make sure they are placed within test suite directories.

lib/Frontend/CompilerInvocation.cpp
dtarditi: This is a bug.   I was trying to protect against someone adding a new major language extension by specifically calling out that only C was supported.  However, I misread the code - there is no way to say we "only support this  compiler flag with plain C" in the current system.   

After studying the code more carefully, I realized that clang assumes that ANSI C (C89) is the default and then builds upon that with various flags for C++ and newer versions of the C and C++ standards. We do want to support  to C89.  The fix is to remove the specific checks for C99 and C11.

For the way that clang represents language options, see clang\include\clang\Basic\LangOptions.def.    This is included by LangOptions.h, in the same directory.

The way clang internally represents language options is different from the default language version that is chosen if none is specified.   The driver currently sets the default language version for C to C11 with gnu extensions.

lib/Frontend/CompilerInvocation.cpp
dtarditi: These options are mutually exclusive, except that Objective C code may be compiled side-by-side with C++ code.  CUDA, OpenCL, and Objective C are all treated as different languages by clang (see clang\include\clang\Basic\LangOptions.def for a list of the language options).   I've updated the message for Objective C when compiled side-by-side with C++.  I also switched to using std::string instead of raw C strings, just to clean up the code.

lib/Frontend/CompilerInvocation.cpp
dtarditi: This ends up being fatal.

test/Driver/checkedc-notsupported.cpp
dtarditi: I've added test cases forcing files to be compiled as different file types than the file extension specifies, with the Checked C extension flag enabled:

- I added a new test with a C file type. It is compiled as a regular C file and also as a C++ file..
- I added run lines where several file types are forced to be compiled as C files..
- I added run lines where the cl file type is compiled as several other language types.
- Finally, I also added tests for clang-cl testing the Microsoft command-line options /TP and /TC for compiling files as C++ and C files respectively.

dtarditi:
I've updated the documentation so that we will not maintain expected passing test counts, which could add overhead. Instead we'll just make sure that there are no unexpected test failures.   I addressed your other comments and updated the issue9 branch.  One neat thing: we're now checking that clang-cl understands the -fcheckedc-extension flag when it is passed using the -Xclang option to clang-cl.

PR #17, opened 2016-07-08
Parse bounds expressions in function parameter list scopes.

dtarditi:
This change addresses issue #7 (process bounds expressions for parameters
with all the parameters available).  Currently, bounds expressions for
parameters are being parsed and type checked in the scope of the parameters
seen so far in a parameter list.  They need to be typechecked in a scope with
all the parameters for the parameter list.

To do this, parsing and type checking of bounds expressions for parameters
is deferred until the entire parameter list has been seen.  This is done
in two steps.  During the initial parse, the tokens for a superset of bounds
expressions are parsed and stored in a list.  They are then re-parsed using
ParseBoundsExpression after all parameters have been seen.

This change also improves error messages when a bounds expression is parsed
with misspelled contextual keyword.  For example, int x : boounds(x, x + 5).
Before, the code would simply give up after parsing the misspelled keyword
and the '('.  This led to additional error messages when the code tried to
 match the closing ')'.  This fix is to skip all the tokens up to but not
including the closing ')'.

Testing:
* Passes existing clang test suite.
* Added additional tests for parsing of function parameter lists with bounds
  expressions.  This will be committed separately to the Checked C repo.

### DIFF
--- a/lib/Parse/ParseDecl.cpp
+++ b/lib/Parse/ParseDecl.cpp
@@ -5881,6 +5881,7 @@ void Parser::ParseParameterDeclarationClause(
   // bounds expressions for parameters is done in a scope that includes all
   // the parameters in the parameter list.
   typedef std::pair<ParmVarDecl *, CachedTokens *> BoundsExprInfo;
+  // We are guessing that most functions take 4 or fewer parameters.
   SmallVector<BoundsExprInfo, 4> deferredBoundsExpressions;
   do {
     // FIXME: Issue a diagnostic if we parsed an attribute-specifier-seq
@@ -5961,9 +5962,10 @@ void Parser::ParseParameterDeclarationClause(
         // error message.  This way the error messages from parsing of bounds
         // expressions will be the same or very similar regardless of whether
         // parsing is deferred or not.
+        // FIXME: Can we use a smart pointer for BoundsExprTokens?
         CachedTokens *BoundsExprTokens = new CachedTokens;
         bool ParsingError = !ConsumeAndStoreBoundsExpression(*BoundsExprTokens);
-        deferredBoundsExpressions.push_back(BoundsExprInfo(Param, BoundsExprTokens));
+        deferredBoundsExpressions.emplace_back(Param, BoundsExprTokens);
         if (ParsingError) {
           SkipUntil(tok::comma, tok::r_paren, StopAtSemi | StopBeforeMatch);
         }
@@ -6061,11 +6063,11 @@ void Parser::ParseParameterDeclarationClause(
   } while (TryConsumeToken(tok::comma));
 
   // Now parse the deferred bounds expressions
-  for (auto Pair : deferredBoundsExpressions) {
+  for (const auto &Pair : deferredBoundsExpressions) {
     ParmVarDecl *Param = Pair.first;
     CachedTokens *Tokens = Pair.second;
-    ExprResult Bounds = DeferredParseBoundsExpression(*Tokens);
-    delete Tokens;
+    // DeferredParseBoundsExpression deletes Tokens
+    ExprResult Bounds = DeferredParseBoundsExpression(Tokens);
     if (Bounds.isInvalid())
       Actions.ActOnInvalidBoundsExpr(Param);
     else
--- a/lib/Parse/ParseDecl.cpp
+++ b/lib/Parse/ParseDecl.cpp
@@ -5859,21 +5859,29 @@ void Parser::ParseFunctionDeclaratorIdentifierList(
 ///         parameter-list ',' parameter-declaration
 ///
 ///       parameter-declaration: [C99 6.7.5]
-///         declaration-specifiers declarator
-/// [C++]   declaration-specifiers declarator '=' assignment-expression
-/// [C++11]                                       initializer-clause
-/// [GNU]   declaration-specifiers declarator attributes
-///         declaration-specifiers abstract-declarator[opt]
-/// [C++]   declaration-specifiers abstract-declarator[opt]
-///           '=' assignment-expression
-/// [GNU]   declaration-specifiers abstract-declarator[opt] attributes
-/// [C++11] attribute-specifier-seq parameter-declaration
+///             declaration-specifiers declarator
+/// [Checked C] declaration-specifiers declarator ':' bounds-expression
+/// [C++]       declaration-specifiers declarator '=' assignment-expression
+/// [C++11]                                           initializer-clause
+/// [GNU]       declaration-specifiers declarator attributes
+///             declaration-specifiers abstract-declarator[opt]
+/// [C++]       declaration-specifiers abstract-declarator[opt]
+///               '=' assignment-expression
+/// [GNU]       declaration-specifiers abstract-declarator[opt] attributes
+/// [C++11]     attribute-specifier-seq parameter-declaration
 ///
 void Parser::ParseParameterDeclarationClause(
        Declarator &D,
        ParsedAttributes &FirstArgAttrs,
        SmallVectorImpl<DeclaratorChunk::ParamInfo> &ParamInfo,
        SourceLocation &EllipsisLoc) {
+  // Delay parsing/semantic checking of bounds expressions until after the
+  // parameter list is parsed. For each variable with a bounds declaration,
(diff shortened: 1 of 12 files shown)


awruef:
Looks good to me.

lib/Parse/ParseDecl.cpp
```diff
@@ -5859,21 +5859,29 @@ void Parser::ParseFunctionDeclaratorIdentifierList(
...
        ParsedAttributes &FirstArgAttrs,
        SmallVectorImpl<DeclaratorChunk::ParamInfo> &ParamInfo,
        SourceLocation &EllipsisLoc) {
+  // Delay parsing/semantic checking of bounds expressions until after the
+  // parameter list is parsed. For each variable with a bounds declaration,
+  // keep a list of tokens for the bounds expression.   Parsing/checking
+  // bounds expressions for parameters is done in a scope that includes all
+  // the parameters in the parameter list.
+  typedef std::pair<ParmVarDecl *, CachedTokens *> BoundsExprInfo;
+  SmallVector<BoundsExprInfo, 4> deferredBoundsExpressions;
```
reubeno: >4 [](start = 30, length = 1)

Is there specific thinking behind the selection of 4 here?  It would be helpful to understand if there's something special about that value or if it was a reasonable guess at likely length...

lib/Parse/ParseDecl.cpp
```diff
@@ -5944,14 +5952,21 @@ void Parser::ParseParameterDeclarationClause(
...
         ConsumeToken();
-        ExprResult Bounds = ParseBoundsExpression();
-        if (Bounds.isInvalid()) {
+        // Consume and store tokens until a bounds-like expression has been
+        // read or a parsing error has happened.  Store the tokens even if a
+        // parsing error occurs so that ParseBoundsExpression can generate the
+        // error message.  This way the error messages from parsing of bounds
+        // expressions will be the same or very similar regardless of whether
+        // parsing is deferred or not.
+        CachedTokens *BoundsExprTokens = new CachedTokens;
```
reubeno: >new CachedTokens [](start = 41, length = 16)

Similar code below has a FIXME comment asking if we can use a smart pointer of some kind; can we?  Otherwise, a thrown exception means leaking the memory.

lib/Parse/ParseDecl.cpp
```diff
@@ -5944,14 +5952,21 @@ void Parser::ParseParameterDeclarationClause(
...
+        bool ParsingError = !ConsumeAndStoreBoundsExpression(*BoundsExprTokens);
+        deferredBoundsExpressions.push_back(BoundsExprInfo(Param, BoundsExprTokens));
```
reubeno: >push_back [](start = 34, length = 9)

Does SmallVector implement emplace_back()?  Would it be preferable to use that here?

lib/Parse/ParseDecl.cpp
```diff
@@ -6044,6 +6059,18 @@ void Parser::ParseParameterDeclarationClause(
 
     // If the next token is a comma, consume it and keep reading arguments.
   } while (TryConsumeToken(tok::comma));
+
+  // Now parse the deferred bounds expressions
+  for (auto Pair : deferredBoundsExpressions) {
```
reubeno: >auto [](start = 7, length = 4)

Did you mean to use an 'auto &' instead?  Or better yet, perhaps a 'const auto &'?  That avoids copies and makes sure you're not mutating the source list.

lib/Parse/ParseExpr.cpp
```diff
@@ -2813,13 +2813,71 @@ ExprResult Parser::ParseBoundsExpression() {
...
+  }
+  ConsumeToken();
+
+  Toks.push_back(Tok);
+  if (Tok.getKind() != tok::l_paren) {
+    return false;
+  }
+  ConsumeParen();
+
+  return ConsumeAndStoreUntil(tok::r_paren, Toks, /*StopAtSemi=*/true);
```
reubeno: >ConsumeAndStoreUntil [](start = 9, length = 20)

Does this take care of ignoring nested matched parens?

lib/Parse/ParseDecl.cpp
```diff
@@ -5859,21 +5859,29 @@ void Parser::ParseFunctionDeclaratorIdentifierList(
...
+/// [C++]       declaration-specifiers declarator '=' assignment-expression
+/// [C++11]                                           initializer-clause
+/// [GNU]       declaration-specifiers declarator attributes
+///             declaration-specifiers abstract-declarator[opt]
+/// [C++]       declaration-specifiers abstract-declarator[opt]
+///               '=' assignment-expression
+/// [GNU]       declaration-specifiers abstract-declarator[opt] attributes
+/// [C++11]     attribute-specifier-seq parameter-declaration
 ///
 void Parser::ParseParameterDeclarationClause(
```
reubeno: >ParseParameterDeclarationClause [](start = 13, length = 31)

High-level comment: the changes to this function to support parsing bounds expressions are getting more and more interesting; is there any factoring that can be done to simplify the logic needed inline here (vs. logic in helper classes/structs/functions) and thus also reduce repeated logic across the other functions you'll later be extending to support bounds exprs?  Or are you planning to defer that work until you change those other parse routines?

lib/Parse/ParseExpr.cpp
```diff
@@ -2813,13 +2813,71 @@ ExprResult Parser::ParseBoundsExpression() {
...
+/// Given a list of tokens that have the same shape as a bounds
+/// expression, parse them to create a bounds expression.
+ExprResult Parser::DeferredParseBoundsExpression(CachedTokens &Toks) {
+  Token LastBoundsExprToken = Toks.back();
+  Token BoundsExprEnd;
+  BoundsExprEnd.startToken();
+  BoundsExprEnd.setKind(tok::eof);
+  SourceLocation OrigLoc = LastBoundsExprToken.getEndLoc();
+  BoundsExprEnd.setLocation(OrigLoc);
+  Toks.push_back(BoundsExprEnd);
```
reubeno: >push_back [](start = 7, length = 9)

It's a bit odd that this code is mutating its input.  At the risk of copying data and consuming slightly more memory, I think it would be cleaner to have the Toks parameter be const and use a separate private list here if you can...

lib/Parse/ParseDecl.cpp
dtarditi: It's a guess that most functions take 4 parameters or less.  SmallVector<T,n> doesn't heap-allocate unless you add more than n elements.  I'll add a comment about the magic constant.

reubeno:
Apart from my questions above, changes generally look good.

lib/Parse/ParseDecl.cpp
dtarditi: clang and LLVM use an explicit error code/value style, instead of C++ exceptions.    I couldn't find documentation on which smart pointer class they want to use.  I'll add a FIXME comment like there is in the similar code below.

lib/Parse/ParseDecl.cpp
dtarditi: Yes, SmallVector provides emplace_back().  It does make the code slightly shorter, so I think it is preferable to use.  I think it is a wash performance-wise for classes with only a few fields.

lib/Parse/ParseDecl.cpp
dtarditi: Good catch.  Yes, I did not want any copies.

lib/Parse/ParseExpr.cpp
dtarditi: Yes, it does take care of handling nested matched parens.

lib/Parse/ParseExpr.cpp
dtarditi: Similar methods in clang actually delete the list of tokens passed to them at the end of the function.   I've changed the code to do that and updated the comment.   I didn't like the fact that code is deleting its argument.  That does make sense if the code is mutating the argument, which as you pointed out is a bit odd.  clang focused on fast parsing, so I'm avoiding the extra copy.

lib/Parse/ParseDecl.cpp
dtarditi: I'm going to wait until I've changed the other parse routines.  The code will need to be a little different at each location, so I want to see what can be shared.

Issue #13, opened 2016-06-28
Parse bounds declarations for variable declarations


dtarditi:
Parse in-line bounds declaratons for variable declarations.

dtarditi:
This functionality has been implemented.

PR #18, opened 2016-07-14
Parse bounds declarations for variable declarations.

dtarditi:
This change adds  parsing of variable declarations with bounds declarations (issue #13).
For each declarator in a declaration, the declarator is parsed and then the
optional Checked C bounds declaration  is parsed.  The bounds declaration is
parsed before the optional initializing expression for the declarator.  Because 
the declarator has already been parsed and added to the current scope, the 
bounds expression can be eagerly parsed.

One surprise with clang was that placing declarators for a declaration
on multiple lines caused a parsing error in the initial implementation,
while having all the declarators on one line did not.   I traced this back to
special case code that looks for typographical mistakes 
at line endings by calling MightBeDeclarator and generating an
error if MightBeDeclarator is false.   MightBeDeclarator returns true
for syntactic items that might start a declarator.  It has special
case checks to make sure that an identifier is followed by something
that might also be part of a declarator.  For Checked C, an identifier 
that starts a declarator may be followed by ':' and a bounds expression,
so allow ':' when the  language options include Checked C.

This change also improves error handling during the parsing of bounds
expressions.
- When an error occurs after having parsed an identifier and a left parenthesis,
  always scan for the matching right parenthesis.  The scan for the matching
  right parenthesis was only happening in one specific case, leading to
  hard-to-understand spurious parsing errors.
- Make a best effort to continue if an error occurs while parsing a
  bounds expression of the form bounds '(' e1 ',' e2, ')'.  clang does not
  differentiate during parsing of expressions between semantic errors and
  parsing failures.  It is important to continue parsing so that a semantic
  error does not cause a cascade of parsing errors.

These problems were uncovered during testing of parsing of variable declarations
with bounds expressions.  Specifically, using an incorrect bounds expression
in a variable declaration with an initializer caused a spurious parsing
errors.

Testing:
- Created a new feature test for parsing of declarations with bounds
  (parsing_bounds_var_declarations.c).  This will be committed separately to the
  checkedc repo.
- Passes existing clang base line tests.

### DIFF
--- a/lib/Parse/ParseDecl.cpp
+++ b/lib/Parse/ParseDecl.cpp
@@ -1623,7 +1623,8 @@ bool Parser::MightBeDeclarator(unsigned Context) {
       // and in block scope it's probably a label. Inside a class definition,
       // this is a bit-field.
       return Context == Declarator::MemberContext ||
-             (getLangOpts().CPlusPlus && Context == Declarator::FileContext);
+             (getLangOpts().CPlusPlus && Context == Declarator::FileContext) ||
+             getLangOpts().CheckedC;
 
     case tok::identifier: // Possible virt-specifier.
       return getLangOpts().CPlusPlus11 && isCXX11VirtSpecifier(NextToken());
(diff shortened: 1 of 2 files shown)


lib/Parse/ParseDecl.cpp
```diff
@@ -1623,7 +1623,8 @@ bool Parser::MightBeDeclarator(unsigned Context) {
       // and in block scope it's probably a label. Inside a class definition,
       // this is a bit-field.
       return Context == Declarator::MemberContext ||
-             (getLangOpts().CPlusPlus && Context == Declarator::FileContext);
+             (getLangOpts().CPlusPlus && Context == Declarator::FileContext) ||
+             getLangOpts().CheckedC;
```
reubeno: >CheckedC [](start = 27, length = 8)

For posterity, it might be helpful to update the comment above to explain why there's a special-case here for Checked C.  (I assume it's for the ':' bounds expr syntax for namespace-scope variable declarations.)

reubeno:
Modulo my one comment comment :), looks good to me!

dtarditi:
Good idea - I've added a comment explaining the case for Checked C.

Issue #20, opened 2016-07-19
Represent bounds information in function types


dtarditi:
There needs to be a representation of bounds information in function types.  Currently the bounds information is not being included in function types.   This is problematic because in clang, typedefs are represent as names that map to type object.  Uses of typedef'ed names refer to those function objects.  It is also problematic for typechecking pointers to function types.

One approach is to use DeBruijn indices: number all the arguments and then change bounds expressions  used in types to refer to those arguments.   We would need a way to represent uses of those arguments in bounds expressions.   For now, we could require that two bounds expressions be exactly identical after this transformation.   Later on, we could canonicalize bounds expressions and require that they have the same canonical representation.  Even later on, we might check to see if one bounds expression implies another.

dtarditi:
Completed by pull request #91.

PR #21, opened 2016-07-20
Parse bounds declarations for function return values.

dtarditi:
This extends parsing of function declarators to parse bounds expressions for
function return values.  A return bounds expression is declared by following
a parameter list with ':' bounds-expression.

The return bounds information is propagated to the representation of function
declarations in clang.  It is not propagated yet to function types because
there is no way to represent it in function types.  I have opened issue #20
to track that and added a few TODO's to the code flagging where the
information will need to be added.

Testing:
- Add feature tests for parsing return bounds.  The tests are in 
the file return_bounds.c, which will be committed separately to the checkedc
Github repo.
- Passes existing clang regression test baseline.


include/clang/Sema/DeclSpec.h
```diff
@@ -1287,11 +1287,18 @@ struct DeclaratorChunk {
     /// \brief The end location of the exception specification, if any.
     unsigned ExceptionSpecLocEnd;
 
+    /// The location of the ':' for the return bounds expression in the source
+    unsigned ReturnBoundsColon;
```
reubeno: >ReturnBoundsColon [](start = 13, length = 17)

[Nit] I'd suggest including 'Loc' in the name to make it clear that this is a location.

include/clang/Sema/DeclSpec.h
```diff
@@ -1352,6 +1359,10 @@ struct DeclaratorChunk {
       return SourceLocation::getFromRawEncoding(RParenLoc);
     }
 
+    SourceLocation getReturnBoundsColon() const {
```
reubeno: >getReturnBoundsColon [](start = 19, length = 20)

[Nit] Ditto from above -- suggest including Loc in the name.

lib/Parse/ParseDecl.cpp
```diff
@@ -5741,6 +5747,17 @@ void Parser::ParseFunctionDeclarator(Declarator &D,
     }
   }
 
+  if (getLangOpts().CheckedC && Tok.is(tok::colon)) {
```
reubeno: >if (getLangOpts().CheckedC && Tok.is(tok::colon)) { [](start = 2, length = 51)

I know you called it out in the header comment for this function, but a quick brief comment calling out that this is for the return bounds expression might be nice (similar to the ones in the few code chunks directly above).

reubeno:
I called out a few very low-priority suggestions; otherwise, looks great.

awruef:
Looks good to me.

Issue #19, opened 2016-07-15
Parse bounds declarations for struct members


dtarditi:
Parse in-line bounds declarations for struct members.

dtarditi:
Completed by pull request #23.

PR #23, opened 2016-07-29
Parsing of member bounds declarations.

dtarditi:
This change adds parsing of member bounds declarations (issue #19).  Member bounds
declarations use member bounds expressions, which can only use other
members of the structure.  The changes include:
- IR: no change is needed to represent member bounds expressions. Because
  clang handles C++, it already can represent members directly in
  the IR for expressions.  Bounds expressions are a subclass of expressions.

  We need to attach bounds expressions to member declarations.  We already
  have bounds expressions on function declarations and variable declarations,
  which have the same base class as member declarations.  Instead of having
  a 3rd copy of the code, we factor the common code for representing bounds
  expressions and move it to the common base.

- Parsing: we modify the parsing of member declarators that are followed
  by ':' to parse either a bounds expression or a constant-expression
  that is a bit-field size. We check whether the first token after the
  ':' is an identifier that is contextual keyword for bounds expressions.

  The parsing of bounds expressions is deferred until all members of the
  structure have been seen.  This is so that member bounds expressions may
  refer to any member of the structure.   Constant expressions for bitfield
  sizes continue to be eagerly parsed.

  The decision to just look for a bounds expression keyword can lead
  to several error messages from clang if there is a spelling
  error in a contextual keyword.   Clang interprets the misspelled keyword
  as an implicit function declaration and issues unknown identifier errors
  for uses of members.

  For now, we'll live this because this specific case is a fairly
  low priority to address.  The type for the identifier in the declarator
  hasn't been built yet at the time that we make the parsing decision.
  We could move that up or defer parsing of constant expressions.  Either
  change is largely orthogonal to this change.

- Semantic processing: member bounds expressions are restricted to accessing
only members.  We add a state flag to Sema that changes the lookup of C
identifiers during expression parsing to look up identifiers in the
member namespace instead of the ordinary namespace.  The state flag is
protected by an RAII class.

Testing:
- Wrote new feature tests of parsing member bounds
  declaration.  These will be checked in separately to the Checked C
  repo in parsing\member_bounds.c.

- Passes existing clang test baseline.


include/clang/AST/Decl.h
```diff
@@ -1584,8 +1582,6 @@ class FunctionDecl : public DeclaratorDecl, public DeclContext,
   /// 'enum Y' in 'void f(enum Y {AA} x) {}'.
   ArrayRef<NamedDecl *> DeclsInPrototypeScope;
 
-  BoundsExpr *ReturnBoundsExpr;
```
reubeno: >ReturnBoundsExpr [](start = 14, length = 16)

I'm not sure where it would go, but it might be helpful to have a comment somewhere explain the intended meaning of 'Bounds' for FunctionDecls--that it's intended to describe the return value of the function.

include/clang/AST/Decl.h
```diff
@@ -731,6 +734,25 @@ class DeclaratorDecl : public ValueDecl {
...
+
+  // Checked C bounds information
+
+  /// \brief Return true if this declaration has bounds declared for it.
+  bool hasBoundsExpr() const;
+
+  /// \brief The declared bounds for this declaration. For function
+  /// declarations, this is the return bounds of the function. Null if no
+  /// bounds have been declared.
+  const BoundsExpr *getBoundsExpr() const {
```
reubeno: >getBoundsExpr [](start = 20, length = 13)

Does clang have a general standard for how to indicate that certain members in base classes may not be applicable to certain derived classes?  Would it, say, be appropriate to add an assertion on setBoundsExpr() that it's only being set on an object on which it makes sense?

lib/Sema/SemaDecl.cpp
```diff
@@ -13083,9 +13083,13 @@ ExprResult Sema::VerifyBitField(SourceLocation FieldLoc,
     // Handle incomplete types with specific error.
     if (RequireCompleteType(FieldLoc, FieldTy, diag::err_field_incomplete))
       return ExprError();
-    if (FieldName)
-      return Diag(FieldLoc, diag::err_not_integral_type_bitfield)
-        << FieldName << FieldTy << BitWidth->getSourceRange();
+    if (FieldName) {
+      if (getLangOpts().CheckedC && FieldTy->isPointerType())
+        return Diag(BitWidth->getExprLoc(), diag::err_expected_bounds_expr_for_member);
```
reubeno: >BitWidth->getExprLoc(), diag::err_expected_bounds_expr_for_member [](start = 20, length = 65)

Following the logic below, should we also append 'FieldName', 'FieldTy', and the source range for whatever was after the colon?

lib/Sema/SemaExpr.cpp
```diff
@@ -2109,10 +2116,18 @@ Sema::ActOnIdExpression(Scope *S, CXXScopeSpec &SS,
     return ActOnDependentIdExpression(SS, TemplateKWLoc, NameInfo,
                                       IsAddressOfOperand, TemplateArgs);
 
+  clang::Sema::LookupNameKind lookupKind = LookupOrdinaryName;
+  if (this->IsMemberBoundsExpr) {
+    assert(getLangOpts().CheckedC);
+    lookupKind = LookupMemberName;
+  }
+
+   if (Id.getKind() == UnqualifiedId::IK_ImplicitSelfParam) {
```
reubeno: >  [](start = 2, length = 1)

[Nit] Stray extra space?

lib/Sema/SemaExpr.cpp
```diff
@@ -2861,19 +2880,27 @@ ExprResult Sema::BuildDeclarationNameExpr(
       valueKind = VK_RValue;
       break;
 
-    // Fields and indirect fields that got here must be for
-    // pointer-to-member expressions; we just call them l-values for
-    // internal consistency, because this subexpression doesn't really
-    // exist in the high-level semantics.
+    // For C++, fields and indirect fields that got here must be for
+    // pointer-to-member expressions.  For Checked C, fields that get here
+    //  must be for member bounds expressions.  For C++, we just call them
```
reubeno: >  [](start = 7, length = 1)

[Nit] Stray space.

lib/Parse/ParseDecl.cpp
```diff
@@ -3618,19 +3618,34 @@ void Parser::ParseStructDeclaration(
...
-      if (Res.isInvalid())
-        SkipUntil(tok::semi, StopBeforeMatch);
-      else
-        DeclaratorInfo.BitfieldSize = Res.get();
+      bool IsBoundsExpr = false;
+      if (getLangOpts().CheckedC && Tok.getKind() == tok::identifier) {
+        IdentifierInfo *Ident = Tok.getIdentifierInfo();
+        IsBoundsExpr = Ident == Ident_byte_count || Ident == Ident_count ||
+                       Ident == Ident_bounds;
+      }
```
reubeno: This logic--peeking at the upcoming tokens and seeing if they form the start of a valid bounds expression--is now included inline in a few places in clang.  Would it make sense to consolidate them into a shared helper function, ideally close to the code that parses bounds expressions?  That might help to consolidate all logic that understands how to detect and parse them and increase the likelihood that it's kept consistent.

lib/Parse/ParseDecl.cpp
```diff
@@ -3618,19 +3618,34 @@ void Parser::ParseStructDeclaration(
...
+      if (IsBoundsExpr) {
+         BoundsExprTokens = new CachedTokens;
```
reubeno: >CachedTokens [](start = 32, length = 12)

Where are these supposed to be freed?

lib/Parse/ParseDecl.cpp
```diff
@@ -3703,14 +3728,18 @@ void Parser::ParseStructUnionBody(SourceLocation RecordLoc,
...
+                                CachedTokens *BoundsExprTokens) {
         // Install the declarator into the current TagDecl.
         Decl *Field =
             Actions.ActOnField(getCurScope(), TagDecl,
                                FD.D.getDeclSpec().getSourceRange().getBegin(),
                                FD.D, FD.BitfieldSize);
         FieldDecls.push_back(Field);
         FD.complete(Field);
+        if (BoundsExprTokens != nullptr) {
+          deferredBoundsExpressions.emplace_back(Field, BoundsExprTokens);
```
reubeno: >emplace_back [](start = 36, length = 12)

This is capturing pointers to the bounds expr tokens into the vector that outlasts the lifetime of the invocation to ParseStructDeclaration().  What's the contract around how long these tokens stay alive, who frees them, etc.?  This is precarious enough that I'd at least suggest that we carefully document how it's supposed to work.  If you have alternative ideas that allow a greater decoupling, I'd be all for that too.

lib/Parse/ParseDecl.cpp
```diff
@@ -3763,6 +3792,18 @@ void Parser::ParseStructUnionBody(SourceLocation RecordLoc,
                       RecordLoc, TagDecl, FieldDecls,
                       T.getOpenLocation(), T.getCloseLocation(),
                       attrs.getList());
+  // Parse the deferred bounds expressions
+  for (const auto &Pair : deferredBoundsExpressions) {
+    EnterMemberBoundsExprRAII MemberBoundsContext(Actions);
+    DeclaratorDecl *FieldDecl = dyn_cast<DeclaratorDecl>(Pair.first);
```
reubeno: >dyn_cast [](start = 32, length = 8)

Do we need a (safer) isa check first?  Just to doubly make sure that they really are all DeclaratorDecls?

reubeno:
This generally looks good to me--just had a few low-priority suggestions around the code and a few questions around the lifetime of cached tokens.

include/clang/AST/Decl.h
dtarditi: There are only 3 subclasses of DeclaratorDecl and it makes sense to set the bounds on all of them.

include/clang/AST/Decl.h
dtarditi: There are already comments on the getter functions in the superclass of FunctionDecl.  These will be picked up the automatic doc generation and by IDE tool tips (for IDEs that understand /brief comments).  I've add a comment to the setter function too.

lib/Sema/SemaDecl.cpp
dtarditi: clang tries to avoid printing out redundant information when source information already has that information.  In this case, the error message will point directly at the beginning of the offending bounds expression or bit-width expression.   I don't think we need add more information here.

lib/Parse/ParseDecl.cpp
dtarditi: Good suggestion - I'll consolidate the logic.

lib/Parse/ParseDecl.cpp
dtarditi: This same like a case for unique_ptr :).

lib/Parse/ParseDecl.cpp
dtarditi: I'll update the code to use unique_ptr's instead, removing the ambiguity about when things will be freed.

lib/Parse/ParseDecl.cpp
dtarditi: I looked at the code and I can just change all of this to be statically typed instead.

---
In reply to: [72877071](https://github.com/Microsoft/checkedc-clang/pull/23#discussion_r72877071) [](ancestors = 72877071)

include/clang/AST/Decl.h
dtarditi: I don't there is a standard, but I've seen asserts using dynamic types that guard against the misuse of members for derived classes.

---
In reply to: [72876100](https://github.com/Microsoft/checkedc-clang/pull/23#discussion_r72876100) [](ancestors = 72876100)

Issue #27, opened 2016-08-12
Remove declaring unchecked arrays using the 'unchecked' keyword


dtarditi:
We decided to stick with the existing rule in the specification that all dimensions of a multi-dimensional array must all be checked or unchecked.   There is no need for the ability to declare unchecked arrays using the 'unchecked' keyword, so we should remove this feature.

dtarditi:
This issue is fixed.

PR #32, opened 2016-08-23
Dimensions of multi-dimensional arrays must have consistent checked properties

dtarditi:
The implementation of Checked C in clang allows  dimensions of multi-dimensional arrays to have different checked properties.  A dimension could be declared as unchecked by prefixing it with the `unchecked` keyword.   The Checked C specification says all dimensions of multi-dimensional arrays have to either be all checked or all unchecked.  We thought about changing the spec, but have decided to stick with it because allowing dimensions to have different checked properties is confusing.

This changes removes the ability to use the `unchecked' keyword to declare unchecked dimensions for multi-dimensional arrays (addressing GitHub issue #27).  In addition, it adds error checking that dimensions in a multi-dimensional array are either all checked or unchecked.

There are some corner cases to handle due to typedefs.  In C, an array is an array of arrays.  A typedef can define an array type that is used as a nested array type.   The case where the enclosing array and the nested array differ in their checked property because of a typedef is now an error.  We take some care to generate an error message that explains what is going on. 

In addition, we also have to handle parenthesized types.  The checked property for a multi-dimensional array propagates from the outer array type to inner array types declared as part of the a declaration (int a checked[10][10] declares a checked array of checked arrays).   Testing for this change uncovered that the checked property was not being propagating properly to nested types when a declarator was parenthesized (int (a checked[10])[10]).  This change fixes that.

Testing:
- Wrote new feature tests for testing that array dimensions have consistent checked properties.  This changes will be checked in separately to the Checked C repo in typechecking\checked_arrays.c.
- Wrote new feature tests for testing parenthesized array declarators for checked arrays.
- Code still passes the existing clang baseline tests.

### DIFF
--- a/lib/Sema/SemaType.cpp
+++ b/lib/Sema/SemaType.cpp
@@ -3466,12 +3466,12 @@ static void checkNullabilityConsistency(TypeProcessingState &state,
 }
 
 // Propagate checked property to directly-nested array types.  Stops at 
-// typedefs, non-array types, and array types that cannot be checked right
-// now (such as variably-sized arrays).
+// typedefs, non-array types, and array types that cannot be checked array
+// types (such as variable-length array types).
 //
-// Issue an error message if the propagation stops at a typedef  thatis an 
+// Issue an error message if the propagation stops at a typedef that is an
 // unchecked array type.  Dimensions of multi-dimensional arrays must either 
-// all be checked.
+// all be checked or all be unchecked.
 static QualType makeNestedArrayChecked(Sema &S, QualType T,
                                        SourceLocation Loc) {
   if (isa<ArrayType>(T)) {
--- a/lib/Sema/SemaType.cpp
+++ b/lib/Sema/SemaType.cpp
@@ -3516,6 +3516,10 @@ static QualType makeNestedArrayChecked(Sema &S, QualType T,
     QualType innerType = makeNestedArrayChecked(S, PT->getInnerType(), Loc);
     ASTContext &Context = S.Context;
     return Context.getParenType(innerType);
+  } else {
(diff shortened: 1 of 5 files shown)


lib/Sema/SemaType.cpp
```diff
@@ -3465,49 +3465,59 @@ static void checkNullabilityConsistency(TypeProcessingState &state,
...
     }
+  } else if (const TypedefType *TD = dyn_cast<TypedefType>(T)) {
+    const ArrayType *AT = dyn_cast<ArrayType>(T.getCanonicalType());
+    if (AT != nullptr && !AT->isChecked())
+      S.Diag(Loc, diag::err_checked_array_of_unchecked_array) << TD->getDecl();
+  } else if (const ParenType *PT = dyn_cast<ParenType>(T)) {
+    assert(!T.hasLocalQualifiers());
+    QualType innerType = makeNestedArrayChecked(S, PT->getInnerType(), Loc);
+    ASTContext &Context = S.Context;
+    return Context.getParenType(innerType);
```
reubeno: Naive question: is there any more generic way of figuring out if the type is a trivial wrapper for another one and grabbing the inner type?  If ever a new type "wrapper" were to be created, we'd miss that case here, right?

reubeno:
Apart from my one comment, looks good to me.

lib/Sema/SemaType.cpp
dtarditi: Good point - I'll add a defensive assert here.  We can get the "canonical" type for a type and check that it is not an array type.   There is no generic way to process wrapper types, but we can at least check that we're not seeing a wrapper type whose canonical type is an array type unexpected.

PR #33, opened 2016-08-24
Initial version of the checked-c conversion tool. Current features:

awruef:
- Processes single C files or multiple C files (up to whole programs)
 - Infers "ptr-ness" for local variables used in those programs.
 - Re-write many source files
 - Supports many aspects of the C grammar
 - Contains unit tests

Current weaknesses:

 - Macro support is limited
 - Casting support is limited, needs structural type equality checks
 - No inherent support for standard library functions.
 - Doesn't honor existing checked C types


include/clang/Constraints/Constraints.h
```diff
@@ -0,0 +1,457 @@
+//                     The LLVM Compiler Infrastructure
```
dtarditi: Why does this file need to go into the include/clang subdirectory?

awruef: Constraints.cpp / Constraints.h are an interface that's more general than clang, so I made Constraints a library. I was envisioning a potential future where something else would use the Constraint interface. That future might never come, though. They could move into the folder that has the rest of the source code for the tool.

dtarditi: I think it'd be a good to move them into the folder with the rest of the code.    I'm trying to minimize changes to clang to make moving Checked C changes upstream easier later.

PR #34, opened 2016-08-24
Rewrite tool

awruef:
Initial version of the checked-c conversion tool. Current features:
 - Processes single C files or multiple C files (up to whole programs)
 - Infers "ptr-ness" for local variables used in those programs.
 - Re-write many source files
 - Supports many aspects of the C grammar
 - Contains unit tests

Current weaknesses:

 - Macro support is limited
 - Casting support is limited, needs structural type equality checks
 - No inherent support for standard library functions.
 - Doesn't honor existing checked C types


dtarditi:
Looks good to me.  Could you file an issue to create some documentation?   It can live clang\docs\checked.

awruef:
Added issue at #35

Issue #26, opened 2016-08-12
Type check count bounds expressions


dtarditi:
This should be fairly simple: ensure that the expression has an integral type.

dtarditi:
Changes committed in pull request #36.

PR #36, opened 2016-08-25
Typecheck count bounds expressions.

dtarditi:
This addresses issue #26.

### DIFF
--- a/lib/Sema/SemaExpr.cpp
+++ b/lib/Sema/SemaExpr.cpp
@@ -11935,7 +11935,20 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
                                       CountBoundsExpr::Kind Kind,
                                       Expr *CountExpr,
                                       SourceLocation RParenLoc) {
-  return new (Context) CountBoundsExpr(Kind, CountExpr, BoundsKWLoc, RParenLoc);
+  ExprResult Result = UsualUnaryConversions(CountExpr);
+  if (Result.isInvalid())
+    return ExprError();
+  Expr *PromotedCountExpr = Result.get();
+  QualType ResultType = PromotedCountExpr->getType();
+  if (!ResultType->isIntegerType()) {
+    std::string CountName = (Kind == CountBoundsExpr::Kind::ElementCount)
+      ? "count" : "byte_count";
+    return ExprError(Diag(CountExpr->getLocStart(),
+                          diag::err_typecheck_count_bounds_expr)
+                       << ResultType << CountName);
+  }
+  return new (Context) CountBoundsExpr(Kind, PromotedCountExpr, BoundsKWLoc,
+                                       RParenLoc);
 }
 
 ExprResult Sema::ActOnRangeBoundsExpr(SourceLocation BoundsKWLoc,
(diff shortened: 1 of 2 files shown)


lib/Sema/SemaExpr.cpp
```diff
@@ -11935,7 +11935,20 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
                                       CountBoundsExpr::Kind Kind,
                                       Expr *CountExpr,
                                       SourceLocation RParenLoc) {
-  return new (Context) CountBoundsExpr(Kind, CountExpr, BoundsKWLoc, RParenLoc);
+  ExprResult Result = UsualUnaryConversions(CountExpr);
```
reubeno: >UsualUnaryConversions [](start = 22, length = 21)

Is this a standard canonicalization?  A comment or two in this function explaining the general approach would be helpful.

lib/Sema/SemaExpr.cpp
```diff
@@ -11935,7 +11935,20 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
...
+  if (Result.isInvalid())
+    return ExprError();
+  Expr *PromotedCountExpr = Result.get();
+  QualType ResultType = PromotedCountExpr->getType();
+  if (!ResultType->isIntegerType()) {
```
reubeno: >isIntegerType [](start = 19, length = 13)

Do we also need to confirm it's no larger than a pointer-sized integer (a.k.a. size_t or ptrdiff_t or something like that)?  Also, are requiring it to be signed or unsigned?  I don't recall what the spec says.

lib/Sema/SemaExpr.cpp
```diff
@@ -11935,7 +11935,20 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
...
+    std::string CountName = (Kind == CountBoundsExpr::Kind::ElementCount)
+      ? "count" : "byte_count";
```
reubeno: >"count" : "byte_count" [](start = 8, length = 22)

Should these be standard constants somewhere rather than direct literals?  I'm assuming they'll get used elsewhere, and there should already be constants for their use as keywords under bounds expressions, right?

lib/Sema/SemaExpr.cpp
dtarditi: This function does the "usual C integer conversions".   If If the type is a character, a short integer, a bit field, or an enumeration type, the expression is promoted to an \keyword{int} type if that is large enough to  hold all values of the type or an unsigned int type otherwise.

lib/Sema/SemaExpr.cpp
dtarditi: The spec allows the integer to be signed or unsigned (it has to be an integral type).

Interestingly, the C definition itself is silent on what happens to pointer arithmetic if you add a pointer type and an integer whose range is larger than that of pointer types.  The definition simply requires that the operand have integer type.   In the case of Checked C, we require that it be provable that count expressions not cause an overflow when they are scaled by the size of the pointer referent type.  This implies that the value must fit into a pointer-sized integer.   This doesn't need to part of typechecking.  We still would need to make sure compilers do something reasonable when asked to add a 64-bit integer to a 32-bit pointer type.

lib/Sema/SemaExpr.cpp
dtarditi: I agree - we should avoid embedding direct literals into the code.   Unfortunately, these are context-sensitive keywords defined in a header file for the parser.  That header file isn't pulled into Sema.h and it'd probably be a layering violation to do so.    I've created two error messages instead.  clang has no localization of error messages, but if it did, this would make it easier :).

dtarditi:
Thanks!   I've addressed your code review feedback

reubeno:
Okay, looks good.

Issue #41, opened 2016-08-25
Checked additonal type requirements on expressions with count bounds expressions


dtarditi:
An expression with a count bounds expressions must have a pointer type.  The type cannot be a pointer to void.

dtarditi:
This work is complete.

PR #42, opened 2016-08-31
Check type requirements for bounds declarations.

dtarditi:
This change implements checking of type requirements for bounds declarations for variables, members, and function return values.  It implements the checking described in Chapter 3 and Chapter 5 (interoperation) of the Checked C specification.   This addresses Github issue #41.

The requirements in Chapter 3 include:
- Bounds declaration are not allowed for variables, members, and return values with ptr type.
- They are also not allowed for function types and function pointer types.
-  Count bounds expressions are allowed for  variables or members that have pointer type or array type.  They are also allowed for pointer-typed function return values.   The pointer type must be a non-void pointer type.
- `byte_count `and `bounds` bounds expressions are allowed for variables  and members with pointer or array types.  They are also allowed for pointer-typed function return values..

The interoperation requirements include:
- Local variables with unchecked pointer or array types cannot have bounds declared for them.    Other declarations (global variables, members, and function return values) with unchecked pointer or array types can have bounds declared for them.  This is how bounds-safe interfaces are described.
- Integer-typed variables, members, and return values can have bounds declared for them using `byte_count` or `bounds` expressions

This change also:
- Fixes the implementation of clang type tests for BoundsExpr and its subclasses.  The static classof methods were not implemented.  In addition, the kind information for CountBoundsExpr and NullaryBoundsExpr was not properly padded, so it was overwriting the superclass kind information.
- Renames ActOnBoundsExpr to ActOnBoundsDecl, to more accurately describe what the function is doing.

Testing:
- New tests have been added to typechecking\bounds.c.   This will be committed separately to the GitHub CheckedC repo.   The new tests systematically test the cross product of variable/member/function return bounds declarations, the different kinds of bounds expressions, and the different types of variables.
- Passes existing clang tests.

### DIFF
--- a/lib/Sema/SemaDecl.cpp
+++ b/lib/Sema/SemaDecl.cpp
@@ -10909,6 +10909,23 @@ void Sema::ActOnBoundsDecl(DeclaratorDecl *D, BoundsExpr *Expr,
       return;
   }
 
+  // Check for errors. Order the error messages from broader
+  // problems to more specific problems.   We don't want to suggest
+  // fixes that will not work because there's a more general problem.
+
+  if (Ty->isCheckedPointerPtrType())
+    // ptr types cannot have bounds expressions
+    DiagId = checkReturnBounds ? diag::err_typecheck_ptr_return_with_bounds
+    : diag::err_typecheck_ptr_decl_with_bounds;
+    // Function pointer types cannot have bounds declared for them because
+    // arithmetic on function pointer types makes no sense. Check for
+    // function types too.  C allows function types to be used in some places.
+    // They are implicitly replaced with function pointer types. Guard against
+    // them not having been replaced yet.
+  else if (Ty->isFunctionPointerType() || Ty->isFunctionType())
+    DiagId = checkReturnBounds
+    ? diag::err_typecheck_function_pointer_return_with_bounds
+    : diag::err_typecheck_function_pointer_decl_with_bounds;
   // Do bounds-safe interface checks. Local variables with unchecked pointer
   // or array types cannot have bounds declarations.  This reduces the chance
   // that programmers accidentally declare variables with unchecked types to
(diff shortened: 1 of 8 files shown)


lib/Sema/SemaDecl.cpp
```diff
@@ -10876,15 +10876,121 @@ void Sema::ActOnFinishKNRParamDeclarations(Scope *S, Declarator &D,
...
 
-/// ActOnBoundsExpr: attach a bounds expression to a parameter declaration.
-void Sema::ActOnBoundsExpr(DeclaratorDecl *D, BoundsExpr *Expr) {
+// ActOnBoundsDecl: handle a Checked C bounds declaration for a declarator.
+// Check that the declarator has the right kind of type for the bounds
+// expression and attach the bounds expression to the declarator.
+//
+// If checkReturnBounds is true, check the return bounds declaration
+// for a function declarator.  The typechecking logic is the same, but
+// the error messages need are slightly different.
```
awruef: typo? needed?

lib/Sema/SemaDecl.cpp
```diff
@@ -10876,15 +10876,121 @@ void Sema::ActOnFinishKNRParamDeclarations(Scope *S, Declarator &D,
...
+  // If we are checking the return bounds, get the function return type.
+  // If the type for the declaration wasn't a function type, bail out.
+  if (checkReturnBounds) {
+    assert(Ty->isFunctionType());
+    const FunctionType *FuncTy = Ty->getAs<FunctionType>();
+    if (FuncTy)
+      Ty = FuncTy->getReturnType();
+    else
+      return;
+  }
```
awruef: I think a more LLVM-idiomatic way to do this is by putting the `getAs` call as a predicate on an `if` statement. Although I don't really know what the code around this does, because the style guide says one thing and much of the existing code does something else.

awruef:
Looks alright to me

PR #43, opened 2016-08-31
Initial support for linking constraints

awruef:
- Link constraints for all function declarations seen in the program
 - Constrain fields


tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -287,7 +453,7 @@ ProgramInfo::getVariableHelper(Expr *E,
     r |= getVariableHelper(CO->getLHS(), V, C);
     r |= getVariableHelper(CO->getRHS(), V, C);
     return r;
-  } else {
+  }  else {
```
dtarditi: Typo: extra space?

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -105,6 +175,10 @@ class ProgramInfo {
   VariableMap &getVarMap() { return Variables;  }
 
 private:
+    bool declHelper(clang::Decl *D,
```
dtarditi: Add a comment about what this does?

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -64,6 +126,14 @@ class ProgramInfo {
   // if this returns false, then both U and V must be constrained to wild.
   bool checkStructuralEquality(uint32_t V, uint32_t U);
 
+  // Called when we are done adding constraints and visiting ASTs. 
+  // Links information about global symbols together and adds 
+  // constraints where appropriate.
+  bool link();
+
+  void seeFunctionDecl(clang::FunctionDecl *, clang::ASTContext *);
```
dtarditi: Add comments about these do?

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -336,9 +343,16 @@ int main(int argc, const char **argv) {
   else
     llvm_unreachable("No action");
 
+  if (!Info.link()) {
+    errs() << "Linking failed!\n";
+    return 0;
```
dtarditi: You should probably return an error code.

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -336,9 +343,16 @@ int main(int argc, const char **argv) {
...
+  }
+
   // 2. Solve constraints.
+  outs() << "solving constraints\n";
```
dtarditi: > outs() << "solving constraints\n"; [](start = 1, length = 35)

This should be under some noisy or verbose flag, not default behavior.

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -230,8 +235,10 @@ void emit(Rewriter &R, ASTContext &C, std::set<FileID> &Files) {
             std::error_code EC;
             raw_fd_ostream out(nFile.str(), EC, sys::fs::F_None);
 
-            if (!EC)
+            if (!EC) {
+              outs() << "writing out " << nFile.str() << "\n";
```
dtarditi: Output should be under diagnostic/verbose flag.

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -134,6 +208,10 @@ class ProgramInfo {
   Constraints CS;
   // Is the ProgramInfo persisted? Only tested in asserts. Starts at true.
   bool persisted;
+  // Global symbol information used for mapping
+  // Set of global functions for whom we don't have a body
+  std::set<std::string> ExternFunctions;
```
dtarditi: >std::set<std::string>  [](start = 2, length = 22)

What is the string?

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -31,6 +31,163 @@ bool ProgramInfo::checkStructuralEquality(uint32_t V, uint32_t U) {
...
+    new GlobalFunctionSymbol(fn, PLoc, parameterVars, returnVars);
+
+  if (F->isThisDeclarationADefinition() && F->hasBody()) {
+    // Check if this function has a body and if it appears in our set of 
+    // functions with no body. If it does, remove it. 
+    std::set<std::string>::iterator I = ExternFunctions.find(fn);
+    if (I != ExternFunctions.end())
+      ExternFunctions.erase(I);
+  } else {
+    // If this function has no body, add it to the set of functions that
```
dtarditi: A duplicate declaration could occur after the function has been defined.   You need some other way to check for that.   I think the code:

extern void f(void) {}

extern void f(void);

is valid.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -31,6 +31,163 @@ bool ProgramInfo::checkStructuralEquality(uint32_t V, uint32_t U) {
...
+        funcs.insert(K);
+      else if (GlobalVariableSymbol *V = dyn_cast<GlobalVariableSymbol>(U))
+        vars.insert(V);
+
+    // Then, pairwise-iterate over each of the function symbols found for this 
+    // symbol. What we want to do is for a sequence of constraint variables on
+    // F1,F2, set the constraint variables for F1(v0,vi,vN) and F2(v0,vj,vN) to
+    // be equal to each other, i.e. to enter a series of constraints of the form
+    // vi == vj.
+    //
```
dtarditi: Because this is a unification based approach, do you need to iterate pairwise overall function definitions?    Given N copies of f (f1, ... fn), doesn't it suffice to set the constraints for f1 = f2, f2 = f3 ... and so on?

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -31,6 +31,163 @@ bool ProgramInfo::checkStructuralEquality(uint32_t V, uint32_t U) {
...
+    // To do that, we need to get the constraints for each parmvar decl position
+    // individually and set them equal, pairwise. 
+    for (std::set<GlobalFunctionSymbol*>::iterator I = funcs.begin();
+      I != funcs.end(); ++I) {
+      std::set<GlobalFunctionSymbol*>::iterator J = I;
+      J++;
+      if (J != funcs.end()) {
+        std::set<uint32_t> &rVars1 = (*I)->getReturns();
+        std::set<uint32_t> &rVars2 = (*J)->getReturns();
+        assert(rVars1.size() == rVars2.size());
```
dtarditi: In real-world code bases, there might be mismatches in types.   The tool should probably report a problem and become conservative when it sees them.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -31,6 +31,163 @@ bool ProgramInfo::checkStructuralEquality(uint32_t V, uint32_t U) {
...
+
+        for (std::set<uint32_t>::iterator V1 = rVars1.begin(), V2 = rVars2.begin();
+          V1 != rVars1.end() && V2 != rVars2.end(); ++V1, ++V2)
+          CS.addConstraint(CS.createEq(
+            CS.getOrCreateVar(*V1), CS.getOrCreateVar(*V2)));
+        
+        std::vector<std::set<uint32_t> > &pVars1 = (*I)->getParams();
+        std::vector<std::set<uint32_t> > &pVars2 = (*J)->getParams();
+        assert(pVars1.size() == pVars2.size());
```
dtarditi: >       assert(pVars1.size() == pVars2.size()); [](start = 1, length = 46)

In real-world code bases, there might be mismatches in the number of parameters.   The tool should probably report a problem and become conservative when it sees them.

tools/checked-c-convert/ProgramInfo.cpp
awruef: Yes, they would. I think the comment doesn't communicate the intent of the code clearly. The iteration happens with `I` and `J` where `J` is always one-ahead of `I`, so I think the code is doing what you say. I'll try to come up with a better way to write the comment.

dtarditi: Ah, OK.  Yes, please clarify the comment.

awruef:
Addressed code review feedback.

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -134,6 +212,12 @@ class ProgramInfo {
   Constraints CS;
   // Is the ProgramInfo persisted? Only tested in asserts. Starts at true.
   bool persisted;
+  // Global symbol information used for mapping
+  // Map of global functions for whom we don't have a body, the keys are 
+  // names of external functions, the value is whether the symbol has been
+  // seen before.
+  std::map<std::string, bool> ExternFunctions;
```
dtarditi: This comment seems inaccurate now.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -31,6 +31,183 @@ bool ProgramInfo::checkStructuralEquality(uint32_t V, uint32_t U) {
...
+        if (rVars1.size() == rVars2.size()) {
+
+          for (std::set<uint32_t>::iterator V1 = rVars1.begin(), V2 = rVars2.begin();
+            V1 != rVars1.end() && V2 != rVars2.end(); ++V1, ++V2)
+            CS.addConstraint(CS.createEq(
+              CS.getOrCreateVar(*V1), CS.getOrCreateVar(*V2)));
+        } else {
+          // Nothing makes sense because this means that the types of two 
+          // functions with the same name is different. Constrain 
+          // everything to top.
```
dtarditi: Probably worth creating a helper function for setting everything to top.   It's fine to do that in a separate commit.

dtarditi:
Looks good to me.

PR #44, opened 2016-09-07
Bug fixes and modest extensions

awruef:
Many changes to deal with running on "real" code:

 * In certain situations (macros) two AST nodes can occupy precisely the same source location. This is a problem for the re-writer that maintains a strong mapping between source locations and constraint variables, especially for whole-program analysis. The rewriter now does something conservative where two constraint variables that occupy the same source location implicitly alias.
 * In other situations, it's difficult to know what to do when re-writing structure typedefs or other kinds of typedefs. We now do something conservative where structure typedefs are left alone. 
 * There is now a bunch of defensive programming to keep the rewriting logic from trying to do rash things like rewrite a declaration that starts in one file and ends in another file.


tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -197,8 +217,16 @@ void rewrite(Rewriter &R, std::set<NewTyp *> &toRewrite, SourceManager &S,
       //       then clang will give back an invalid source range for the 
       //       return type source range. For now, check that the source
       //       range is valid. 
+      //       Additionally, a source range can be (mis) identified as 
+      //       spanning multiple files. We don't know how to re-write that,
+      //       so don't.
       SourceRange SR = UD->getReturnTypeSourceRange();
-      if(SR.isValid())
+      if(SR.isValid() && (R.getRangeSize(SR) != -1))
```
dtarditi: Nit: need a space after the 'if'

tools/checked-c-convert/MappingVisitor.cpp
```diff
@@ -28,12 +27,6 @@ bool MappingVisitor::VisitDeclStmt(DeclStmt *S) {
       Stmt *So = NULL;
       Type *T = NULL;
       std::tie<Stmt *, Decl *, Type *>(So, D, T) = PSLtoSDT[PSL];
-      if (So != NULL) {
```
dtarditi: Do you want to keep this code and put it under the Verbose flag?

tools/checked-c-convert/NewTyp.cpp
```diff
@@ -16,51 +16,55 @@ using namespace llvm;
...
-  Constraints::EnvironmentMap env = CS.getVariables();
 
+  assert(!TL.isNull());
+  
   // Strip off function definitions from the type.
   while (!TL.isNull()) {
     QualType T = TL.getType();
     if (T->isFunctionNoProtoType() || T->isFunctionType() ||
-        T->isFunctionProtoType())
+      T->isFunctionProtoType())
```
dtarditi: Nit: unintentional deletion of space?

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -46,6 +48,7 @@ bool ProgramInfo::link() {
       else if (GlobalVariableSymbol *V = dyn_cast<GlobalVariableSymbol>(U))
         vars.insert(V);
 
+    // Then, iterate over each of the function symbols F1=F,F2=F+1 found for this
```
dtarditi: I found the wording of this comment confusing.

tools/checked-c-convert/NewTyp.h
```diff
@@ -84,6 +98,12 @@ class ArrTyp : public NewTyp {
   std::string mkStr() {
     return ReferentTyp->mkStr() + "* ";
   }
+
+  virtual bool anyChanges() { return false || ReferentTyp->anyChanges(); }
```
dtarditi: Why do you have "return false || ..." here?  Can you eliminate the false clause?

tools/checked-c-convert/NewTyp.h
```diff
@@ -94,5 +114,11 @@ class WildTyp : public NewTyp {
...
```
dtarditi: Same question here about why you have "false || ..." here.

tools/checked-c-convert/CheckedCConvert.cpp
dtarditi: Nit: missing space after 'if'

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -197,8 +217,16 @@ void rewrite(Rewriter &R, std::set<NewTyp *> &toRewrite, SourceManager &S,
...
+        R.ReplaceText(SR, N->mkStr());
+    }
+    else if (FieldDecl *FD = dyn_cast<FieldDecl>(D)) {
```
dtarditi: There's a redundant case for FieldDecl immediately after this one; I think the second one should be deleted.

dtarditi:
Generally looks good - I've pointed out some minor nits in the code.  

The PersistentSourceLoc file needs more comments describing what it is doing.   Please clarify that the source location used for comparisons is the original textual location, not the expanded source location.
 
You might want a helper function for validating that a range can be rewritten.   Currently, you are checking that a range is valid and that getRangeSize(...) != -1.   The test that getRangeSize(...) != -1 is idiomatic,  so using a helper function would make the code clearer to read.

awruef:
I think I've addressed the code review feedback.

PR #45, opened 2016-09-08
Type check bounds expression arguments

dtarditi:
This change adds type checking for bounds expressions of the form (bounds(e1, e2).  The type checking rules are:
- The types of e1 and e2 must both be pointer types.
- The types e1 and e2 must both be pointers to objects or incomplete types, and
- The types of e1 and e2 must both be pointers to the same type or one of them must be a pointer to void.

Testing: 
- New tests will be added to typechecking\bounds.c in the Checked C GitHub repo.  These will be committed separately.
- Passes existing clang tests.

### DIFF
--- a/lib/Sema/SemaExpr.cpp
+++ b/lib/Sema/SemaExpr.cpp
@@ -11955,7 +11955,8 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
 
 /// \brief Validate the type of an argument expression to a bounds
 /// expression. If valid, return the unqualified pointee type.  Otherwise
-/// return null.
+/// return null.  Assumes that the usual C conversions for expressions
+/// have been applied already, including array->pointer conversions.
 const Type *Sema::ValidateBoundsExprArgument(Expr *Arg) {
   QualType ArgType = Arg->getType();
   const Type *ArgPointerType = ArgType->getAs<PointerType>();
@@ -11965,18 +11966,27 @@ const Type *Sema::ValidateBoundsExprArgument(Expr *Arg) {
     return nullptr;
   }
   QualType ArgTypePointee = ArgPointerType->getPointeeType();
-  // For the Checked C extension, we don't expect to see of the clang
-  // extensions for qualifiers.  Assert if we do.
+  // We return the unqualified type so that we can compare
+  // types ignoring qualifier information.  For constant, volatile,
+  // and restrict, it is valid to have argument types that differ
+  // in those qualifiers.  However, clang extends qualifiers with
+  // GC qualifiers (for Objective C) and address-space qualifiers (for
+  // OpenCL).  These should not be ignored when comparing types.  For
+  // Checked C, though, we never expect to see either of those qualifiers
+  // because we restrict usage of the extension to only C. For now, we 
+  // can ignore the existence of those additional qualifiers.  Assert if
+  // we do see them.
   Qualifiers ArgTypeQuals = ArgTypePointee.getQualifiers();
   ArgTypeQuals.removeCVRQualifiers();
   assert(!ArgTypeQuals.hasQualifiers() &&
          "unexpected non-CVR qualifiers on type");
 
   if (!ArgTypePointee->isIncompleteOrObjectType()) {
-    Diag(Arg->getLocStart(), diag::err_typecheck_bounds_expr) << ArgType;
-   return nullptr;
+    Diag(Arg->getLocStart(), diag::err_typecheck_bounds_ex pr) << ArgType;
+    return nullptr;
   }
 
+  // Return the canonical unqualified pointee type.
   return ArgTypePointee.getTypePtr();
 }
 
--- a/lib/Sema/SemaExpr.cpp
+++ b/lib/Sema/SemaExpr.cpp
@@ -11939,8 +11939,8 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
   ExprResult Result = UsualUnaryConversions(CountExpr);
   if (Result.isInvalid())
     return ExprError();
-  Expr *PromotedCountExpr = Result.get();
-  QualType ResultType = PromotedCountExpr->getType();
+  CountExpr = Result.get();
+  QualType ResultType = CountExpr->getType();
   if (!ResultType->isIntegerType()) {
     Diag(CountExpr->getLocStart(),
          Kind == CountBoundsExpr::Kind::ElementCount ?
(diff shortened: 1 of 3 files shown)


lib/Sema/SemaExpr.cpp
```diff
@@ -11949,14 +11949,67 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
...
+  Result = UsualUnaryConversions(UpperBound);
+  if (Result.isInvalid())
+    return ExprError();
+  UpperBound = Result.get();
+
+  const Type *LowerBoundPointee = ValidateBoundsExprArgument(LowerBound);
+  const Type *UpperBoundPointee = ValidateBoundsExprArgument(UpperBound);
+
+  if (!LowerBoundPointee || !UpperBoundPointee)
+    return ExprError();
```
awruef: How could this happen? Should there be a Diag here too?

lib/Sema/SemaExpr.cpp
```diff
@@ -11949,14 +11949,67 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
...
 ExprResult Sema::ActOnRangeBoundsExpr(SourceLocation BoundsKWLoc,
-                                            Expr *LowerBound,
-                                            Expr *UpperBound,
-                                            SourceLocation RParenLoc) {
+                                      Expr *LowerBound,
+                                      Expr *UpperBound,
+                                      SourceLocation RParenLoc) {
+  ExprResult Result = UsualUnaryConversions(LowerBound);
+  if (Result.isInvalid())
+     return ExprError();
```
awruef: How could this happen? Should there be a Diag here too?

awruef:
Left two comments, otherwise looks fine to me.

lib/Sema/SemaExpr.cpp
```diff
@@ -11949,14 +11949,67 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
...
+  return new (Context) CountBoundsExpr(Kind, CountExpr, BoundsKWLoc,
                                        RParenLoc);
 }
 
+/// \brief Validate the type of an argument expression to a bounds
+/// expression. If valid, return the unqualified pointee type.  Otherwise
+/// return null.
+const Type *Sema::ValidateBoundsExprArgument(Expr *Arg) {
+  QualType ArgType = Arg->getType();
+  const Type *ArgPointerType = ArgType->getAs<PointerType>();
```
reubeno: >PointerType [](start = 46, length = 11)

Naive qn: does this rule out array types, e.g.: an int[]?

lib/Sema/SemaExpr.cpp
```diff
@@ -11949,14 +11949,67 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
...
+  if (!ArgPointerType) {
+    Diag(Arg->getLocStart(),
+         diag::err_typecheck_pointer_type_expected) << ArgType;
+    return nullptr;
+  }
+  QualType ArgTypePointee = ArgPointerType->getPointeeType();
+  // For the Checked C extension, we don't expect to see of the clang
```
reubeno: >of [](start = 57, length = 2)

[Nit] Missing word in the comment?

lib/Sema/SemaExpr.cpp
```diff
@@ -11949,14 +11949,67 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
...
+  // extensions for qualifiers.  Assert if we do.
+  Qualifiers ArgTypeQuals = ArgTypePointee.getQualifiers();
+  ArgTypeQuals.removeCVRQualifiers();
+  assert(!ArgTypeQuals.hasQualifiers() &&
+         "unexpected non-CVR qualifiers on type");
+
+  if (!ArgTypePointee->isIncompleteOrObjectType()) {
+    Diag(Arg->getLocStart(), diag::err_typecheck_bounds_expr) << ArgType;
+   return nullptr;
```
reubeno: >    [](start = 0, length = 3)

[Nit] Spacing looks to be off by a char.

lib/Sema/SemaExpr.cpp
```diff
@@ -11949,14 +11949,67 @@ ExprResult Sema::ActOnCountBoundsExpr(SourceLocation BoundsKWLoc,
...
+  QualType ArgType = Arg->getType();
+  const Type *ArgPointerType = ArgType->getAs<PointerType>();
+  if (!ArgPointerType) {
+    Diag(Arg->getLocStart(),
+         diag::err_typecheck_pointer_type_expected) << ArgType;
+    return nullptr;
+  }
+  QualType ArgTypePointee = ArgPointerType->getPointeeType();
+  // For the Checked C extension, we don't expect to see of the clang
+  // extensions for qualifiers.  Assert if we do.
```
reubeno: >extensions for qualifiers [](start = 5, length = 25)

Out of curiosity, does clang provide support for some of the Microsoft C/C++ pointer extensions (e.g. __sptr/__uptr, __ptr32/__ptr64, __unaligned)?  If so, do they show up as qualifiers or in some other way?  Will they get in our way?

reubeno:
Ditto.  Left a few questions, otherwise looks good.

lib/Sema/SemaExpr.cpp
dtarditi: For the C path, there appear to a few relatively obscure (but valid) ways in which UsualUnaryConversions can return an invalid ExprResult.   We don't need diagnostics here.  The convention in clang is that the function that returns an invalid ExprResult should emit any necessary diagnostic.

lib/Sema/SemaExpr.cpp
dtarditi: See the comment below.

lib/Sema/SemaExpr.cpp
dtarditi: Will fix this.

lib/Sema/SemaExpr.cpp
dtarditi: They do support them, but they are treated as attributes on types, not modifiers.    I'll add a longer comment explaining what is going on with clang extensions to type qualifiers.

lib/Sema/SemaExpr.cpp
dtarditi: Good question.  The function assumes that array->pointer conversion has already been done, so it does not rule out array types.   I've added a comment to the header for the function noting the assumption.

dtarditi:
I've address the code review feedback and improved commenting. I also answered the questions above.  I'll commit this now.  If you want additional changes, let me know.

PR #46, opened 2016-09-09
No more test baselines

dtarditi:
It turns out that the unexpected test failures we were seeing were caused by LLVM/clang having some tests that depend on Unix line endings.  On Windows, Git can alter the line endings to match the Windows line ending convention, breaking those tests.  The fix recommended by the LLVM/clang project documentation is to set autocrlf to false, matching the behavior of subversion.

This change:
- Updates the enlistment instructions so that line endings for
source files and test files are configured to be Unix line endings
on Windows.
- Removes the test baseline instructions.  We don't need baselines,
as there should be no unexpected failures.
- Adds some information about line endings to the test section
because they can cause unexpected test failures.
- Updates the working on Windows section to caution developers to
terminate files with line feeds on Windows.


docs/checkedc/Setup-and-Build.md
```diff
@@ -24,6 +24,15 @@ virtual memory and become unresponsive when building.  In VS 2015, go to _Debug-
 set the maximum number of parallel project builds to a fraction of the actual number processors
 (for example, number of processor cores on your machine/3).
 
+LLVM/clang have some tests that depend on using Unix line ending conventions
+(line feeds only).  This means that the sources you will be working with
+need to end with line feeds. Visual Studio preserves line endings for files, so
```
awruef: "end with line feeds only"?

awruef:
Looks good to me.

reubeno:
Looks good to me.

PR #47, opened 2016-09-09
Statistics printing and typedef changes

awruef:
* Only de-sugar typedefs if the immediate value is a pointer value, and then, only perform one "step" of de-sugaring. This reduces cases where `wchar_t` gets re-written as `unsigned short` but there is still a more elegant fix waiting to be implemented.
 * Started adding documentation
 * Fixed a bug where we would only ever constrain the top-most constraint variable in a Decl. Added test cases. 
 * Constrain compound assignment of pointers via `+=`.


tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -292,7 +320,8 @@ class RewriteConsumer : public ASTConsumer {
       Info.getDeclStmtForDecl(J, K);
 
       NewTyp *NT = NewTyp::mkTypForConstrainedType(J, K, Info, &Context);
-      rewriteThese.insert(NT);
+      if(NT)
```
dtarditi: Missing a space.

tools/checked-c-convert/Constraints.cpp
```diff
@@ -285,6 +285,16 @@ VarAtom *Constraints::getOrCreateVar(uint32_t v) {
...
 }
 
+VarAtom *Constraints::getVar(uint32_t v) const {
+  VarAtom tv(v);
+  EnvironmentMap::const_iterator I = environment.find(&tv);
+
+  if (I != environment.end())
+    return I->first;
+  else
+    return NULL;
```
dtarditi: Use nullptr instead?

tools/checked-c-convert/NewTyp.cpp
```diff
@@ -8,6 +8,7 @@
 //===----------------------------------------------------------------------===//
 #include "NewTyp.h"
 
+
 using namespace clang;
```
dtarditi: Unnecessary blank line?

dtarditi:
I found it hard to understand the invariant for what NewTyp::mkTypForConstrainedType is computing.    It is rewriting pointer types and rewriting typedefs that are pointer types.  It is leaving other types and other typedefs along.  This takes care of not desugaring typedefs involving other types (such as typedefs to structs or primitive types).  Could you add a comment describing this?   

I think that loop in MkTypForConstrainedType needs to be rewritten to walk over types structurally at a finer-grain and create structurally analogous NewTyp's.  The use of desugaring causes it to lose type qualifiers and parenthesize for example.    Could you file a GitHub issue for this?   

Could you also file a GitHub issue for revisiting the handling of typedefs involving pointers to types.   In the case where all uses of the typedef'ed type can be rewritten to the same Checked C type, we should rewrite the typedef.   There are a few choices when different uses would end up with different type.  We could do nothing (which would be very conservative), create multiple type-defs (better if there are only a few), or desugar occurrences where types differ.

tools/checked-c-convert/NewTyp.cpp
```diff
@@ -53,18 +59,15 @@ NewTyp *NewTyp::mkTypForConstrainedType(Decl *D, DeclStmt *K,
...
   Constraints::EnvironmentMap env = PI.getConstraints().getVariables();
 
-  // What if we don't have anything to do after stripping off function
-  // definitions?
-  assert(!TL.isNull());
-
-  while (!TL.isNull()) {
-    // What should the current type be qualified as? This can be answered by
-    // looking the constraint up for the current variable.
+  // We step through each level of the type. If the type if a pointer type,
```
dtarditi: Typo: if -> is

dtarditi:
LGTM.

awruef:
I opened issues #48 and #49.

Issue #50, opened 2016-09-13
Revamp kinds for bounds expressions


dtarditi:
It would be good to have one kind field on bounds expressions instead of specialized kinds per type of bounds expressions.  This would make it easier to write code that traverses bounds expressions.  Currently, we have to test the type of a bounds expression and then look at the kind field.

dtarditi:
This work is complete.

PR #52, opened 2016-09-14
Revamp kinds for bounds expressions.

dtarditi:
Instead of having separate kind enumerations for different bounds expression
classes, have just one kind enumeration for all bounds expression classes.
This makes it easier to write predicates that check for a specific kind of
bounds expression, without having to do a dynamic type test and cast to the
specific bounds expression type.  This addresses Github issue #50.

The downside of the unified kind enumeration is that a developer can use a
kind that is invalid for a specific bounds expression class.  Add code to
validate the kind information.

Add predicates for testing for specific kinds of bounds expressions and
clean up code for type checking bounds expressions by using the predicates.

This also adds a kind for the `: ptr` interop bounds annotation to the
kind enumeration for bounds expressions.

Testing:

### DIFF
--- a/include/clang/AST/Expr.h
+++ b/include/clang/AST/Expr.h
@@ -4522,9 +4522,16 @@ public:
     ByteCount = 3,
     // bounds(e1, e2)
     Range = 4,
-    // ptr interop annotation
-    PtrInteropAnnotation
+    // ptr interop annotation.  This isn't really a bounds expression.
+    // To save space and for programmng convenience, we store the 
+    // ": ptr" interop annotation as a bounds expression.
+    PtrInteropAnnotation = 5,
+
+    // Sentinel marker for maximum bounds kind.
+    MaxBoundsKind = PtrInteropAnnotation
   };
+
+  static_assert(MaxBoundsKind < (1 << NumBoundsExprKindBits), "kind field too small");
 
   BoundsExpr(StmtClass StmtClass, Kind BoundsKind, SourceLocation StartLoc,
              SourceLocation RParenLoc)
@@ -4548,8 +4555,8 @@ public:
 
   Kind getKind() const { return (Kind)BoundsExprBits.Kind; }
   void setKind(Kind Kind) {
+    assert(validateKind(Kind));
     BoundsExprBits.Kind = Kind;
-    assert(validateKind());
   }
 
   bool isInvalid() {
@@ -4584,7 +4591,7 @@ public:
 private:
   /// \brief: check that the kind is valid for the type
   /// of bounds expression.
-  bool validateKind();
+  bool validateKind(Kind kind);
 };
 
 /// \brief Represents a Checked C nullary bounds expression.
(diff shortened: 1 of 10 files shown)


lib/AST/Expr.cpp
```diff
@@ -3814,6 +3814,20 @@ PseudoObjectExpr::PseudoObjectExpr(QualType type, ExprValueKind VK,
...
+  else if (isa<NullaryBoundsExpr>(this))
+    return K == None || K == PtrInteropAnnotation;
+  else if (isa<CountBoundsExpr>(this))
+    return K == ElementCount || K == ByteCount;
+  else if (isa<RangeBoundsExpr>(this))
+    return K == Range;
+
+  return false;
+}
+
```
awruef: It's not idiomatic LLVM, but could this be replaced with a switch case?

include/clang/AST/Expr.h
```diff
@@ -4525,36 +4546,57 @@ class BoundsExpr : public Expr {
...
-    Invalid = 0,
-    // bounds(none)
-    None = 1
-  };
-
-public:
   NullaryBoundsExpr(Kind Kind, SourceLocation StartLoc, SourceLocation RParenLoc)
-    : BoundsExpr(NullaryBoundsExprClass, StartLoc, RParenLoc)  {
-    NullaryBoundsExprBits.Kind = Kind;
+    : BoundsExpr(NullaryBoundsExprClass, Kind, StartLoc, RParenLoc)  {
```
reubeno: In these constructors, can we at least add some assertions that the kind indicated is appropriate for the class being constructed?  It otherwise makes me nervous that we'll somewhere accidentally describe a bounds expression that's invalid.

include/clang/AST/Stmt.h
```diff
@@ -252,18 +252,11 @@ class LLVM_ALIGNAS(LLVM_PTR_SIZE) Stmt {
...
     unsigned : NumExprBits;
-    unsigned Kind : 1;
-  };
-
-  class NullaryBoundsExprBitFields {
-    friend class NullaryBoundsExpr;
-
-    unsigned : NumExprBits;
-    unsigned Kind : 1;
+    unsigned Kind : 3;
```
reubeno: Are there any safeguards in place that would catch someone adding to the Kind type but not updating the length of the bitfield?  IMHO, compile-time asserts are great for this sort of thing--possibly adding a sentinel value to the enum type, using a #define for the bit-field length, and a static assert to compare them with appropriate bit arithmetic.

dtarditi: clang code generally doesn't safe guard against this kind of mistake, although it could, as this is a somewhat difficult to debug mistake if you make it.

I've added a constant that names the size of the bit field. and added a static assert to guard against this.   I used the trick of using an enum member to define the constant.

dtarditi:
I've update this with changes that address the code review feedback.

Issue #53, opened 2016-09-14
prefix new keywords with underscore


dtarditi:
C has a well-established design pattern for introducing new keyword: prefix them with _.    This reduces the chance of inadvertent naming conflicts from new keywords.  This was used to introduce the Boolean type, for example. The keyword in C is _Bool. The unprefixed keywords can be used by including a header file that #defines the unprefixed keyword to the underscore based keyword. For example, the standard header file stdbool.h does this for Boolean.

We have observed that real-word code such as the Windows header files uses `ptr` as variable and field names, so we will switch to the C pattern.

dtarditi:
This work is complete.

PR #54, opened 2016-09-15
Change Checked C keywords to avoid conflicts with existing identifiers.

dtarditi:
This change modifies the existing Checked C keywords to start with
an underscore and a capital letter.  This avoids conflicts with existing
C identifiers.  This addresses issue #53.

This change follows the rules in the C Standard: all identifiers beginning
with an underscore followed by a capital letter or underscore are reserved.
This allows them to be used for language extensions without conflicting with
existing identifiers. Of course, some programs do not follow the standard
and use these identifiers.

There is a corresponding change to the GitHub Checked C repo that adds a header "stdchecked.h" that 
defines macros that map the original keywords to the new standards-compliant keywords.

Testing:
- The Checked C repo tests will be updated separately to use stdchecked.h.
- Thee existing clang and Checked C tests pass.


dtarditi:
Yes, I missed a bunch of array_ptr references and ptr references.   I'll update them to be` _Array_ptr` and `_Ptr`.  I plan to use the following rule within compiler code: when talking about concrete syntax or the language, use the _official_ name.  When implementing data structures in the compiler,, use the shortened lower-case names.  This follows what is done with _Atomic.

I don't plan to update the names for bounds expressions.   As far as I know, the names for bounds expressions occur in syntactic constructs that are new to Checked C (such as after a `:` on a declarator), so they cannot conflict with existing code.  This will save typing and keystrokes for programmers.  I think that benefit outweighs the slight inconsistency in naming.  I'd rather type `_Ptr<int> x : bounds(x, x + 5)`  insteaf of `_Ptr<int> x : _Bounds(x, x + 5)`

PR #55, opened 2016-09-17
Parse and represent bounds-safe interface type annotations

dtarditi:
This change implements parsing of bounds-safe interface type annotations.  The type annotations are a generalization of the `int * x : ptr` annotation described in the specification.    That annotation does not provide a way to represent information for pointers with multiple levels of indirection (`int **x` for example).   Instead of just using `ptr`, we will allow a checked pointer type be specified.   The annotation for something like `int **` could be `_Ptr<_Ptr<int>>`.   This richer form is needed so that we can represent bounds-safe interfaces where pointers are passed by reference, for example.'

A bounds-safe type annotation has the syntax `type` `(` _type name_ `)` where _type name_
is a C type name.   _type name_ can include typedef'ed names and type modifies.  The type annotation
can be used in place of a bounds expression (following a `:`).   For example:
```
void f(int *p : type(ptr<int>)) {
}
```
We create a subclass of `BoundsExpr` to represent the new annotation, even though this annotation is not really a bounds expression.   It will be needed in the same places in the compiler where bounds expressions are needed.

The bounds-safe interface type annotations do not need to be delay parsed for parameters or structure/union members.   They don't refer to other parameters or members.  We special case this in the parser.   This simplifies the code because we don't have to implement delay parsing for types.

For structure/union members, we now store the bounds annotation or the tokens to be delayed parsed in the FieldDeclarator structure (which already stores the bitsize expression for a member).   This uet us revent changes to the callback method that passed the tokens as extra arguments.

This change also fixes a bug that I noticed in stmt serialization.  We weren't writing out the kind field for RangeBoundsExpr.   It is possible to have an invalid kind field (this could happen if someone serializes IR with invalid bounds expressions.  I also cleaned up serialization code that needs to access some private members.  We had setter methods - I noticed the rest of the serialization code is using friend classes instead.

Testing:
- A new test for parsing these annotations will be added to the Checked C repo by a separate commit.

### DIFF
--- a/include/clang/AST/Expr.h
+++ b/include/clang/AST/Expr.h
@@ -4505,6 +4505,7 @@ public:
 
 /// \brief Represents a Checked C bounds expression.
 class BoundsExpr : public Expr {
+private:
   SourceLocation StartLoc, EndLoc;
   friend class ASTStmtReader;
 
@@ -4683,45 +4684,51 @@ public:
 /// \brief Represents a Checked C interop bounds annotation.
 ///
 /// Checked C has bounds-safe interfaces that allow global variables,
-/// functions, and members defined by unchecked code to be used in ch+ecked
-/// contexts and treated as having checked pointer types. This annotation is
-/// used to declare that a variable, parameter, or member has a bounds-safe
-//  interface that is a checked pointer type.
+/// function parameters and return values, and members that have unchecked
+/// pointer types to be used in checked contexts and be treated as having
+/// checked pointer types. This annotation declares the checked pointer
+/// type to be used as the type of the entity in the bounds-safe interface.
 ///
-/// Typically this is used to declare an unchecked-pointer typed entity
-/// as having a _Ptr type.  There is no bounds expression for this.  More
-/// generally, it can declare a bounds-safe interface this is a checked
-/// pointer to a checked pointer.  This is useful for unchecked declarations
-///  such as `int **y', where y could have a bounds-safe interface that
-/// is `ptr<ptr<int>>` or `array_ptr<ptr<int>>`.
+/// This information is needed typically at the same points where bounds
+/// information is needed, so it is convenient to store the information as a
+/// bounds expression.
 ///
-/// We typically need bounds-safe interface information at the same points
-/// where we need bounds information, so it is convenient to store the
-/// information as a bounds expression.
-class InteropTypeBoundsAnnot : public BoundsExpr {
+/// The annotation is typically used to declare that an entity has _Ptr type
+/// as its bounds-safe interface type.  More generally, an entity can have a
+/// checked pointer type to a checked pointer type and so on as its bounds-safe
+/// interface type.  This is useful for declarations such as `int **y', where
+/// `y' might have a bounds-safe interface that is `_Ptr<_Ptr<int>>` or
+/// `_Array_ptr<_Ptr<int>>`.
+///
+/// This annotation is not necessary for entities for which a bounds expression
+/// is declared and whci do not have a referent typat this a pointer  It will be
+/// inferred that the entity has _Array_ptr type as its bounds-safe interface
+/// type. Some entities will need both this annotation and a bounds expression.
+class InteropTypeBoundsAnnotation : public BoundsExpr {
 private:
   TypeSourceInfo *TIInfo;
 public:
-  InteropTypeBoundsAnnot(QualType Ty, SourceLocation StartLoc,
-                        SourceLocation EndLoc, TypeSourceInfo *tyAsWritten)
-    : BoundsExpr(InteropTypeBoundsAnnotClass, Ty, InteropTypeAnnotation,
-                 StartLoc, EndLoc), TIInfo(tyAsWritten) {
+  InteropTypeBoundsAnnotation(QualType Ty, SourceLocation StartLoc,
+                              SourceLocation EndLoc,
+                              TypeSourceInfo *TyAsWritten)
+    : BoundsExpr(InteropTypeBoundsAnnotationClass, Ty, InteropTypeAnnotation,
+                 StartLoc, EndLoc), TIInfo(TyAsWritten) {
   }
 
-  explicit InteropTypeBoundsAnnot(EmptyShell Empty)
-    : BoundsExpr(InteropTypeBoundsAnnotClass, Empty) {}
+  explicit InteropTypeBoundsAnnotation(EmptyShell Empty)
+    : BoundsExpr(InteropTypeBoundsAnnotationClass, Empty), TIInfo(nullptr) {}
 
   /// getTypeInfoAsWritten - Returns the type source info for the type
   /// in the interop annotaiton
   TypeSourceInfo *getTypeInfoAsWritten() const { return TIInfo; }
-  void setTypeInfoAsWritten(TypeSourceInfo *writtenTy) { TIInfo = writtenTy; }
+  void setTypeInfoAsWritten(TypeSourceInfo *WrittenTy) { TIInfo = WrittenTy; }
 
   /// getTypeAsWritten - Returns the type that this expression is
   /// casting to, as written in the source code.
   QualType getTypeAsWritten() const { return TIInfo->getType(); }
 
   static bool classof(const Stmt *T) {
-    return T->getStmtClass() == InteropTypeBoundsAnnotClass;
+    return T->getStmtClass() == InteropTypeBoundsAnnotationClass;
   }
 
   // Iterators
@@ -4848,7 +4855,8 @@ private:
   SourceLocation BuiltinLoc, RParenLoc;
 
   friend class ASTReader;
-  friend class ASTStmtReader;  explicit AsTypeExpr(EmptyShell Empty) : Expr(AsTypeExprClass, Empty) {}
+  friend class ASTStmtReader;
+  explicit AsTypeExpr(EmptyShell Empty) : Expr(AsTypeExprClass, Empty) {}
 
 public:
   AsTypeExpr(Expr* SrcExpr, QualType DstType,
--- a/include/clang/AST/Expr.h
+++ b/include/clang/AST/Expr.h
@@ -4505,8 +4505,8 @@ public:
 
 /// \brief Represents a Checked C bounds expression.
 class BoundsExpr : public Expr {
-private:
-  SourceLocation StartLoc, RParenLoc;
+  SourceLocation StartLoc, EndLoc;
+  friend class ASTStmtReader;
 
 public:
   enum Kind {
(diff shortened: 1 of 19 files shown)


include/clang/AST/Expr.h
```diff
@@ -4675,6 +4680,57 @@ class RangeBoundsExpr : public BoundsExpr {
   }
 };
 
+/// \brief Represents a Checked C interop bounds annotation.
+///
+/// Checked C has bounds-safe interfaces that allow global variables,
+/// functions, and members defined by unchecked code to be used in ch+ecked
```
awruef: Typo?

include/clang/AST/Expr.h
```diff
@@ -4675,6 +4680,57 @@ class RangeBoundsExpr : public BoundsExpr {
...
+/// contexts and treated as having checked pointer types. This annotation is
+/// used to declare that a variable, parameter, or member has a bounds-safe
+//  interface that is a checked pointer type.
+///
+/// Typically this is used to declare an unchecked-pointer typed entity
+/// as having a _Ptr type.  There is no bounds expression for this.  More
+/// generally, it can declare a bounds-safe interface this is a checked
+/// pointer to a checked pointer.  This is useful for unchecked declarations
+///  such as `int **y', where y could have a bounds-safe interface that
+/// is `ptr<ptr<int>>` or `array_ptr<ptr<int>>`.
```
awruef: _Ptr?

include/clang/AST/Expr.h
```diff
@@ -4675,6 +4680,57 @@ class RangeBoundsExpr : public BoundsExpr {
...
+  TypeSourceInfo *TIInfo;
+public:
+  InteropTypeBoundsAnnot(QualType Ty, SourceLocation StartLoc,
+                        SourceLocation EndLoc, TypeSourceInfo *tyAsWritten)
+    : BoundsExpr(InteropTypeBoundsAnnotClass, Ty, InteropTypeAnnotation,
+                 StartLoc, EndLoc), TIInfo(tyAsWritten) {
+  }
+
+  explicit InteropTypeBoundsAnnot(EmptyShell Empty)
+    : BoundsExpr(InteropTypeBoundsAnnotClass, Empty) {}
```
awruef: Should TTInfo be initialized to nullptr?

include/clang/AST/Expr.h
```diff
@@ -4505,8 +4505,8 @@ class GenericSelectionExpr : public Expr {
 
 /// \brief Represents a Checked C bounds expression.
 class BoundsExpr : public Expr {
-private:
-  SourceLocation StartLoc, RParenLoc;
+  SourceLocation StartLoc, EndLoc;
+  friend class ASTStmtReader;
```
reubeno: Eek, friends.  Is this a standard clang pattern?  If we can't find a clean way to avoid doing this, I'd suggest that we at least clearly comment what it is that `ASTStmtReader` will be reaching in and grabbing/changing.

include/clang/AST/Expr.h
```diff
@@ -4505,8 +4505,8 @@ class GenericSelectionExpr : public Expr {
 
 /// \brief Represents a Checked C bounds expression.
 class BoundsExpr : public Expr {
-private:
```
reubeno: Why was this explicit access label removed?  I know that members of `class`es default to being `private`, but I find it's always cleaner to be explicit, nonetheless.

include/clang/AST/Expr.h
```diff
@@ -4525,33 +4525,38 @@ class BoundsExpr : public Expr {
     // ptr interop annotation.  This isn't really a bounds expression.
     // To save space and for programmng convenience, we store the 
     // ": ptr" interop annotation as a bounds expression.
-    PtrInteropAnnotation = 5,
+    InteropTypeAnnotation = 5,
```
reubeno: I like the better name 👍

include/clang/AST/Expr.h
```diff
@@ -4675,6 +4680,57 @@ class RangeBoundsExpr : public BoundsExpr {
...
+///
+/// Checked C has bounds-safe interfaces that allow global variables,
+/// functions, and members defined by unchecked code to be used in ch+ecked
+/// contexts and treated as having checked pointer types. This annotation is
+/// used to declare that a variable, parameter, or member has a bounds-safe
+//  interface that is a checked pointer type.
+///
+/// Typically this is used to declare an unchecked-pointer typed entity
+/// as having a _Ptr type.  There is no bounds expression for this.  More
+/// generally, it can declare a bounds-safe interface this is a checked
```
reubeno: Missing word?

include/clang/AST/Expr.h
```diff
@@ -4675,6 +4680,57 @@ class RangeBoundsExpr : public BoundsExpr {
...
+///
+/// We typically need bounds-safe interface information at the same points
+/// where we need bounds information, so it is convenient to store the
+/// information as a bounds expression.
+class InteropTypeBoundsAnnot : public BoundsExpr {
+private:
+  TypeSourceInfo *TIInfo;
+public:
+  InteropTypeBoundsAnnot(QualType Ty, SourceLocation StartLoc,
+                        SourceLocation EndLoc, TypeSourceInfo *tyAsWritten)
```
reubeno: Small nit: should the `t` of `tyAsWritten` be capitalized to be consistent with the other parameters?

include/clang/AST/Expr.h
```diff
@@ -4792,8 +4848,7 @@ class AsTypeExpr : public Expr {
   SourceLocation BuiltinLoc, RParenLoc;
 
   friend class ASTReader;
-  friend class ASTStmtReader;
-  explicit AsTypeExpr(EmptyShell Empty) : Expr(AsTypeExprClass, Empty) {}
+  friend class ASTStmtReader;  explicit AsTypeExpr(EmptyShell Empty) : Expr(AsTypeExprClass, Empty) {}
```
reubeno: Did you mean to combine these onto the same source line?

include/clang/AST/Expr.h
```diff
@@ -4675,6 +4680,57 @@ class RangeBoundsExpr : public BoundsExpr {
...
+/// as having a _Ptr type.  There is no bounds expression for this.  More
+/// generally, it can declare a bounds-safe interface this is a checked
+/// pointer to a checked pointer.  This is useful for unchecked declarations
+///  such as `int **y', where y could have a bounds-safe interface that
+/// is `ptr<ptr<int>>` or `array_ptr<ptr<int>>`.
+///
+/// We typically need bounds-safe interface information at the same points
+/// where we need bounds information, so it is convenient to store the
+/// information as a bounds expression.
+class InteropTypeBoundsAnnot : public BoundsExpr {
```
reubeno: _Style_: unless there's a strong preference in the clang code base to aggressively shorten names, I'm not sure we're buying much by leaving off the `ation` from the end of `InteropTypeBoundsAnnotation`. While not all modern developers use IDEs with context-sensitive, language-aware auto-completion, we all have free access to editors with dumb yet effective token auto-completion.

Or at least those are my 2 cents 😃.

lib/AST/StmtPrinter.cpp
```diff
@@ -1845,10 +1845,7 @@ void StmtPrinter::VisitNullaryBoundsExpr(NullaryBoundsExpr *Node) {
       break;
     case BoundsExpr::None:
       OS << "bounds(none)";
-      break;
-    case BoundsExpr::PtrInteropAnnotation:
-      OS << "ptr";
-      break;
+      break;;
```
reubeno: Stray semicolon?

lib/Parse/ParseExpr.cpp
```diff
@@ -2753,17 +2753,49 @@ void Parser::ParseBlockId(SourceLocation CaretLoc) {
...
+  if (tok.getKind() == tok::identifier) {
     IdentifierInfo *Ident = Tok.getIdentifierInfo();
     return (Ident == Ident_byte_count || Ident == Ident_count ||
             Ident == Ident_bounds);
   }
   return false;
 }
 
+bool Parser::StartsInteropTypeAnnotation(Token &tok) {
+  switch (tok.getKind()) {
```
reubeno: Does this mean that the type annotation can't contain any qualifiers (e.g. `const`)?  I'm wondering how confusing that will be to see something like:

    void some_func(const int *p : _Ptr<int>);

It's sort of the "safe" revisionist type for `p` and sort of not.

lib/Parse/ParseExpr.cpp
```diff
@@ -2753,17 +2753,49 @@ void Parser::ParseBlockId(SourceLocation CaretLoc) {
...
+    case tok::kw__Ptr:
+    case tok::kw__Array_ptr:
```
reubeno: We'll want to keep this list in sync with the full list of all checked-c pointer types (including spans and whatever else we add).  Is there already a helper function elsewhere that checks if a given token is a checked-c type token?

lib/Parse/ParseExpr.cpp
```diff
@@ -2753,17 +2753,49 @@ void Parser::ParseBlockId(SourceLocation CaretLoc) {
...
+      return true;
+    default:
+      return false;
+  }
+}
+
+ExprResult Parser::ParseInteropTypeAnnotation() {
+  if (StartsInteropTypeAnnotation(Tok)) {
+    TypeResult Ty = ParseTypeName();
```
reubeno: Should we perform any additional inspection/validation?  See my overarching question on the PR.

lib/Sema/TreeTransform.h
```diff
@@ -11535,6 +11539,12 @@ TreeTransform<Derived>::TransformRangeBoundsExpr(RangeBoundsExpr *E) {
                                              E->getRParenLoc());
 }
 
+template<typename Derived>
+ExprResult
+TreeTransform<Derived>::TransformInteropTypeBoundsAnnot(InteropTypeBoundsAnnot *E) {
```
reubeno: Don't we need to "recur" on the type inside the annotation?  I'm relatively unfamiliar with this code, but I'm guessing there's some facility that might be used, say, to rename or transform types within an AST.  I'd assume we need to avoid pruning that recursive walk at these annotations, in case the type inside it matches a type of interest for the caller.

Feel free to ignore if I'm making no sense here.

lib/Serialization/ASTReaderStmt.cpp
```diff
@@ -960,23 +960,31 @@ void ASTStmtReader::VisitCountBoundsExpr(CountBoundsExpr *E) {
   VisitExpr(E);
   E->setKind((BoundsExpr::Kind)Record[Idx++]);
   E->setCountExpr(Reader.ReadSubExpr());
-  E->setStartLoc(ReadSourceLocation(Record, Idx));
-  E->setRParenLoc(ReadSourceLocation(Record, Idx));
+  E->StartLoc = ReadSourceLocation(Record, Idx);
```
reubeno: What was the reason to make this change?  Were you thinking that making it private avoided other code from making unsupported changes (at the risk of letting this one piece of code in on the private members)?

lib/Serialization/ASTWriterStmt.cpp
```diff
@@ -929,13 +929,22 @@ void ASTStmtWriter::VisitNullaryBoundsExpr(NullaryBoundsExpr *E) {
...
   VisitExpr(E);
+  Record.push_back(E->getKind());
   VisitExpr(E->getLowerExpr());
   VisitExpr(E->getUpperExpr());
   Record.AddSourceLocation(E->getStartLoc());
   Record.AddSourceLocation(E->getRParenLoc());
   Code = serialization::EXPR_RANGE_BOUNDS_EXPR;
 }
 
+void ASTStmtWriter::VisitInteropTypeBoundsAnnot(InteropTypeBoundsAnnot *E) {
```
reubeno: Visit the inner types?

dtarditi:
Reuben has suggested that we make type annotations in bounds-safe interfaces be parallel bounds declarations and have a keyword that starts them.   This would allow us to parse an arbitrary type name, and allow the use of type names in bounds-safe type annotation.   So we would write something like:

int *x : type(ptr<int>);

or

typedef ptr<int> pint

int *x : type(pint);

The idenitifer 'type' is my suggestion for the keyword.   I'm open to better suggestions, of course.

I think Reuben's suggestion is great, so I'm going to update the PR with this change.

include/clang/AST/Expr.h
dtarditi: Sure, I'll put it back.  I noticed some of the other code in the file was taking advantage of the default behavior.

include/clang/AST/Expr.h
dtarditi: clang has an 80 character line limit, so long identifier names can cause problems with fitting expressions or parameter lists on to the line.   I agree that a more descriptive name is better, so I'll update the name.

dtarditi:
I've updated the pull request based to address Reuben's code review feedback.  The biggest change is the change to the syntax of the type annotations.  They now have the form `type` `(` _type name_ `)`.

PR #56, opened 2016-09-20
Constraint system is higher order now

awruef:
This isn't totally finished (more on that later) but it's so divergent from where it was a week ago that I'd like to get this reviewed now and work on finishing the edges tomorrow. What's changed: 

 * The core data structure has changed. Before, there was a mapping from VarDecl to a constraint variable. Now, there is a mapping from a PersistentSourceLoc to a set of ConstraintVariables, which is a new class. This is good because both PersistentSourceLoc and ConstraintVariable are not tied to a particular AST data structure, so the system is inherently persistent. 
 * There is now a new ConstraintVariable class. A ConstraintVariable is either a PointerVariableConstraint or a FunctionVariableConstraint. FunctionVariableConstraint tracks ConstraintVariables for the return and parameter types. 
 * There is a new unit test file.

These two changes necessitated many downstream changes that touched almost everything. Currently, it passes the unit tests I've committed. What's left: 

 * Write more tests, mostly for variables / functions declared in macro regions, and functions / multi-line decls where the type names are macro-defined.
 * Change the function linking phase to work with the new information, and test it.


tools/checked-c-convert/MappingVisitor.cpp
```diff
@@ -28,13 +28,19 @@ bool MappingVisitor::VisitDeclStmt(DeclStmt *S) {
       Type *T = NULL;
       std::tie<Stmt *, Decl *, Type *>(So, D, T) = PSLtoSDT[PSL];
       if (So != NULL && Verbose) {
+        errs() << "\nOverrding ";
```
dtarditi: >     errs() << "\nOverrding "; [](start = 3, length = 30)

Typo?

dtarditi:
> // to AST elements of the program.

Does this comment need to be updated?


---
Refers to: tools/checked-c-convert/ProgramInfo.h:9 in 184f3b5. [](commit_id = 184f3b58d851f360138876390c1c58787c2b57c4, deletion_comment = False)

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -33,80 +33,123 @@
...
+  virtual void constrainTo(Constraints &CS, ConstAtom *C) = 0;
+  virtual bool anyChanges(Constraints::EnvironmentMap &E) = 0;
 
-  std::string getName() { return Name; }
+  std::string getTy() { return BaseType; }
 };
 
-class GlobalVariableSymbol : public GlobalSymbol {
+class PointerVariableConstraint;
+class FunctionVariableConstraint;
```
dtarditi: Rename these to PointerConstraint and FunctionConstraint?

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -33,80 +33,123 @@
...
+
+// Represents an individual constraint on a pointer variable. 
+// This could contain a reference to a FunctionVariableConstraint
+// in the case of a function pointer declaration.
+class PointerVariableConstraint : public ConstraintVariable {
 private:
-  std::set<uint32_t> ConstraintVars;
+  CVars vars;
```
dtarditi: Add a comment explaining what vars is?

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -33,80 +33,123 @@
...
   }
+
+  std::string mkString(Constraints::EnvironmentMap &E);
+
+  FunctionVariableConstraint *getFV() { return FV; }
+
+  void print(llvm::raw_ostream &O) const ;
+  void dump() const { print(llvm::errs()); }
+  void constrainTo(Constraints &CS, ConstAtom *C);
+  bool anyChanges(Constraints::EnvironmentMap &E);
```
dtarditi: Please add comments describing what these functions do.

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -120,15 +163,16 @@ class ProgramInfo {
 
   // For each pointer type in the declaration of D, add a variable to the 
   // constraint system for that pointer type. 
-  bool addVariable(clang::Decl *D, clang::DeclStmt *St, clang::ASTContext *C);
+  bool addVariable(clang::DeclaratorDecl *D, clang::DeclStmt *St, clang::ASTContext *C);
 
   bool getDeclStmtForDecl(clang::Decl *D, clang::DeclStmt *&St);
 
   // Checks the structural type equality of two constraint variables. This is 
   // needed if you are casting from U to V. If this returns true, then it's
```
dtarditi: Out-of-date comment?

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -33,80 +33,123 @@
...
 public:
-  GlobalVariableSymbol(std::string N, PersistentSourceLoc P, 
-    std::set<uint32_t> S) 
-    : GlobalSymbol(Variable, N, P),ConstraintVars(S) {}
+  PointerVariableConstraint(CVars V, std::string T) : 
+    ConstraintVariable(PointerVariable, T),vars(V),FV(nullptr) {}
+  PointerVariableConstraint(clang::DeclaratorDecl *D, uint32_t &K,
+    Constraints &CS);
+  PointerVariableConstraint(const clang::Type *Ty, uint32_t &K,
+    std::string N, Constraints &CS);
```
dtarditi: Please add a comment explaining what this function does.

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -33,80 +33,123 @@
 #include "utils.h"
 #include "PersistentSourceLoc.h"
 
-// A helper class used to track global information in the program. 
-class GlobalSymbol {
+typedef std::set<uint32_t> CVars;
```
dtarditi: It'd be helpful to add a comment stating that set actually keeps its contents sorted.    This might be obvious to an expert C++ developer - it will not be obvious to developers less familiar with C.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -8,52 +8,313 @@
...
+      llvm_unreachable("impossible");
+      break;
+    }
+  }
+
+  if(emittedBase == false) {
+    // If we have a FV pointer, then our "base" type is a function pointer
+    // type.
+    if (FV) {
+      s = s + FV->mkString(E);
```
dtarditi: I think this will generate an incorrect string in the case where the enclosing pointer type is an arr or wild pointer type.    The C syntax for pointers to function is baroque.  Given a pointer to a function that returns an integer and takes a char as an input, I think you need to write int (*)(char c);

tools/checked-c-convert/ProgramInfo.h
awruef: I added comments at the virtual declaration in the base class.

tools/checked-c-convert/ProgramInfo.h
awruef: I think that's still what this function does. I just refined the type from Decl to DeclaratorDecl.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -8,52 +8,313 @@
...
+      } else {
+        assert(BaseType.size() > 0);
+        emittedBase = true;
+        if (FV) {
+          s = s + FV->mkString(E);
+          wroteFPtr = true;
+        } else {
+          s = s + BaseType + "*";
+        }
+      }
```
dtarditi: I'm not convinced this produces the right type for function pointers.   The C syntax is very baroque.    If you have a pointer to a function that takes an integer and returns a character, the type is char (*)(int).   If you have a pointer to a pointer to a function that does, the type is char (**)(int).

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -33,80 +33,123 @@
...
+typedef PointerVariableConstraint PVConstraint;
+
+// Constraints on a function type. Also contains a 'name' parameter for 
+// when a re-write of a function pointer is needed.
+class FunctionVariableConstraint : public ConstraintVariable {
 private:
-  std::vector<std::set<uint32_t> > ParameterConstraintVars;
-  std::set<uint32_t> ReturnConstraintVars;
+  std::set<ConstraintVariable*> returnVars;
+  std::vector<std::set<ConstraintVariable*>> paramVars;
```
dtarditi: Please add comments explaining how these constraints work.  As far as I can tell, the first one is N constraints on the return value.   The second is a vector of N constrains, where each element k of the vector corresponds to the kth parameter.

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -33,80 +33,123 @@
...
+  std::string name;
```
dtarditi: > std::string name; [](start = 1, length = 18)

I'm missing what the 'name' instance member is.  Is it the name of a function?   Is it the name of a function pointer variable?

tools/checked-c-convert/Constraints.h
```diff
@@ -416,8 +416,8 @@ class Constraints {
   typedef std::map<VarAtom*, ConstAtom*, PComp<VarAtom*> > EnvironmentMap;
 
   bool addConstraint(Constraint *c);
-  ConstraintSet getConstraints() { return constraints; }
-  EnvironmentMap getVariables() { return environment; }
+  ConstraintSet &getConstraints() { return constraints; }
```
dtarditi: Explain that it is important for efficiency to return them by reference.

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -77,29 +79,38 @@ const Type *getNextTy(const Type *Ty) {
 }
 
 // Test to see if we can rewrite a given SourceRange. 
+// Note that R.getRangeSize will return -1 if SR is within
+// a macro as well.
```
dtarditi: Please clarify that the clang rewriter infrastructure won't let us rewrite macros at this time.   This is because R.getRangeSize returns -1 if SR is within a macro as well.

tools/checked-c-convert/ProgramInfo.cpp
awruef: This probably won't work in the case where any of the constraints on the function pointer are not PTR.

PR #57, opened 2016-09-20
Constraint generation is higher order now.

awruef:
Changed to a squash merge.


tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -12,6 +12,125 @@
...
+void constrainEq(std::set<ConstraintVariable*> &RHS,
+  std::set<ConstraintVariable*> &LHS, ProgramInfo &Info);
+// Given two ConstraintVariables, do the right thing to assign 
+// constraints. 
+// If they are both PVConstraint, then do an element-wise constraint
+// generation.
+// If they are both FVConstraint, then do a return-value and parameter
+// by parameter constraint generation.
+// If they are of an unequal parameter type, constrain everything in both
+// to top.
```
dtarditi: >top. [](start = 6, length = 4)

Use the term wild instead?

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -8,52 +8,313 @@
...
+      } else {
+        assert(BaseType.size() > 0);
+        emittedBase = true;
+        if (FV) {
+          s = s + FV->mkString(E);
+          wroteFPtr = true;
+        } else {
+          s = s + BaseType + "*";
+        }
+      }
```
dtarditi: We agree that this probably won't work where any of the constraints on the function pointer are not PTR.   You should probably make the code not do rewrites (for now) in the cases that will go wrong.

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -33,80 +33,140 @@
...
+  CVars vars;
+  FunctionVariableConstraint *FV;
 public:
-  GlobalVariableSymbol(std::string N, PersistentSourceLoc P, 
-    std::set<uint32_t> S) 
-    : GlobalSymbol(Variable, N, P),ConstraintVars(S) {}
+  PointerVariableConstraint(CVars V, std::string T) : 
+    ConstraintVariable(PointerVariable, T),vars(V),FV(nullptr) {}
+  PointerVariableConstraint(clang::DeclaratorDecl *D, uint32_t &K,
+    Constraints &CS);
```
dtarditi: Please add a comment explaining what this function does.

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -120,15 +180,16 @@ class ProgramInfo {
 
   // For each pointer type in the declaration of D, add a variable to the 
   // constraint system for that pointer type. 
-  bool addVariable(clang::Decl *D, clang::DeclStmt *St, clang::ASTContext *C);
+  bool addVariable(clang::DeclaratorDecl *D, clang::DeclStmt *St, clang::ASTContext *C);
 
   bool getDeclStmtForDecl(clang::Decl *D, clang::DeclStmt *&St);
 
   // Checks the structural type equality of two constraint variables. This is
```
dtarditi: >two constraint variables. [](start = 44, length = 25)

The arguments are now sets, now individual variables.   What does this do for the the sets of variables?

test/CheckedCRewriter/function_tests.c
```diff
@@ -0,0 +1,52 @@
+// Tests for Checked C rewriter tool.
```
dtarditi: How about tests for parameters that take function pointer types, returning a function pointer type, and arrays of function pointer types?

tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -12,6 +12,125 @@
...
+        constrainEq(FCLHS, FCRHS, Info);
+      else
+        llvm_unreachable("TODO");
+    } else {
+      CRHS->dump();
+      errs() << "\n";
+      CLHS->dump();
+      errs() << "\n";
+      // Constrain everything in both to top.
+      llvm_unreachable("TODO");
```
dtarditi: Could you file an issue about the "TODO" work in this section?

tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -57,51 +178,75 @@ class FunctionVisitor : public RecursiveASTVisitor<FunctionVisitor> {
   // In any of these cases, due to conditional expressions, the number of
   // variables on the RHS could be 0 or more. We just do the same rule
   // for each pair of q_i to q_j \forall j in variables_on_rhs.
-  void constrainAssign(uint32_t V, Expr *RHS) {
+  void constrainAssign( std::set<ConstraintVariable*> V, 
+                        Expr *RHS) {
```
dtarditi: Please add an explanation of what V is.

tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -57,51 +178,75 @@ class FunctionVisitor : public RecursiveASTVisitor<FunctionVisitor> {
...
+      constrainEq(V, W, Info);
     } else {
       // Cases 2-4.
       if (RHS->isIntegerConstantExpr(*Context)) {
         // Case 2.
         if (!RHS->isNullPointerConstant(*Context,
-                                        Expr::NPC_ValueDependentIsNotNull))
-          CS.addConstraint(CS.createEq(CS.getOrCreateVar(V), CS.getWild()));
+          Expr::NPC_ValueDependentIsNotNull))
+          for (const auto &U : W)
```
dtarditi: I don't understand why the iteration is over W.   I would think that it should be over V.

tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -152,96 +291,88 @@ class FunctionVisitor : public RecursiveASTVisitor<FunctionVisitor> {
...
-    }
+    std::set<ConstraintVariable*> Fun =
+      Info.getVariable(Function, Context);
+    std::set<ConstraintVariable*> Var =
+      Info.getVariable(S->getRetValue(), Context);
+
+    // Constrain the value returned (if present) against the return value
+    // of the function.   
+    for (const auto &F : Fun )
+      if (FVConstraint *FU = dyn_cast<FVConstraint>(F))
```
dtarditi: >FU [](start = 24, length = 2)

Probably not a good abbbreviation for a variable name.

tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -152,96 +291,88 @@ class FunctionVisitor : public RecursiveASTVisitor<FunctionVisitor> {
...
+       constrainEq(FU->getReturnVars(), Var, Info); 
 
     return true;
   }
 
-  bool VisitUnaryPreInc(UnaryOperator *O) {
-    std::set<uint32_t> V;
+  void constrainExprFirst(Expr *E) {
```
dtarditi: >void constrainExprFirst(Expr *E) { [](start = 2, length = 34)

Please add a comment explaining what this function does.

dtarditi:
This is a nice generalization of the existing code.  Thanks for addressing my first set of comments.  I've finished the rest of my first rounds of reviews.   I"ve made some requests for additional comments and have a few questions about the code.

test/CheckedCRewriter/function_tests.c
awruef: I don't think arrays of function pointer types will work yet but I'll add it in the next branch.

tools/checked-c-convert/ConstraintBuilder.cpp
awruef: That makes sense, changing.

tools/checked-c-convert/ProgramInfo.cpp
awruef: I added a test to `PointerVariableConstraint::anyChanges` so that the re-write won't happen in that case.

tools/checked-c-convert/ProgramInfo.h
awruef: Right now, nothing, since `checkStructuralEquality` always returns `false`. When it starts returning something other than `false`, it will probably have to perform the comparison on the LUB of `V` and `U`.

awruef:
I've addressed most of the feedback. For now we don't do any re-writing in "extreme" function pointer cases like arrays of function pointers. I'll write tests for those and other cases next.

tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -178,6 +177,9 @@ class FunctionVisitor : public RecursiveASTVisitor<FunctionVisitor> {
   // In any of these cases, due to conditional expressions, the number of
   // variables on the RHS could be 0 or more. We just do the same rule
   // for each pair of q_i to q_j \forall j in variables_on_rhs.
+  //
+  // V is the set of constraint variables on the left hand side that we are
```
dtarditi: Could you amplify this comment some more?   It's the set of constraint variables for the LHS.    The LHS will have a pointer type and the constraint variables represent the constraints for that type.

tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -57,51 +177,78 @@ class FunctionVisitor : public RecursiveASTVisitor<FunctionVisitor> {
   // In any of these cases, due to conditional expressions, the number of
   // variables on the RHS could be 0 or more. We just do the same rule
   // for each pair of q_i to q_j \forall j in variables_on_rhs.
-  void constrainAssign(uint32_t V, Expr *RHS) {
+  //
+  // V is the set of constraint variables on the left hand side that we are
+  // assigning to.
+  void constrainAssign( std::set<ConstraintVariable*> V,
```
dtarditi: Could you amplify this comment some more? It's the set of constraint variables for the LHS. The LHS will have a pointer type and the constraint variables represent the constraints for that type.

PR #58, opened 2016-09-21
Finishing the migration to the new Variable data type.

awruef:
* Conservative casting implemented so both sides are constrained to WILD
 * Statistics printing works again
 * More tests

### DIFF
--- a/tools/checked-c-convert/CheckedCConvert.cpp
+++ b/tools/checked-c-convert/CheckedCConvert.cpp
@@ -379,13 +379,13 @@ void emit(Rewriter &R, ASTContext &C, std::set<FileID> &Files,
           // For example \foo\bar\a.c should become \foo\bar\a.checked.c
           // if the OutputPostfix parameter is "checked" .
 
-          StringRef pfName = sys::path::filename(FE->getName());
-          StringRef dirName = sys::path::parent_path(FE->getName());
-          StringRef fileName = sys::path::remove_leading_dotslash(pfName);
-          StringRef ext = sys::path::extension(fileName);
-          StringRef stem = sys::path::stem(fileName);
-          Twine nFileName = stem + "." + OutputPostfix + ext;
-          Twine nFile = dirName + sys::path::get_separator() + nFileName;
+          std::string pfName = sys::path::filename(FE->getName()).str();
+          std::string dirName = sys::path::parent_path(FE->getName()).str();
+          std::string fileName = sys::path::remove_leading_dotslash(pfName).str();
+          std::string ext = sys::path::extension(fileName).str();
+          std::string stem = sys::path::stem(fileName).str();
+          std::string nFileName = stem + "." + OutputPostfix + ext;
+          std::string nFile = dirName + sys::path::get_separator().str() + nFileName;
           
           // Write this file out if it was specified as a file on the command
           // line.
@@ -399,11 +399,11 @@ void emit(Rewriter &R, ASTContext &C, std::set<FileID> &Files,
 
           if(canWrite(feAbsS, InOutFiles, BaseDir)) {
             std::error_code EC;
-            raw_fd_ostream out(nFile.str(), EC, sys::fs::F_None);
+            raw_fd_ostream out(nFile, EC, sys::fs::F_None);
 
             if (!EC) {
               if (Verbose)
-                outs() << "writing out " << nFile.str() << "\n";
+                outs() << "writing out " << nFile << "\n";
               B->write(out);
             }
             else
@@ -445,7 +445,11 @@ public:
     for (const auto &V : Info.getVarMap()) {
       PersistentSourceLoc PLoc = V.first;
       std::set<ConstraintVariable*> Vars = V.second;
-      assert(Vars.size() > 0 && Vars.size() <= 2);
+      // I don't think it's important that Vars have any especial size, but 
+      // at one point I did so I'm keeping this comment here. It's possible 
+      // that what we really need to do is to ensure that when we work with
+      // either PV or FV below, that they are the LUB of what is in Vars.
+      // assert(Vars.size() > 0 && Vars.size() <= 2);
 
       // PLoc specifies the location of the variable whose type it is to 
       // re-write, but not where the actual type storage is. To get that, we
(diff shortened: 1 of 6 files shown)


tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -470,13 +474,26 @@ class RewriteConsumer : public ASTConsumer {
...
+          else if (FVConstraint *T = dyn_cast<FVConstraint>(V))
+            FV = T;
+        }
 
         if (PV && PV->anyChanges(Info.getConstraints().getVariables())) {
+          // Rewrite a declaration.
           std::string newTy = PV->mkString(Info.getConstraints().getVariables());
           rewriteThese.insert(DAndReplace(DeclNStmt(D, DS), newTy));
+        } else if (FV && FV->anyChanges(Info.getConstraints().getVariables())) {
+          // Rewrite a function variables return value.
```
dtarditi: Typo in function variables return value?  I think you need an apostrophe after variables.  Alternately, you could write "return value of a function variable".

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -470,13 +474,26 @@ class RewriteConsumer : public ASTConsumer {
...
+          std::set<ConstraintVariable*> V = FV->getReturnVars();
+          if (V.size() > 0) {
+            std::string newTy =
```
dtarditi: Why is it OK to use only the first variable in the set?

tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -235,6 +242,7 @@ class FunctionVisitor : public RecursiveASTVisitor<FunctionVisitor> {
                 for (const auto &B : PVC->getCvars())
                   CS.addConstraint(
                     CS.createEq(CS.getOrCreateVar(B), CS.getWild()));
+            CS.dump();
```
dtarditi: Stray debug print statement?

tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -311,11 +317,62 @@ class FunctionVisitor : public RecursiveASTVisitor<FunctionVisitor> {
...
-          llvm_unreachable("TODO");
+          Constraints &CS = Info.getConstraints();
+          for (const auto &C : ParameterEC)
+            C->constrainTo(CS, CS.getWild());
         }
 
         i++;
       }
+    } else if (DeclaratorDecl *DD = dyn_cast<DeclaratorDecl>(D)){
+      // This could be a function pointer.
```
dtarditi: This comment is a little confusing: it could be a function pointer or a function, right?

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -46,7 +50,7 @@ PointerVariableConstraint::PointerVariableConstraint(const Type *_Ty,
   // function, then create a base-level FVConstraint that we carry 
   // around too.
   if (Ty->isFunctionType()) {
-    FV = new FVConstraint(Ty, K, N, CS);
+    FV = new FVConstraint(Ty, K, (isTypedef ? "" : N), CS);
```
dtarditi: This deserves a comment.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -63,6 +67,9 @@ void PointerVariableConstraint::print(raw_ostream &O) const {
   for (const auto &I : vars) 
     O << "q_" << I << " ";
   O << " }";
+
+  if (FV)
+    FV->print(O);
```
dtarditi: Probably worth parenthesizing the dumped FV.   This will avoid confusion in reading dumps when you have pointers to functions that return pointers and so on.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -248,7 +255,6 @@ void FunctionVariableConstraint::print(raw_ostream &O) const {
 
 std::string
 FunctionVariableConstraint::mkString(Constraints::EnvironmentMap &E) {
-  assert(name.size() > 0);
   std::string s = "";
   // TODO punting on what to do here. The right thing to do is to figure out
```
dtarditi: Please file an issue on the GitHub issue tracker for this.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -293,14 +299,38 @@ void ProgramInfo::print(raw_ostream &O) const {
   }
 }
 
+static
```
dtarditi: Add a comment describing this function.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -363,123 +400,42 @@ bool ProgramInfo::link() {
...
+    if (P.size() > 1) {
+      std::set<FVConstraint*>::iterator I = P.begin();
+      std::set<FVConstraint*>::iterator J = P.begin();
+      ++J;
+      
+      while (J != P.end()) {
+        FVConstraint *P1 = *I;
+        FVConstraint *P2 = *J;
+
+        // Constraint the return values to be equal
```
dtarditi: Typo?

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -363,123 +400,42 @@ bool ProgramInfo::link() {
...
-
-          for (const auto &VV : pVars2)
-            for (const auto &V : VV)
-              CS.addConstraint(CS.createEq(
-                CS.getOrCreateVar(V), CS.getWild()));
+          // It could be the case that P1 or P2 is missing a prototype, in
+          // which case we don't need to constrain anything.
+          if (P1->hasProtoType() && P2->hasProtoType()) {
+            // Nope, we have no choice. Constrain everything to wild.
+            llvm_unreachable("TODO");
```
dtarditi: Please file an issue in the GitHub issue system or file this in.

tools/checked-c-convert/CheckedCConvert.cpp
awruef: Because they're all alised to be equal to each other, so one should be just as good as the other.

tools/checked-c-convert/ConstraintBuilder.cpp
awruef: D is the `Decl` from `CallExpr::getCalleeDecl`. The `FunctionDecl` case is the prior `if` case, so in this case we check to see if it's a call through a function pointer. I added a comment to that case.

tools/checked-c-convert/ProgramInfo.cpp
awruef: Opened #59.

awruef:
I think I addressed this in the last two commits.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -49,9 +49,18 @@ PointerVariableConstraint::PointerVariableConstraint(const Type *_Ty,
   // If, after boiling off the pointer-ness from this type, we hit a 
   // function, then create a base-level FVConstraint that we carry 
   // around too.
-  if (Ty->isFunctionType()) {
+  if (Ty->isFunctionType()) 
+    // For the moment, if a type is typedefed, we don't need to print
```
dtarditi: I think this comment is still a little unclear.  How about:  "C function-pointer type declarator syntax embeds the variable name within the function-like syntax.  For example:
void (*fname)(int, int) = ...;
If a typedef'ed type name is used, the name can be omitted because it is not embedded like that.  Instead, it has the form
tn fname = ...,
where tn is the typedef'ed type name.

dtarditi:
I think one comment could be clarified a bit.    Otherwise, LGTM.

PR #60, opened 2016-09-23
Small changes for evaluation

awruef:
Last stuff before leaving:
 * Constrain varargs and tests
 * Explicitly constrain variables that occur at the same location during linking
 * Fix case where function `anyChanges` call would be overly aggressive


tools/checked-c-convert/utils/run.py
```diff
@@ -0,0 +1,66 @@
...
+import argparse
+import traceback
+import subprocess
+
+"""
+This tool will invoke checked-c-convert on a compile_commands.json database. 
+It contains some work-arounds for cmake+nmake generated compile_commands.json 
+files, where the files are malformed. 
+
+To use, set PROG_NAME to the path where checked-c-convert.exe is built.
```
dtarditi: How about making PROG_NAME an argument to the script?  This will be easier for people to use than having to edit the script.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -44,6 +44,9 @@ PointerVariableConstraint::PointerVariableConstraint(const Type *_Ty,
     vars.insert(K);
     CS.getOrCreateVar(K);
     K++;
+
+    if (tyToStr(Ty) == "struct __va_list_tag *")
```
dtarditi: There should be some way in clang to test this without having to serialize it to string.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -63,7 +66,7 @@ PointerVariableConstraint::PointerVariableConstraint(const Type *_Ty,
   BaseType = tyToStr(Ty);
 
   // Special case for void to not make _Ptr<void> pointers.
-  if( BaseType == "void" ) 
+  if( BaseType == "void" || BaseType == "struct __va_list_tag *" )
```
dtarditi: There should be some way to test the type directly, instead of serializing it to a string.

awruef: I agree. I looked for this, but couldn't find it. varargs is not a "builtin" type, it's introduced as something else (the list of builtins is enumerated [here](https://github.com/Microsoft/checkedc-clang/blob/master/include/clang/AST/BuiltinTypes.def). I'll make a work item (#61) for it. I can deal with the void case, though.

PR #62, opened 2016-10-03
Implement implicit type conversions for bounds-safe interfaces

dtarditi:
This change implements implicit conversions from checked pointers to unchecked pointers at bounds-safe interfaces.  This conversions may happen at assignments to variables or members and at function calls.

The conversions are straightforward to implement.  The code needs to choose whether to use the original type or the interoperation type.  This change adds code to SemaExpr.cpp and SemaInit.cpp that makes this decision.    Because the code is quite similar in both cases, we do not put the decision-making inline around the current calls to CheckSingleAssignmentConstraints.  Instead we factor it into a function called ResolveSingleAssignmentType.  This also helps deal with fact that CheckSingleAssignmentContraints is tricky to use: it has default parameters that control whether it emits diagnostics and side-effects the right-hand side expression.   ResolveSingleAssignmentType needs to try each potential left-hand side by calling CheckSingleAssignmentConstraints.  It invokes CheckSingleAssignmentContraints with parameters that prevent side-effects.   

This change also adds code that computes the Checked C interop type for a variable, member, or expression being initialized.

Testing:
- A new file interop.c will be committed separately to the Checked C repo.  The file has tests for implicit conversions at bounds-safe interfaces.  It will be added to the test\typechecking subdirectory.

### DIFF
--- a/include/clang/Sema/Sema.h
+++ b/include/clang/Sema/Sema.h
@@ -8839,8 +8839,17 @@ public:
 private:
   QualType GetInteropType(const ValueDecl *Decl);
 public:
+  /// \brief Get the bounds-safe interface type for Entity.
+  /// Returns a null QualType if there isn't one.
   QualType GetCheckedCInteropType(const InitializedEntity &Entity);
+
+  /// \brief Get the bounds-safe interface type for LHS.
+  // Returns a null QualType if there isn't one.
   QualType GetCheckedCInteropType(ExprResult LHS);
+
+  /// \brief Chose between using LHSType or LHSInteropType for the type
+  /// of the left-hand side of a single assignment from the RHS. Try using
+  /// LHSType and if that doesn't work, try uisng LHSInteropType.
   QualType ResolveSingleAssignmentType(QualType LHSType, QualType LHSInteropType, 
                                        ExprResult &RHS);
 
--- a/include/clang/Sema/Sema.h
+++ b/include/clang/Sema/Sema.h
@@ -8836,8 +8836,13 @@ public:
                                                      bool DiagnoseCFAudited = false,
                                                      bool ConvertRHS = true);
 
-  bool CheckedCInteropConversion(const InitializedEntity &Entity, ExprResult &RHS);
-  bool CheckedCInteropConversion(ExprResult LHS, ExprResult &RHS);  
+private:
+  QualType GetInteropType(const ValueDecl *Decl);
+public:
+  QualType GetCheckedCInteropType(const InitializedEntity &Entity);
+  QualType GetCheckedCInteropType(ExprResult LHS);
+  QualType ResolveSingleAssignmentType(QualType LHSType, QualType LHSInteropType, 
+                                       ExprResult &RHS);
 
   // \brief If the lhs type is a transparent union, check whether we
   // can initialize the transparent union with the given expression.
--- a/include/clang/Sema/Sema.h
+++ b/include/clang/Sema/Sema.h
@@ -8844,13 +8844,19 @@ public:
   QualType GetCheckedCInteropType(const InitializedEntity &Entity);
 
   /// \brief Get the bounds-safe interface type for LHS.
-  // Returns a null QualType if there isn't one.
+  /// Returns a null QualType if there isn't one.
   QualType GetCheckedCInteropType(ExprResult LHS);
 
-  /// \brief Chose between using LHSType or LHSInteropType for the type
-  /// of the left-hand side of a single assignment from the RHS. Try using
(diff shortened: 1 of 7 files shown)


include/clang/Sema/Sema.h
```diff
@@ -8836,6 +8836,24 @@ class Sema {
                                                      bool DiagnoseCFAudited = false,
                                                      bool ConvertRHS = true);
 
+private:
+  QualType GetInteropType(const ValueDecl *Decl);
```
reubeno: Should this functions' also be qualified with `CheckedC`?

lib/Sema/SemaExpr.cpp
```diff
@@ -8058,6 +8058,112 @@ Sema::CheckSingleAssignmentConstraints(QualType LHSType, ExprResult &CallerRHS,
...
+///
+/// This falls into two cases:
+/// 1. The interface is a type annotation: return the type in the
+///    annotation.
+/// 2. The interface is a bounds expression.  This implies the checked
+/// type should be an _Array_ptr type.  Construct an _Array_ptr version
+/// of the unchecked ponter type for the declaration and return it.
+QualType Sema::GetInteropType(const ValueDecl *Decl) {
+  const DeclaratorDecl *TargetDecl = nullptr;
+  if (const FieldDecl *Field = dyn_cast<FieldDecl>(Decl))
```
reubeno: Is the field- and var-specific logic needed?  Is there not a way to just directly cast to a `DeclaratorDecl`?  Or is this also serving the purpose of limiting which declarations can have interop types?

lib/Sema/SemaExpr.cpp
```diff
@@ -8058,6 +8058,112 @@ Sema::CheckSingleAssignmentConstraints(QualType LHSType, ExprResult &CallerRHS,
...
+    TargetDecl = Field;
+  else if (const VarDecl *Var = dyn_cast<VarDecl>(Decl))
+    TargetDecl = Var;
+
+  if (TargetDecl) {
+    if (const BoundsExpr *Bounds = TargetDecl->getBoundsExpr()) {
```
reubeno: In the future, if and when we support multiple bounds expressions on a declaration--I'm assuming you'll need to take a full pass over all calls to this accessor.  Even if we don't have the concrete syntax now, might it make sense to start preparing for that eventuality in the code--so that new code you write will at least consider the possibility of multiple declarations?

lib/Sema/SemaExpr.cpp
```diff
@@ -8058,6 +8058,112 @@ Sema::CheckSingleAssignmentConstraints(QualType LHSType, ExprResult &CallerRHS,
...
+      switch (Bounds->getKind()) {
+        case BoundsExpr::Kind::InteropTypeAnnotation: {
+          const InteropTypeBoundsAnnotation *Annot =
+            dyn_cast<InteropTypeBoundsAnnotation>(Bounds);
+          assert(Annot && "unexpected dyn_cast failure");
+          if (Annot != nullptr)
+            return Annot->getType();
```
reubeno: Did you mean to fall-through here?  I know you don't expect `Annot` to be `nullptr`, but if it were, should we bail?

lib/Sema/SemaExpr.cpp
```diff
@@ -8058,6 +8058,112 @@ Sema::CheckSingleAssignmentConstraints(QualType LHSType, ExprResult &CallerRHS,
   return result;
 }
 
+/// If a declaration has a Checked C bounds-safe interface
+/// attached to it, return the checked pointer type for the interface.
+/// Otherwise return a null QualType.
+///
+/// This falls into two cases:
```
reubeno: Feel free to tell me otherwise, but I'm wondering if this logic belongs here, or if it belongs in helpers on the declaration objects themselves?  Perhaps I need to get used to the world of compilers--I'm used to going with stronger encapsulation/containment.

lib/Sema/SemaExpr.cpp
dtarditi: Yes, at some point, we'll have code that may need one or more bounds expressions and a bounds type annotation.   I agree that it makes sense to  prepare for that sooner rather than later.  Perhaps we should have separate accessors for the bounds expression and the bounds interface type annotation bounds type annotation, so that we can hid the concrete representation better?   This seems like something that shouldn't be part of this change, but should be done soon.

lib/Sema/SemaExpr.cpp
dtarditi: No, there was a break statement missing.

PR #73, opened 2016-10-12
Allow type qualifiers and static keyword in parameter type annotations.

dtarditi:
This change extends the parsing of interop type annotations for parameters.  The code now allows the static keyword or type qualifiers to occur inside the first dimension of an array type in an interop type annotation, just as they can occur for regular parameter types.

This required adding the declarator as an argument to some parsing methods for interop type annotations, so that the code can tell when it is in the special context of parsing an interop type   annotation for a parameter declaration.

The code for parsing type names allows a Context value to be passed in.  It can fail an assert in SemaType.cpp if a Context other than TypeName is passed in, though.  Other contexts can allow variable names to occur within the parsing of  a type name.  Modify the assert in SemaType.cpp to apply to only the TypeName context and issue an error message if an identifier is used in other contexts.  I weighed adding a new context for parsing interop bounds expression, but that change seem worse, requiring fixing up a bunch of places in the code.  It seemed better to generalize parsing of type names slightly instead.

This fixes a few places where existing code for Checked C exceeded the 80 character line limit for the LLVM coding convention.  The parameter IsReturn is intentionally unused.  This change was extracted from a larger commit that uses it.

Testing:
- Passes clang regression tests.
- Passes existing Checked C regression tests.
- Added new tests for parsing interop bounds declarations with array types to the Checked C GitHub repo.  The tests will be added to tests/parsing/interop_types.c.  They will be committed separately.

### DIFF
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -8790,4 +8790,7 @@ def err_typecheck_non_count_bounds_decl : Error<
 def err_typecheck_non_count_return_bounds : Error<
   "expected %0 to have a pointer or integer return type">;
 
+def err_typecheck_bounds_type_annotation_identifier: Error<
+  "type name cannot have identifier in it">;
+
 } // end of sema component.
--- a/lib/Parse/ParseDecl.cpp
+++ b/lib/Parse/ParseDecl.cpp
@@ -5966,7 +5966,8 @@ void Parser::ParseFunctionDeclarator(Declarator &D,
   if (getLangOpts().CheckedC && Tok.is(tok::colon)) {
     BoundsColonLoc = Tok.getLocation();
     ConsumeToken();
-    ExprResult BoundsExprResult = ParseBoundsExpressionOrInteropType();
+    ExprResult BoundsExprResult =
+      ParseBoundsExpressionOrInteropType(D,/*IsReturn=*/true);
     if (BoundsExprResult.isInvalid())
       // We don't have enough context to try to do syntactic error recovery
       // here.  It is done instead in Parser::ParseDeclGroup, which recognizes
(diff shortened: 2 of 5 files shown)


lib/Parse/ParseDecl.cpp
```diff
@@ -5964,7 +5966,8 @@ void Parser::ParseFunctionDeclarator(Declarator &D,
   if (getLangOpts().CheckedC && Tok.is(tok::colon)) {
     BoundsColonLoc = Tok.getLocation();
     ConsumeToken();
-    ExprResult BoundsExprResult = ParseBoundsExpressionOrInteropType();
+    ExprResult BoundsExprResult =
+      ParseBoundsExpressionOrInteropType(D,/*IsReturn=*/true);
```
reubeno: Nit: should there be white space between the `,` and the second argument?

lib/Parse/ParseExpr.cpp
```diff
@@ -2778,7 +2778,14 @@ ExprResult Parser::ParseInteropTypeAnnotation() {
     if (PT.expectAndConsume(diag::err_expected_lparen_after,
                             Ident->getNameStart()))
       return ExprError();
-    TypeResult Ty = ParseTypeName();
+    // Interop type annotations for parameter declarators must be parsed in a
+    // special context.  Array types may have static or a type qualifier
+    // declared as part of the first dimension, that is,
+    //     int a[static 10] : itype(int [static 10])
```
reubeno: I'm probably missing the connection here; is it that `static` or other type qualifiers are only present in the `Declarator`'s context because they're stripped out otherwise?

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8790,4 +8790,7 @@ def err_typecheck_non_count_bounds_decl : Error<
 def err_typecheck_non_count_return_bounds : Error<
   "expected %0 to have a pointer or integer return type">;
 
+def err_typecheck_bounds_type_annotation_identifier: Error<
+  "type name cannot have identifier in it">;
```
dtarditi: clang points to the problematic type name as part of the error message.    The ability to point at the problematic code is a pretty nice feature of clang's error reporting.

Issue #64, opened 2016-10-03
Validate type in bounds-safe interface type annotation


dtarditi:
The compiler needs to validate the type of a bounds-safe type annotation relative to the declared type of the variable, member, or function return value:

- The annotation type must be a checked pointer type.
- It must be identical to the declared type if all occurrences of checked pointer types are replaced with unchecked pointer types.
- The referent type of the annotation type must be ``at least as checked'' as the referent type of the declared type.

dtarditi:
This work is complete.

PR #74, opened 2016-10-14
Check bounds-safe interface type annotations.

dtarditi:
This change adds checking of bounds-safe interface type annotations.  It addresses issue #64.

The type in the bounds-safe interface type annotation is called the bounds-safe interface type.  For a bounds-safe interface type annotation, we require that:
1. The variable to which the annotation is attached have a pointer or array type.  For annotations on return values, the return type must be a pointer type.
2. The  bounds-safe interface type be a pointer or array type.
3. The bounds-safe interface type be a checked type.  A checked type defined inductively.  A checked type is (a) a checked pointer or array type, (b) a pointer type whose referent type is a checked type, or (c) an array type whose element type is a referent type, or (d) a function type whose return type is a checked type or that has a parameter whose type is a checked type.
4.  The declared type and the bounds-safe interface type be compatible types when the checked property is ignored on pointer types and array types.
5.  The bounds-safe interface type be at least as checked as the original  declared type.

This change adds new methods for types to compute properties 4 and 5.   There is already a method for computing property 3.  The code for checking bounds annotation is added at the point where a bounds-safe interface type is attached to a declaration.   

It took a bit of effort to come up with accurate and helpful error messages. I decided to have more general messages that gloss over some of the subtle points.  For example, for property 4, the error message says that there is a mismatch between the interface and declaration type, and doesn't add the subtlety that it is compatibility modulo checkedness.  For property 3, the error message says that the the type must be checked.    The message points at the problematic interface type.  For simple cases, such as omitting the checked modifier in an array type or not using a checked type, this will help the user.  We actually allow types that are mix of checked and unchecked pointers, but that subtlety is omitted.  When there are nested levels of pointers, we don't know which pointers are intended to be checked.  Also, declarations may have multiple problems.  An interface type might not be a checked type and it might not be compatible.   For now, we just point out what seems to be the broadest problem, instead of emitting multiple diagnostics.

The code in lib/Parse/ParseDecl.cpp for setting an interop type annotation for a structure/union member just set the bounds expression on the FieldDecl.  It did not invoke ActOnBoundsDecl, which caused it to miss new checking adding to ActOnBoundsDecl.  The code for other bounds expressions did invoke ActOnBoundsDecl. (There are two separate code paths because bounds interop types are easily parsed and bounds expressions are lazily parsed.)

Testing:
- An extensive set of new feature tests will be added via a separate commit to the Checked C github repo.
- Passed existing Checked C tests.
- Passed existing clang regression tests.


include/clang/AST/ASTContext.h
```diff
@@ -2333,6 +2320,35 @@ class ASTContext : public RefCountedBase<ASTContext> {
...
 
   //===--------------------------------------------------------------------===//
+  //                    Predicates For Checked C checked types
+  //===--------------------------------------------------------------------===//
+
+  /// \brief Determine whether this pointer, array or function type
+  /// provides at least as much checking as the other type.  Return true if
+  /// it does or false if it does not or the types differ in some other way
+  /// than checkedness.
+  bool isAtLeastAsCheckedAs(QualType T1, QualType T2) const;
```
reubeno: It's great to see the function name implying the direction of the comparison between `T1` and `T2`, but I might suggest explicitly adding to the comment that you're checking that `T1` is at least as checked as `T2`.

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8793,4 +8793,32 @@ def err_typecheck_non_count_return_bounds : Error<
 def err_typecheck_bounds_type_annotation_identifier: Error<
   "type name cannot have identifier in it">;
 
+def err_typecheck_bounds_type_annotation_for_illegal_type : Error<
+  "interface type only allowed for a declaration with pointer "
```
reubeno: When we refer to "interface type" in the errors, do you think we should further qualify and instead indicate "Checked C interface type"?  Or something of that vein?  I'm considering the case of someone who tweaks code in an existing Checked C code base, isn't terribly familiar with Checked C, and sees this error -- wondering what an "interface type" is in C.

lib/AST/ASTContext.cpp
```diff
@@ -8191,6 +8191,225 @@ QualType ASTContext::mergeObjCGCQualifiers(QualType LHS, QualType RHS) {
...
+  T2 = T2.getCanonicalType();
+
+  const Type *T1Type = T1.getTypePtr();
+  const Type *T2Type = T2.getTypePtr();
+
+  if (T1Type == T2Type)
+    return true;
+
+  Type::TypeClass T1TypeClass = T1->getTypeClass();
+  if (T1TypeClass != T2Type->getTypeClass())
```
reubeno: Would this, for example, consider `IncompleteArray` types and `ConstantArray` types separate?  Do we not consider `int [3]` to be at least as checked as `int[]`?

lib/AST/ASTContext.cpp
```diff
@@ -8191,6 +8191,225 @@ QualType ASTContext::mergeObjCGCQualifiers(QualType LHS, QualType RHS) {
...
+  case Type::IncompleteArray: {
+    // check common array properties
+    const ArrayType *T1ArrayType = cast<ArrayType>(T1Type);
+    const ArrayType *T2ArrayType = cast<ArrayType>(T2Type);
+    if (T1ArrayType->getSizeModifier() != T2ArrayType->getSizeModifier())
+      return false;
+
+    if (lessThan(T1ArrayType->isChecked(),T2ArrayType->isChecked()))
+      return false;
+
```
reubeno: Minor nit: is the extra newline here intended?

lib/AST/ASTContext.cpp
```diff
@@ -8191,6 +8191,225 @@ QualType ASTContext::mergeObjCGCQualifiers(QualType LHS, QualType RHS) {
...
+  case Type::FunctionProto: {
+    const FunctionProtoType *T1FuncType =
+      cast<FunctionProtoType>(T1Type);
+    const FunctionProtoType *T2FuncType =
+      cast<FunctionProtoType>(T2Type);
+
+    // Check return types
+    QualType T1ReturnType = T1FuncType->getReturnType();
+    QualType T2ReturnType = T2FuncType->getReturnType();
+    if (!isAtLeastAsCheckedAs(T2ReturnType, T1ReturnType))
```
reubeno: Did you mean `isAtLeastAsCheckedAs(T1ReturnType, T2ReturnType)`?  This seems inconsistent with the checks above.  (Is there a test case covering this?)

lib/AST/ASTContext.cpp
```diff
@@ -8191,6 +8191,225 @@ QualType ASTContext::mergeObjCGCQualifiers(QualType LHS, QualType RHS) {
...
+    unsigned int paramCount = T1FuncType->getNumParams();
+    if (paramCount != T2FuncType->getNumParams())
+      return false;
+
+    for (unsigned int i = 0; i < paramCount; i++) {
+      QualType T1ParamType = T1FuncType->getParamType(i);
+      QualType T2ParamType = T2FuncType->getParamType(i);
+      if (!isAtLeastAsCheckedAs(T1ParamType, T2ParamType))
+        return false;
+      // check additional non-type information about parameters
```
reubeno: Naive question: what falls into this bucket?

lib/AST/ASTContext.cpp
```diff
@@ -8191,6 +8191,225 @@ QualType ASTContext::mergeObjCGCQualifiers(QualType LHS, QualType RHS) {
...
+      const ConstantArrayType *T2ConstantArrayType =
+        cast<ConstantArrayType>(T2Type);
+      if (!llvm::APInt::isSameValue(T1ConstantArrayType->getSize(),
+          T2ConstantArrayType->getSize()))
+        return false;
+    }
+
+    return true;
+  }
+  case Type::FunctionProto: {
```
reubeno: Are we guaranteed to get a FunctionProto as opposed to the non-proto Function type?

lib/AST/ASTContext.cpp
```diff
@@ -8191,6 +8191,225 @@ QualType ASTContext::mergeObjCGCQualifiers(QualType LHS, QualType RHS) {
...
+    }
+
+    // Check T2 properties of the function type.
+    if (T1FuncType->getExtInfo() != T2FuncType->getExtInfo())
+      return false;
+
+    if (T1FuncType->isVariadic() != T2FuncType->isVariadic())
+      return false;
+
+    // TODO: if we extend Checked C to C++, check C++-specific properties of
```
reubeno: What about other C11 properties of functions, such as `_Noreturn` (it's a function specifier, I believe).  Is that covered in the `ExtInfo` above?  Or is that only found on the declarator, not in the type?

lib/AST/ASTContext.cpp
```diff
@@ -8191,6 +8191,225 @@ QualType ASTContext::mergeObjCGCQualifiers(QualType LHS, QualType RHS) {
...
 
+//===--------------------------------------------------------------------===//
+//                    Predicates For Checked C checked types
+//===--------------------------------------------------------------------===//
+
+static bool lessThan(bool Self, bool Other) {
+  return (Self != Other && !Self);
+}
+
+bool ASTContext::isAtLeastAsCheckedAs(QualType T1, QualType T2) const {
```
reubeno: Do we need to worry about paren types, or other odd things?

lib/AST/ASTContext.cpp
```diff
@@ -8191,6 +8191,225 @@ QualType ASTContext::mergeObjCGCQualifiers(QualType LHS, QualType RHS) {
...
+    // function types.
+
+    return true;
+  }
+  default:
+    return false;
+  }
+}
+
+bool ASTContext::isEqualIgnoringChecked(QualType T1, QualType T2) const {
```
reubeno: My questions for the above function generally apply here too :)

lib/Sema/SemaDecl.cpp
```diff
@@ -11213,28 +11213,107 @@ void Sema::ActOnFinishKNRParamDeclarations(Scope *S, Declarator &D,
...
+  if (DiagId) {
+    S.Diag(AnnotTyLoc, DiagId);
+    return;
+  }
+
+  // Check that the types are identical if checking is ignored.
+  if (!S.Context.isEqualIgnoringChecked(AnnotTy, DeclaredTy)) {
+    DiagId = diag::err_typecheck_bounds_type_annotation_incompatible;
+    S.Diag(AnnotTyLoc, DiagId) << AnnotTy << DeclaredTy;
+    return;
```
reubeno: The pattern in this function is to return as soon as we find an error.  Would it make more sense to generate all relevant errors before returning?  For example, if an annotated type is neither checked nor equivalent, should we report both issues at the same time?

lib/AST/ASTContext.cpp
```diff
@@ -8191,6 +8191,225 @@ QualType ASTContext::mergeObjCGCQualifiers(QualType LHS, QualType RHS) {
...
+  if (T1TypeClass != T2Type->getTypeClass())
+    return false;
+
+  switch (T1TypeClass) {
+  case Type::Pointer: {
+    const PointerType *T1PtrType = cast<PointerType>(T1Type);
+    const PointerType *T2PtrType = cast<PointerType>(T2Type);
+    QualType T1PointeeType = T1PtrType->getPointeeType();
+    QualType T2PointeeType = T2PtrType->getPointeeType();
+    if (!isEqualIgnoringChecked(T1PointeeType, T2PointeeType)) {
```
reubeno: Small nit: it looks like the code is slightly inconsistent on whether or not to place curly braces around single-statement bodies of conditional statements.  I know the existing code base may not be super-consistent, but we should probably pick one approach or the other and stick to it?

include/clang/Basic/DiagnosticSemaKinds.td
dtarditi: I have the same concern as you do.   The clang developer documentation urges you to keep error messages under 80 characters and to not repeat information or provide unnecessary information.    I'm trying to stay in-line with that.  In this case, the error message will point directly at the itype declaration.  I hope this will give the user the context needed to understand this.

Here an example of an error message:
```
interop_type_annotations.c:78:18: error: interface type only allowed for a declaration with pointer or array type
void f1a(int p : itype(ptr<int>)) {   // expected-error {{interface type only allowed for a declaration with pointer or array type}}
                 ^
```
I could change the error message to use `bounds-safe interface type` instead of just `interface type`.  What are your thoughts?

lib/AST/ASTContext.cpp
```diff
@@ -8191,6 +8191,225 @@ QualType ASTContext::mergeObjCGCQualifiers(QualType LHS, QualType RHS) {
...
+//                    Predicates For Checked C checked types
+//===--------------------------------------------------------------------===//
+
+static bool lessThan(bool Self, bool Other) {
+  return (Self != Other && !Self);
+}
+
+bool ASTContext::isAtLeastAsCheckedAs(QualType T1, QualType T2) const {
+  if (T1.isNull() || T2.isNull())
+    return true;
```
dtarditi: You can have an "empty" QualType that has a null Type * associated with it.   We do need to bail out if we see that (the varaibles at lines 8212 or 8213 might end up with null pointers).    I think I should return false in this case (a null type is never at least as checked as anything, because we know nothing about it).

lib/AST/ASTContext.cpp
dtarditi: We're working with canonical types, which should strip out the parens.

include/clang/AST/ASTContext.h
dtarditi: Good suggestion - I'll do that.

lib/AST/ASTContext.cpp
dtarditi: We're guarantted that both T1 and T2 will be FunctionProto at this point.   A non-proto Function type ends up with a distinct type (which is distinct from FunctionProto).

lib/AST/ASTContext.cpp
dtarditi: Yes, this is a bug.   I'll fix this and add some more tests to cover checkedness on function types.

lib/AST/ASTContext.cpp
dtarditi: I looked at this further - apparently nothing for C.     This looks like an Apple-specific extension (it has some stuff specific to Objective-C and Swift.

lib/AST/ASTContext.cpp
dtarditi: Yes, _Noreturn is covered by the ExtInfo on the function type, as is calling convention informaton..   As far as I can tell, the information attached to function return types that isn't being checked here is all C++-specific.

lib/AST/ASTContext.cpp
dtarditi: Agreed.   The clang code generally doesn't parenthesize single statement bodies of conditional statements.   We should try to stick to that, which means removing the brackets in this case. 

  I am cautious if the single-statement actually spans multiple lines.  For example, I try to always bracket a nested if-statement to avoid falling afoul of the problems with dangling else statements

lib/Sema/SemaDecl.cpp
dtarditi: Yes, it makes sense to make a "best effort" to generate multiple distinct errors at once.    This could be a little tircky beause we need to void generating errors that are implied by other errors, though.  For example, if a bounds-afe interface type isn't an array or pointer type, there's no point in complaining about it being unchecked.

dtarditi:
I've pushed a new commit that addresses the code review feedback.    The error in isAtLeastAsCheckedAs and the unclear comments were the result of moving the method from Type.cpp to ASTContext.cpp - it turns from an instance method on types that take 1 argument to an instance method on ASTContext that took both types as arguments.  I had  to rename all the variables when I did this.

Let me know if you have any additional concerns/questions.

lib/Sema/SemaDecl.cpp
```diff
@@ -11213,28 +11213,129 @@ void Sema::ActOnFinishKNRParamDeclarations(Scope *S, Declarator &D,
...
+
+  int DiagId = 0;
+
+  // Make a best effort to detect multiple errors at once.  This might help
+  // programmers avoid multiple rounds of fixing and recompiling programs.
+  //
+  // Track the kinds of errors that have occurred to avoid emitting errors
+  // implied by prior errors, using a bitmask.
+
+  enum ErrorType {
```
reubeno: Do we think the clang code base is okay with type definitions local to a function body?  It certainly makes it appropriately private to the right scope.

dtarditi: Yes, there are other definitions of enum within function bodies in the clang code base.  An example is Sema::DiagnoseSentinelCalls in SemaExpr.cpp at line 416.

Issue #75, opened 2016-10-31
enforce Checked C rules limiting passing of checked types to no prototype functions


dtarditi:
At function calls, we need to enforce rules that checked types are not passed to functions with no prototypes.

Issue #30, opened 2016-08-12
Handle redeclarations of functions and variables with bounds


dtarditi:
For interoperation, we will redeclare existing library functions with additional bounds information.   We assume that modifying library source code is not possible when modifying programs that use libraries.  We'll have to add bounds information separately.    C allows functions and extern variable to be redeclared, so long as the redeclared versions are compatible with the existing version.

We need to extend the code for checking redeclaration to take bounds into account.  For parameters or variables of array_ptr type, bounds will have to match.  For now, we will use syntactic equality.  If they have unchecked pointer type, if they both of have bounds, they must match.  Otherwise,  the bounds will be regard as being part of a more "complete" declaration of the parameter or variable.

dtarditi:
Checking redeclarations of functions was completed by pull request #91

dtarditi:
This work is complete.

PR #76, opened 2016-10-31
Restrictions on redeclarations of no prototype functions.

dtarditi:
Calls to no-prototype functions are not type checked according to the C specification (Section 6.5.2.2 of the C11 specification).  The lack of type checking allows the checking of bounds declarations to be bypassed.   For the Checked C extension, this is addressed by not allowing redeclarations of no-prototype functions to have prototypes that use checked types, as well as not allowing checked values to be passed to no-prototype functions.  Section 5.5 of the Checked C extension specification describes the restrictions.  

This change implements the Checked C restrictions on redeclarations of no-prototype functions.  It is  part of work on issue #30.    The restrictions are implemented by changing the implementation of C type compatibility in clang.   We make no-prototype function types incompatible with function types that take arguments or return values that have checked types, contain values with checked types embedded within them, or are recursively pointers to function types that take or return check types.  We also change the construction of function types from function declarators to enforce the Checked C rule that a no-prototype function cannot have a checked return type.

When two declarations of a function are incompatible, we try to diagnose whether a failure is due to a Checked C type compatibility rule and print descriptive error messages.   We do this only when the failure  is because of the presence or absence of a prototype of a declaration.  We do not try to diagnose type compatibility problems involving parameters with function pointer types,  In that case, programmers will get a generic "types are not compatible" message.

There is a subtle complication when checking redeclarations in the presence of incomplete structure or union types.  In the Checked C rules, a function type with an incomplete type is compatible with a no-prototype function (because the incomplete type is unknown).  For a series of declarations of a function, clang merges the types so that the latest declaration has the merged type.  This means that a function could start as a no-prototype function, be given a prototype with an incomplete type, the incomplete type could be completed to be  checked type, and then the function could be declared again:
```
int f();      // decl 1
struct S;
int f(S v);   // decl 2
// complete the type
struct S {
  ptr<int> p;
}
...
int f(S v) {  // decl 3 (possibly a definition)
  ...
}
```
clang looks at only the type of the prior declaration when merging declaration types.  For Checked C, this misses the fact that `f` was declared originally  as a no-prototype function at `decl 1`.   The prototype was allowed only because it involved an incomplete type at `decl 2`.  If we had the complete type at `decl 2`, the prototype would not have been allowed.  Instead of looking at just the prior declaration, we look back at all prior declarations of a function to make sure they are compatible with the current declaration of the function.

Note that we still need to prevent `decl 2` from being used after `S` is complete.  The current change only addresses redeclarations of `f`, not uses of `f`.   That is  future work tracked by issue #75. 

In the implementation of type compatibility, we miss the case of no-prototype function pointers being incompatible with prototype function pointers that take integer arguments with bounds.  This depends on issue #20 (adding bounds to function types).  We will address this case as part of issue #20.

Testing:
- This change uncovered places in existing tests in the Checked C repo where we were using no prototype functions.  Update the tests to properly declare the functions as having no arguments (using the `f(void)` pattern).
- Add a new tests on to Checked C repo in typechecking\no_prototype_functions.c.
- Add some clang-specific tests under tests\CheckedC\typechecking.c.
- Passes the updated and new Checked C tests.
- Passes the existing clang regression tests.

### DIFF
--- a/include/clang/AST/ASTContext.h
+++ b/include/clang/AST/ASTContext.h
@@ -2335,6 +2335,10 @@ public:
   /// differ in some other way than checkedness.
   bool isEqualIgnoringChecked(QualType T1, QualType T2) const;
 
+  /// \brief Return true if this type is a checked type that is not
+  /// allowed to be passed or returned from a no prototype function.
+  bool isNotAllowedForNoProtoTypeFunction(QualType T1) const;
+
   // Methods to support checking assignments in the presence of
   // checked pointers.
 
--- a/include/clang/AST/ASTContext.h
+++ b/include/clang/AST/ASTContext.h
@@ -2337,7 +2337,7 @@ public:
 
   /// \brief Return true if this type is a checked type that is not
   /// allowed to be passed or returned from a no prototype function.
-  bool isNotAllowedForNoProtoTypeFunction(QualType T1) const;
+  bool isNotAllowedForNoPrototypeFunction(QualType T1) const;
 
   // Methods to support checking assignments in the presence of
   // checked pointers.
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -8822,20 +8822,23 @@ def err_bounds_type_annotation_lost_checking : Error<
   "loses checking of declared type}0,1">;
 
  def err_no_prototype_function_with_checked_return_type : Error<
-   "function without prototype cannot return a checked value">;
+   "function with no prototype cannot have a return type that is a "
+   "%select{checked type|structure with a member with a checked type|"
+   "union with a member with a checked type}0">;
 
  def err_no_prototype_function_with_return_bounds : Error<
-   "function without prototype cannot have a return bounds">;
+   "function with no prototype cannot have a return bounds">;
 
  def err_no_prototype_function_redeclared_with_checked_arg : Error<
-   "redeclaring a no-prototype function with checked argument is not allowed">;
+   "cannot redeclare a function with no prototype to have an argument type that is a "
+   "%select{checked type|structure with a member with a checked type|"
+   "union with a member with a checked type}0">;
 
  def err_no_prototype_function_redeclared_with_arg_bounds : Error<
-   "redeclaring a no-prototype function with an argument bounds is not"
-   " allowed">;
+   "cannot redeclare a function with no prototype to have an argument bounds">;
 
  def err_checkedc_incompatible_no_prototype_redeclaration : Error<
-   "redeclaring function that has a checked argument or argument "
-   "bounds as a no-prototype function is not allowed">;
+   "cannot redeclare a function that has a checked argument or argument "
+   "bounds to have no prototype">;
 
 } // end of sema component.
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -8828,8 +8828,14 @@ def err_bounds_type_annotation_lost_checking : Error<
    "function without prototype cannot have a return bounds">;
 
  def err_no_prototype_function_redeclared_with_checked_arg : Error<
-   "redeclaring a no-prototype function with checked argument not allowed">;
+   "redeclaring a no-prototype function with checked argument is not allowed">;
 
  def err_no_prototype_function_redeclared_with_arg_bounds : Error<
-   "redeclaring a no-prototype function with an argument bounds not allowed">;
+   "redeclaring a no-prototype function with an argument bounds is not"
+   " allowed">;
+
+ def err_checkedc_incompatible_no_prototype_redeclaration : Error<
+   "redeclaring function that has a checked argument or argument "
+   "bounds as a no-prototype function is not allowed">;
+
 } // end of sema component.
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -8821,4 +8821,15 @@ def err_bounds_type_annotation_lost_checking : Error<
   "type %diff{$ loses checking of declared type $|"
   "loses checking of declared type}0,1">;
 
+ def err_no_prototype_function_with_checked_return_type : Error<
+   "function without prototype cannot return a checked value">;
+
+ def err_no_prototype_function_with_return_bounds : Error<
+   "function without prototype cannot have a return bounds">;
+
+ def err_no_prototype_function_redeclared_with_checked_arg : Error<
+   "redeclaring a no-prototype function with checked argument not allowed">;
+
+ def err_no_prototype_function_redeclared_with_arg_bounds : Error<
+   "redeclaring a no-prototype function with an argument bounds not allowed">;
 } // end of sema component.
--- a/lib/AST/ASTContext.cpp
+++ b/lib/AST/ASTContext.cpp
@@ -7743,7 +7743,7 @@ QualType ASTContext::mergeFunctionTypes(QualType lhs, QualType rhs,
 
       // For Checked C, a no prototype function is not compatible
       // with a prototype with a checked argument.
-      if (isCheckedC && isNotAllowedForNoProtoTypeFunction(paramTy))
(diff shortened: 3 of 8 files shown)


lib/AST/ASTContext.cpp
```diff
@@ -7739,6 +7740,11 @@ QualType ASTContext::mergeFunctionTypes(QualType lhs, QualType rhs,
       if (paramTy->isPromotableIntegerType() ||
           getCanonicalType(paramTy).getUnqualifiedType() == FloatTy)
         return QualType();
+
+      // For Checked C, a no prototype function is not compatible
+      // with a prototype with a checked argument.
+      if (isCheckedC && isNotAllowedForNoProtoTypeFunction(paramTy))
```
reubeno: I might be missing something because I read through this too quickly, but when does this code path gets hit?  Because it's under an `if (proto)`, does that mean it's called in cases when we have a prototype?  If that's the case, I'm not sure I understand why we'd call `isNotAllowedForNoProtoTypeFunction` in cases when we have a yes-has-prototype function.

What am I missing?

lib/AST/ASTContext.cpp
```diff
@@ -8407,6 +8413,48 @@ bool ASTContext::isEqualIgnoringChecked(QualType T1, QualType T2) const {
...
+// specification.
+bool ASTContext::isNotAllowedForNoProtoTypeFunction(QualType QT) const {
+  if (const PointerType *PT = QT->getAs<PointerType>()) {
+    if (PT->isChecked())
+      return true;
+    if (PT->isFunctionPointerType())
+      return isNotAllowedForNoProtoTypeFunction(PT->getPointeeType());
+    // Unchecked pointer types are allowed
+    return false;
+  } else if (QT->isCheckedArrayType())
```
reubeno: Can you have an unchecked array of function pointers, where the function pointers have params or return types that are checked?

I get nervous seeing this sort of logic so specific to a scenario (dealing with no-prototype functions).  It might be simpler and easier to maintain to have a single helper method somewhere in the system that, once and for all, determines if a given type--in any way--contains any sort of checked types or bounds expressions.  And perhaps we can implement that using a proper AST walker?

And feel free to tell me my anxiety about this is unnecessary right now.

lib/AST/ASTContext.cpp
```diff
@@ -8407,6 +8413,48 @@ bool ASTContext::isEqualIgnoringChecked(QualType T1, QualType T2) const {
...
+// For the Checked C extension, compute whether a type is allowed to be an
+// argument or return type for a no-prototype function.   This computes the
+// set of allowed types described in Section 5.5 of the Checked C
+// specification.
+bool ASTContext::isNotAllowedForNoProtoTypeFunction(QualType QT) const {
+  if (const PointerType *PT = QT->getAs<PointerType>()) {
+    if (PT->isChecked())
+      return true;
+    if (PT->isFunctionPointerType())
+      return isNotAllowedForNoProtoTypeFunction(PT->getPointeeType());
```
reubeno: Does `getPointeeType()` take care of desugaring, canonicalizing the inner type?

lib/Sema/SemaDecl.cpp
```diff
@@ -3318,6 +3328,113 @@ void Sema::mergeObjCMethodDecls(ObjCMethodDecl *newMethod,
   CheckObjCMethodOverride(newMethod, oldMethod);
 }
 
+/// \brief Diagnose Checked C-specific compatibility issues for function decls.
+/// Handle cases where one declaration  has no prototype and the other
+/// one has a prototype that uses a checked type or has a bounds interface
+/// (Checked C compatibility rules are decribe in Section 5.5 of the Checked C
```
reubeno: Nit: typo in `decribe`.

lib/Sema/SemaDecl.cpp
```diff
@@ -3318,6 +3328,113 @@ void Sema::mergeObjCMethodDecls(ObjCMethodDecl *newMethod,
   CheckObjCMethodOverride(newMethod, oldMethod);
 }
 
+/// \brief Diagnose Checked C-specific compatibility issues for function decls.
+/// Handle cases where one declaration  has no prototype and the other
```
reubeno: Nit: extra whitespace char between `declaration` and `has`.

test/CheckedC/typechecking.c
```diff
@@ -0,0 +1,30 @@
...
+//
+// RUN: %clang_cc1 -verify -fcheckedc-extension %s
+
+char fn41() : count(5); // expected-error {{expected 'fn41' to have a pointer return type}} // expected-error {{function without prototype cannot have a return bounds}}
+
+// Prototype of a function followed by an old-style K&R definition
+// of the function. 
+
+// The Checked C specification does not allow no prototype functions
+// to return checked values. Techncally, the K& R style function
```
reubeno: Nit: typo in `Techncally`.

test/CheckedC/typechecking.c
```diff
@@ -0,0 +1,30 @@
...
+
+char fn41() : count(5); // expected-error {{expected 'fn41' to have a pointer return type}} // expected-error {{function without prototype cannot have a return bounds}}
+
+// Prototype of a function followed by an old-style K&R definition
+// of the function. 
+
+// The Checked C specification does not allow no prototype functions
+// to return checked values. Techncally, the K& R style function 
+// definition is a no prototype function, so we could say it is illegal.
+// However, clang enforce the prototype declration at the definition of
```
reubeno: Nit typos: `enforce  -> enforces`, `declration -> declaration`.

include/clang/AST/ASTContext.h
```diff
@@ -2335,6 +2335,10 @@ class ASTContext : public RefCountedBase<ASTContext> {
   /// differ in some other way than checkedness.
   bool isEqualIgnoringChecked(QualType T1, QualType T2) const;
 
+  /// \brief Return true if this type is a checked type that is not
+  /// allowed to be passed or returned from a no prototype function.
+  bool isNotAllowedForNoProtoTypeFunction(QualType T1) const;
```
dtarditi: I'll fix this to be prototype (all lower case).   The clang is inconsistent about the capitalization of prototype.

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8821,4 +8821,21 @@ def err_bounds_type_annotation_lost_checking : Error<
   "type %diff{$ loses checking of declared type $|"
   "loses checking of declared type}0,1">;
 
+ def err_no_prototype_function_with_checked_return_type : Error<
+   "function without prototype cannot return a checked value">;
```
dtarditi: Agreed.   Structs and unions make this somewhat difficult message to word: they aren't checked types, but they might contain checked types.  We might need two separate error message for this ultimately.

lib/AST/ASTContext.cpp
dtarditi: There's an if-statement at line 7566 that handles the case where `lproto` and `rproto` are both non-null.    The code under `if (proto)` handles the case where one of them is null (has no prototype) and the other is non-null (has a prototype).

lib/AST/ASTContext.cpp
dtarditi: The `'is*` and`getAs` methods used in the conditional branches take care of desugaring and canonicalizing types if necessary.   The `getAs` method is a clang idiom specific to the Type hierarchy.

lib/AST/ASTContext.cpp
dtarditi: Yes, you can have an unchecked array of function pointers where the function pointers have params or return types that are checked.    They are treated the same as an unchecked pointer type (allowed for no prototype functions and allowed for members of structures passed to no prototype functions.  It is clear in the code that the member isn't a checked.   There's an implicit conversion for uses of the array type to the unchecked pointer type.

You are right to get nervous - I was hesitant about this method about that about it before creating it.  Right now, the definitions are subtly different, so creating methods with generic functionality seemed more likely to cause confusion down the road than creating a very specific function that ties directly to text in the Checked C spec.

It's easy to diff the code later and see how many different concepts we ended up with and figure out whether we can/should unify the functionality.

Issue #82, opened 2016-11-22
Add bounds information for returns to function types.


dtarditi:

dtarditi:
Work is complete.

Issue #81, opened 2016-11-22
Add bounds information for parameters to function types


dtarditi:

dtarditi:
Work is complete.

PR #88, opened 2016-12-02
Add bounds expressions to function types

dtarditi:
Function types incorporate bounds information in Checked C.  This change extends the clang IR to represent bounds information for parameters and returns of function types.   This issues #81 and #82.

- The change adds another dynamically-sized array to the FunctionProtoType data structure that holds bounds for parameters.    If a parameter has no bounds, a null pointer is stored for the corresponding entry in the array. The array is not allocated if there are no bounds expressions for any of the parameters.
- It add a member to FunctionType to hold the return bounds expression. We could save space for FunctionType objects that do not have return bounds by storing the return bounds as part of the dynamically-allocated array for parameter bounds.  We do not do that for now. There's no pressing need to save space and it would make the implementation a little more complicated.
- In the Checked C specification, function type compatibility involves renaming parameters so that corresponding parameters have the same name.  This is awkward to do in a compiler, so instead we chose another representation for parameters in bounds expressions that occur in function types.  We represent parameters using their position in the argument list.  We create an expression PositionalParameterExpr that represents this abstraction of parameters.  To check if two parameters are the same, we just compare the positional parameter expressions by value (using the index and type).

We update AST dumping and AST serialization to handle the new information.  We also change the code for memoizing function types (implemented by functions named "Profile") to take into account the new information.    We use the clang meta-template approach to abstract bounds expressions for parameters of function types.  It only ends up being about 40 lines of code.

Update the Checked C implementation documentation to describe the IR choices and explain
why we chose to use PositionalParameterExpr instead of taking another approach.

While we add the information, we don't do much with it yet.  This change does not update type compatibility testing or the checking of redeclarations.

The memory management surrounding construction of function types is a little tricky and deserves some explanation.  The class ExtProtoInfo holds additional information about a function type being constructed.  We add the bounds information for parameters there.  It includes pointers to arrays that are allocated during function type construction and deallocated after function type construction.   The FunctionType class is allocated with extra space at the end for a number of variably-sized arrays (side note: this could be a lot cleaner and safer when Checked C is extended to handle variable-sized array members).   The array contents are copied into the extra space.   The bounds expressions themselves are allocated in a memory region associated with the AST context.  They live as long as the entire AST.     Later on, we'll need to be careful to not allocate bounds expressions to live forever.   However, for the purposes of constructing the IR, this is fine.

Testing:
- Add new tests to AST dumping that test function types with bounds expressions, making sure
  that parameters have been abstracted properly.
- Passes existing Checked C regression tests.
- Passes clang regression tests.

### DIFF
--- a/docs/checkedc/Implementation-Notes.md
+++ b/docs/checkedc/Implementation-Notes.md
@@ -21,7 +21,120 @@ to handle checked extensions at this time. LLVM/clang will only allow
 the Checked C extension to be used with C and will not allow it to
 be used with C++, Objective C, or OpenCL.
 
-## Processing extensions to the IR
+## IR Extensions
+
+### Types
+
+Pointer types and array types are extended with information about whether the
+types are checked.  For pointer types, an enum is used to represent the
+different kinds of pointers.  For array types, a boolean flag is used to
+represent whether the array type is checked or not.  This information is used
+when determining type compatibility: two pointers types are compatible only
+when their kinds are the same and two array types are compatible only when
+they have the same checked property.
+
+Function types are extended with optional bounds expressions for parameters
+and the return value.  For parameters, the bounds expressions are recorded
+in a variably-sized array allocated within the function type.  The array
+is not allocated if there are no bounds expressions on parameters.
+
+Bounds expressions are canonicalized for function types. References to
+parameters are abstracted to the index of the parameter in the argument
+list, so that names of arguments do not matter.
+
+### Bounds expressions
+
+Bounds expressions are represented using subclasses of the Expr
+class.  This allows the usual machinery for expressions to be
+reused.  There are three kinds of bounds expressions:
+
+- Nullary bounds expressions.  These do not refer to any other
+  expressions.  There are two kinds of nullary bounds expressions:
+  None and Invalid.  Invalid represents bounds expressions that
+  are invalid for some reason.  They are used to prevent errors
+  from cascading during semantic processing.  Just ignoring an invalid
+  bounds expressions could trigger downstream semantic checking errors.
+- Count bounds expressions, which refer to a single expression
+  that is the count.  There are two kinds of count bounds expressions:
+  element count and byte counts.
+- Range bounds expressions, which refer to lower and upper bound
+  expressions.
+
+There is an abtract base class `BoundsExpr` for bounds expressions.
+A unified kind is stored on this base class.
+
+Type annotations are also represented as a subclass of `BoundsExpr`.
+They can appear syntactically where bounds expressions can apppear.
+The type for the type annotation is stored as the type of the
+type annotation expression.
+
+### Other expressions
+
+The `PositionalParameterExpr` class represents a reference to a parameter
+that has been abstracted to the index of the parameter in an argument list.
+This is used in the represention of function types with bounds expressions:
+the bounds expressions are modified to use `PositionalParameterExpr`
+instead of `DeclRefExpr`s that are pointing to `ParmVarDecl`.
+
+For example, given
+
+```
+int f(_Array_ptr<int> arr : bounds(arr, arr + len), int len);
+```
+
+`arr` is given the positional index 0 and `len` is given the positional
+index `1.
+
+This makes type canonicalization of function types with bounds expressions
+straightforward.  The comparion of parameters is easy: two parameters are the
+same if their positional parameter expressions have the same index and type.
+The current type canonicalization algorithm is used, and it is extended to
+check that each bounds expression is identical.  Each bounds expression is
+structurally compared, including any subexpressions of the bounds expression.
+
+We could have modified the type canonicalization algorithm to handle
+canonicalization of parameters specially for the case of function bounds
+expressions. Parameter variables do have the positional index recorded.
+However, this would make the IR confusing to understand. Programmers would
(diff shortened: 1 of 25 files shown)


lib/Serialization/ASTWriter.cpp
```diff
@@ -286,22 +286,28 @@ void ASTTypeWriter::VisitFunctionProtoType(const FunctionProtoType *T) {
...
+
   if (T->hasExtParameterInfos()) {
     for (unsigned I = 0, N = T->getNumParams(); I != N; ++I)
       Record.push_back(T->getExtParameterInfo(I).getOpaqueValue());
   }
 
-  if (T->isVariadic() || T->hasTrailingReturn() || T->getTypeQuals() ||
-      T->getRefQualifier() || T->getExceptionSpecType() != EST_None ||
-      T->hasExtParameterInfos())
+  if (T->isVariadic() || T->hasTrailingReturn() || T->hasParamBounds() ||
```
reubeno: I'm a bit lost; what does 'AbbrevToUse' mean / how is it used?

lib/Serialization/ASTWriter.cpp
```diff
@@ -286,22 +286,28 @@ void ASTTypeWriter::VisitFunctionProtoType(const FunctionProtoType *T) {
...
+      T->hasReturnBounds() || T->getTypeQuals() ||T->getRefQualifier() ||
```
reubeno: Minor nit: missing space between `||` and `T->getRefQualifier()`.

docs/checkedc/Implementation-Notes.md
```diff
@@ -21,7 +21,120 @@ to handle checked extensions at this time. LLVM/clang will only allow
...
+A unified kind is stored on this base class.
+
+Type annotations are also represented as a subclass of `BoundsExpr`.
+They can appear syntactically where bounds expressions can apppear.
+The type for the type annotation is stored as the type of the
+type annotation expression.
+
+### Other expressions
+
+The `PositionalParameterExpr` class represents a reference to a parameter
```
dtarditi: References to variables in expressions done using DeclRefExpr, which points to the declaration of the variable.  We replace DeclRefExpr with PositionalParameterExpr, so that too needs to be an expression.

include/clang/AST/Type.h
```diff
@@ -3281,9 +3286,18 @@ class FunctionProtoType : public FunctionType, public llvm::FoldingSetNode {
   /// Whether this function has a trailing return type.
   unsigned HasTrailingReturn : 1;
 
+  /// Whether this function has bounds information for parameters.
+  unsigned HasParamBounds : 1;
```
dtarditi: We can see one without other.  I've added a comment explain the invariant.  We rely on ReturnBounds to be null to indicate that there is no return bounds for a function.

include/clang/AST/Type.h
```diff
@@ -3357,6 +3379,8 @@ class FunctionProtoType : public FunctionType, public llvm::FoldingSetNode {
     }
     if (hasExtParameterInfos())
       EPI.ExtParameterInfos = getExtParameterInfosBuffer();
+    EPI.ParamBounds = hasParamBounds() ? param_bounds_begin() : nullptr;
+    EPI.ReturnBounds = getReturnBounds();
```
dtarditi: I'll update this for consistency.  If we ever change the representation of ReturnBounds, we'll want this pattern.

lib/Serialization/ASTWriter.cpp
```diff
@@ -286,22 +286,28 @@ void ASTTypeWriter::VisitFunctionProtoType(const FunctionProtoType *T) {
...
   Record.push_back(static_cast<unsigned>(T->getRefQualifier()));
   addExceptionSpec(T, Record);
+  Record.AddStmt(const_cast<BoundsExpr *>(T->getReturnBounds()));
 
   Record.push_back(T->getNumParams());
   for (unsigned I = 0, N = T->getNumParams(); I != N; ++I)
     Record.AddTypeRef(T->getParamType(I));
 
+  if (T->hasParamBounds())
+    for (unsigned I = 0, N = T->getNumParams(); I != N; ++I)
```
dtarditi: I followed the coding style used in the prior loop over params.   I think this is an example of an overzealous compiler writer following a coding style that _might_ save a cycle or two.   I agree it is odd, but I'll keep consistency with the other loop.

lib/Serialization/ASTWriter.cpp
dtarditi: Yes, this is confusing.  AbbrevToUse indicates whether something should be compressed while a the record out.  There are special "domain-specific" compressions wired into the record writer.   For example, there are domain-specific compressions for types that are existing types with some qualifiers.   There is also one for function prototypes, but it looks it is not enabled.  ASTWriter::ASTWriteTypeAbbrevs  seems to create a template for a function prototype abbreviation.   The template omits a bunch of fields.  The code that you pointed to disables abbreviations for those fields.   Because I didn't add the new fields for Checked C to the template, I make sure that compression will be disabled for them too.

Somewhat confusing,  I couldn't find any place where the function prototype abbreviation is actually being used.  It looks like Abbrev is always set to 0 for function prototypes.  

I'll add a comment to the code explaining the if-statement.

PR #98, opened 2016-12-29
Changes to re-writer to support const qualification of pointers

awruef:
This allows the re-writer to support const qualification of pointer values at arbitrary levels. 

There are some ugly things:

 * I can't figure out how to get clang to give m the source location of a declaration that includes the qualifiers. I can either get the source location of the _entire_ declaration, or, I can get the location of the non-qualified portions of the type specification in the declaration. I've tried a few things, none of them work. I asked in `#llvm` on IRC, and that didn't work either. 
 * The higher-order refactoring put `PointerVariable` as a class that potentially contains multiple levels of pointers within it. There is then a `std::map` within the `PointerVariable` class now that maps the integer identifier of a pointer level to the qualification of that level.

There are also some unit tests for this feature.


dtarditi:
I looked into why you can't get the locations of qualified types.   This is a long-running question.  The source code in TypeLoc.h says:
```
/// Currently, we intentionally do not provide source location for
/// type qualifiers.
```
but doesn't explain why, which is probably why this question comes up periodically.    I found two relevant threads in the clang CFE archives:

http://lists.llvm.org/pipermail/cfe-dev/2013-March/028607.html
http://lists.llvm.org/pipermail/cfe-dev/2016-July/050169.html

The problem is that the notion of "source location" for a qualified type isn't well-defined for C.  C allows declaration specifiers to consist of a list of type specifiers, storage class specifiers, and type qualifiers, in any order.  The following declarations are legal C declarations
```
int static const x = 5;
const const int y = 6;
int const extern x = 7;
const int extern z = 8;
```

I checked the parsing for declaration specifiers and the parser does retain source information for the location of type qualifier (presumably the last or first occurrence of the type qualifier).  However, the information is not propagated into the AST,  It would be straightforward to propagate the information to the AST node.   For variable declarations, you'd do this in Sema::ActOnVariableDeclarator at around line 6414 in SemaDecl.cpp (in the current version).

If we wanted to preserve source information for types (such as spacing) after adding the missing information to the AST,, we could start by handling the cases where the type qualifiers appear adjacent to the type specifier (either before or after).    We'd have to make sure that there is only white space in the source text between the type qualifier and the type specifier.   I'm guessing that would cover most cases seen in practice.    (Coding conventions and style guides usually frown on declarations like 'const int extern y').

I'm not proposing that you do this as part of this change - just outlining how we could handle this in the future.

test/CheckedCRewriter/const_tests.c
```diff
@@ -0,0 +1,33 @@
+// Tests for Checked C rewriter tool.
+//
+// Checks properties of functions.
+//
+// RUN: checked-c-convert %s -- | FileCheck -match-full-lines %s
+// RUN: checked-c-convert %s -- | %clang_cc1 -verify -fcheckedc-extension -x c -
+// expected-no-diagnostics
+
```
awruef: Other type qualifiers would need to be supported independently. There's an enum that I populate for this, I'd need to add more cases and move information from the `QualType` into that enum. I can make that change for other type qualifiers as part of this PR.

dtarditi: I think it is fine to handle the other qualifiers in a separate PR.

tools/checked-c-convert/Constraints.h
```diff
@@ -421,7 +421,7 @@ class Constraints {
   // a client wants to examine the environment is untenable.
   ConstraintSet &getConstraints() { return constraints; }
   EnvironmentMap &getVariables() { return environment; }
-  bool solve(void);
+  std::pair<ConstraintSet, bool> solve(void);
```
awruef: Done.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -24,52 +24,67 @@ tyToStr(const Type *T) {
...
+
+	if (Ty->getAs<TypedefType>())
+		isTypedef = true;
+
+	while (Ty->isPointerType()) {
+		// Allocate a new constraint variable for this level of pointer.
+		vars.insert(K);
+		CS.getOrCreateVar(K);
+
+		// TODO: Save here if QTy is qualified or not into a map that
```
awruef: Done

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -100,15 +115,26 @@ PointerVariableConstraint::mkString(Constraints::EnvironmentMap &E) {
...
+          s = s + "const ";
+
       emittedBase = false;
       s = s + "_Ptr<";
 
       caratsToAdd++;
       break;
     case Atom::A_Arr:
     case Atom::A_Wild:
+      q = QualMap.find(V);
```
awruef: I think that what's intended here is that the `QualMap` is tracking the qualification of the pointers, while the qualification of the pointed values is tracked independently. I convinced myself once that this was the right thing to do but maybe it isn't.

awruef: Ah, I think that what should happen here is that the insertion of `const` based on `q` should occur after we print out everything, not before. If the value pointed to is `const` then that should be captured by `BaseType`. I've made that change.

awruef:
Okay! I've incorporated the stuff from #124 and the tests pass on my end.

Issue #99, opened 2017-01-10
AST Reader is missing a case for positional parameter expressions


dtarditi:
In fixing build warnings, I noticed that ASTReader.cpp is missing a case to handle positional parameter expressions.

lenary:
Closed by #101.

Issue #97, opened 2016-12-22
Fix compiler warnings in clang build.


dtarditi:
I'm building the Checked C version of clang on Linux and seeing some compiler warnings about ignored qualifiers in Type.h (line 3357).  There are additional warnings about unhandled switch cases.  We should fix these warnings.

awruef:
I've seen these too, I think that the unhandled switch cases are related to stuff added by Checked C, the names of the unhandled cases seemed to be names of Checked C classes added to clang.

dtarditi:
This is fixed.

PR #100, opened 2017-01-11
Fix some build warnings introduced by changes for Checked C

dtarditi:
This change fixes some warnings during the build of clang introduced by changes for Checked C.   This is for issue #97.  We decided to make bounds expressions be subclasses of expressions to re-use the template metaprogramming support for expressions.   This avoids having to introduce additional template meta-programming support for bounds expressions.  The cost is that it introduces unhandled cases for switch statements over expressions.

For cases where bounds expressions are not expected to occur, I've introduced `llvm_unreachable` statements.  For example, the static analyzer knows nothing about the AST changes that add bounds expressions, so I would expect bounds expressions to never occur there.   For cases where bounds expressions might reasonably occur, such as `HasSideEffects`, I've filled in the missing cases.   `HasSideEffects` apparently defines a side-effect as an assignment (mutation of state).  Bounds expressions must have non-modifying subexpressions, so they have no side-effect according to this definition.

Testing:

- Built and ran clang and Checked C tests on a Linux x64 Ubuntu box.   I did not observe any build warnings except for a missing case in ASTReader (issue #99).   Interestingly, 3 of the Checked C test cases failed because of environmental differences.   There are 32 vs. 64-bit pointer problems related to tests trying to test integers with bounds (https://github.com/Microsoft/checkedc/issues/92). There's also a missing header file problem.  These don't appear to be related to the changes that I made.
- Passed all Checked C tests for a Windowx x86 build.

### DIFF
--- a/lib/AST/Expr.cpp
+++ b/lib/AST/Expr.cpp
@@ -2860,6 +2860,13 @@ bool Expr::HasSideEffects(const ASTContext &Ctx,
   case ObjCAvailabilityCheckExprClass:
   case CXXUuidofExprClass:
   case OpaqueValueExprClass:
+  case PositionalParameterExprClass:
+    // Checked C bounds expressions are not allowed to have assignments
+    // embedded within them.
+  case CountBoundsExprClass:
+  case InteropTypeBoundsAnnotationClass:
+  case NullaryBoundsExprClass:
+  case RangeBoundsExprClass:
     // These never have a side-effect.
     return false;
 
--- a/lib/AST/Expr.cpp
+++ b/lib/AST/Expr.cpp
@@ -2860,12 +2860,12 @@ bool Expr::HasSideEffects(const ASTContext &Ctx,
   case ObjCAvailabilityCheckExprClass:
   case CXXUuidofExprClass:
   case OpaqueValueExprClass:
-  case PositionalParameterExprClass:
     // Checked C bounds expressions are not allowed to have assignments
     // embedded within them.
   case CountBoundsExprClass:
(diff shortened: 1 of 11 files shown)


lib/AST/Expr.cpp
```diff
@@ -2860,6 +2860,13 @@ bool Expr::HasSideEffects(const ASTContext &Ctx,
   case ObjCAvailabilityCheckExprClass:
   case CXXUuidofExprClass:
   case OpaqueValueExprClass:
+  case PositionalParameterExprClass:
```
lenary: nit: this could be grouped with the 4 bounds expressions after the comment, for clarity.

lib/AST/ExprConstant.cpp
```diff
@@ -9438,6 +9438,10 @@ static ICEDiag CheckICE(const Expr* E, const ASTContext &Ctx) {
   case Expr::CXXFoldExprClass:
   case Expr::CoawaitExprClass:
   case Expr::CoyieldExprClass:
+  case Expr::CountBoundsExprClass:
+  case Expr::InteropTypeBoundsAnnotationClass:
+  case Expr::NullaryBoundsExprClass:
+  case Expr::RangeBoundsExprClass:
```
lenary: nit: these cases are the same as the case for Expr::PositionalParameterExprClass, again, it could be merged so we keep the checked-c pieces together.

lenary:
Ok, my build just finished and all the checks passed.

lenary:
Let me re-build, turns out I didn't build the branch I thought I did.

PR #101, opened 2017-01-11
Adds deserialization of Positional Parameter Expressions

### LINKED ISSUE #3: test serialization/deserialization of clang IR with new checked types and new bounds expressions
We added code for serializing and deserializing clang IR with the new checked types.   This is necessary to get the code to compile.  However, the serialization and deserialization of IR with these types has no tests.  We should add tests before people start using compiler functionality that depends on this.

lenary:
This is a fix for #99. 

I'm going to write some tests to make sure we haven't broken PCH, which seems the main (only?) place where ASTs are deserialized. I guess the addition of tests will also close #3.


lenary:
Let me add some tests before we merge it.

dtarditi:
I think it is fine to commit this change and add the tests as a separate change.  We're no worse off than before (which is to say we don't have tests for other new serialization code) and slightly better off because one build warning is gone.   It is your call.

PR #110, opened 2017-01-18
Walk IR for static checking of bounds declarations.

dtarditi:
Add initial code for walking expressions and declarations for static checking of bounds declarations.  The checking code will be invoked by this walking code.

To test the walking code, check some simple properties of bounds for assignments and variables declarations.    If the left-hand side of an assignment has a bounds requirement, check that the right-hand side of the assignment has the right form to have an inferred bounds expression.   This is done by calling a stubbed out inference method that returns bounds(any) for expression forms that might have a bounds expression.  Do a similar check for initializers for variables declarations with bounds.  Also check that automatic variables that have bounds declarations or _Ptr type have initializers.

Add the 'Any` bounds to the IR for bounds expressions.

Testing:
- There will be a separate pull request for the GitHub Checked C repo that updates tests where initializers were missing.    The pull request will also add tests of error messages for missing initializers.
- Passes existing clang regression tests.

### DIFF
--- a/include/clang/AST/Expr.h
+++ b/include/clang/AST/Expr.h
@@ -4517,16 +4517,18 @@ public:
     Invalid = 0,
     // bounds(none)
     None = 1,
+    // bounds(any)
+    Any = 2,
     // count(e)
-    ElementCount = 2,
+    ElementCount = 3,
     // byte_count(e)
-    ByteCount = 3,
+    ByteCount = 4,
     // bounds(e1, e2)
-    Range = 4,
+    Range = 5,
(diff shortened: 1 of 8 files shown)


dtarditi:
Right, that is confusing.  We could use InferBoundsExpression or InferBoundsForExpression.

lenary:
I'd lean towards "InferBoundsForExpression" of those two, but maybe something like "BoundsInferenceSema" (showing it's a Sema-like bounds inferrer object) would be better.

Issue #106, opened 2017-01-14
Add serialization tests for additional types


dtarditi:
From issue #102:

As a follow up item, could you add additional tests for Ptr types, function pointer types, checked arrays, and structures with checked members that have bounds declarations? I would like to test that those types are serialized properly.

Issue #104 already covers structure types.

Issue #105, opened 2017-01-14
Add tests for Struct Member bounds information


lenary:
We're not currently sure if this is (de)serialized or not at the moment. We need to test for it, and there are probably other struct member tests required.

lenary:
@dtarditi this is the test I mentioned this afternoon.

dtarditi:
I think these are reasonably well-tested now.

PR #111, opened 2017-01-19
Adds (De)Serialization of Declaration Bounds

### LINKED ISSUE #104: Add AST Serialization for Struct Member Bounds Information
Struct Member bounds information is currently not serialized into the AST, we believe. We need some tests first to find out, and then we will need to add serialization/deserialization for this data.

lenary:
This pull request comprehensively revisits and expands the PCH tests, and adds (de)serialization of the bounds associated with global variables, function returns, and struct member fields. 

The only way at the moment to test that bounds declarations on global fields and functions are preserved is to redeclare the same function, which should error if you drop the bounds, and not error if you preserve the right bounds as in the precompiled header.

Bounds expressions on struct members are very hard to test, until we start inferring and solving bounds information. The test for `itype()` should work today, the others will become useful later.

As for serialization/deserialization. I've added the code to do so to the VisitDeclaratorDecl methods, because the bounds information is stored within that class (as opposed to its subclasses). Doing this alone caused an assertion failure in the bitcode writer, namely `"Invalid abbrev for record!", file ../llvm/include/llvm/Bitcode/BitstreamWriter.h, line 268`. (If this happens again, I hope searching turns up this ticket). I believe I have added the correct entries to the abbreviation system. The clang regression tests pass on 32-bit, I'm currently running them on 64-bit.

Closes #104. 

Relevant to #105 and #106 but I think the tests need more work before I can close those two.

### DIFF
--- a/lib/Serialization/ASTReaderDecl.cpp
+++ b/lib/Serialization/ASTReaderDecl.cpp
@@ -747,7 +747,7 @@ void ASTDeclReader::VisitDeclaratorDecl(DeclaratorDecl *DD) {
   VisitValueDecl(DD);
   DD->setInnerLocStart(ReadSourceLocation(Record, Idx));
 
-  if (Record[Idx++]) // hasBoundExpr
+  if (Record[Idx++]) // hasBoundsExpr
 	DD->setBoundsExpr(Reader.ReadBoundsExpr(F));
 
   if (Record[Idx++]) { // hasExtInfo
--- a/lib/Serialization/ASTReaderDecl.cpp
+++ b/lib/Serialization/ASTReaderDecl.cpp
@@ -786,7 +786,6 @@ void ASTDeclReader::VisitFunctionDecl(FunctionDecl *FD) {
   FD->setCachedLinkage(Linkage(Record[Idx++]));
   FD->EndRangeLoc = ReadSourceLocation(Record, Idx);
 
-
   switch ((FunctionDecl::TemplatedKind)Record[Idx++]) {
   case FunctionDecl::TK_NonTemplate:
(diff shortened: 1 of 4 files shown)


lenary:
Clang regression tests pass on 64-bit Windows.

lib/Serialization/ASTReaderDecl.cpp
```diff
@@ -746,6 +746,10 @@ void ASTDeclReader::VisitEnumConstantDecl(EnumConstantDecl *ECD) {
 void ASTDeclReader::VisitDeclaratorDecl(DeclaratorDecl *DD) {
   VisitValueDecl(DD);
   DD->setInnerLocStart(ReadSourceLocation(Record, Idx));
+
+  if (Record[Idx++]) // hasBoundExpr
```
dtarditi: One small typo; the comment should be hasBoundsExpr.

lenary:
Every time I look at the`Abv` lines, I become less convinced I know what's going on with them. I'm going to review this file again, and run a lot of clang tests.

My concern is that these seem to be defining a layout. In this line, I'm saying "there's a fixed-width field here, 1 bit wide, that says whether we have a bounds expression", I think. This is correct _only_ when `hasBoundsExpr()` is false. If we have a bounds expression, we have the field for `hasBoundsExpr()` followed immediately by a field containing the bounds expression in purpose, showing this is not fixed width. 

I think I have a more robust fix incoming, which we can always revisit later.

lenary:
Ok, my tests pass. What 9d4e5be does is disable the use of abbreviations if we have a bounds expression, meaning the abbreviations are always correct. This is alright for the moment, but may in future be something we want to optimize.

Issue #114, opened 2017-01-20
Add explicit _Dynamic_check builtin


lenary:
This is just to track the `dynamic_check` workitem that I'll be working on for the next few weeks.

- [x] add `_Dynamic_check` operator builtin (this shouldn't require extending the parser/lexer)
- [x] implement code generation for `_Dynamic_check` operator. This will be based on #109 
- [x] implement static analysis of `_Dynamic_check` operator
- [x] check for any other passes/analysis that need to know about the operator
- [x] implement lots of tests

lenary:
Closed by #115, #122. #128, #132, #135

PR #115, opened 2017-01-20
Add Checked C's _Dynamic_check as Clang Builtin

### LINKED ISSUE #109: Design the Control Flow for Dynamic Checks
We don't yet have a design for the control flow around dynamic checks (including the `dynamic_check` keyword)

This is going to be based on how clang implements C's `assert` (when `NDEBUG` is not defined) and `abort`, given that these are most similar to the intent of `dynamic_check`. Importantly, these should do the right things on Windows and Linux. 

This requires investigation as there may be good reasons why we can't do what `assert` and `abort` do.

lenary:
This is a big pull request for the `_Dynamic_check` functionality.

We implement it as a clang builtin, which means we get compiler hooks for code generation and semantic analysis without having to extend the lexer/parser. 

I have followed the design in #109 (A design I'm yet to document).

The largest omission at the moment is a set of comprehensive tests, and semantic analysis to make sure it's only called on non-modifying expressions. 

References #114.
Closes #109.


lenary:
This is not yet ready for merge, just to be clear.

lenary:
Closes #109

docs/checkedc/Dynamic-check.md
```diff
@@ -0,0 +1,64 @@
+# Dynamic Checks
```
dtarditi: Should probably call this 'Dynamic Check operator'

docs/checkedc/Dynamic-check.md
```diff
@@ -0,0 +1,64 @@
...
+In LLVM IR, the things to look for are `br` for branch, `icmp` for
+integer comparison, and `call` for function calls.
+
+```
+define void @f1(i32* %i) #0 {
+entry:
+  %i.addr = alloca i32*, align 4
+  store i32* %i, i32** %i.addr, align 4
+  %0 = load i32*, i32** %i.addr, align 4
+  %cmp = icmp ne i32* %0, null
```
dtarditi: Can we avoid the two compares here and just do a single compare?

docs/checkedc/Dynamic-check.md
```diff
@@ -0,0 +1,64 @@
...
+  %conv = zext i1 %cmp to i32
+  %_Dynamic_check_result = icmp ne i32 %conv, 0
+  br i1 %_Dynamic_check_result, label %_Dynamic_check_succeeded, label %_Dynamic_check_failed
+
+_Dynamic_check_succeeded:                         ; preds = %entry
```
dtarditi: It would be good for the labels to have the same form (for example, always have a number appended to them).

lib/CodeGen/CGExpr.cpp
```diff
@@ -2767,6 +2767,32 @@ llvm::CallInst *CodeGenFunction::EmitTrapCall(llvm::Intrinsic::ID IntrID) {
...
+  Builder.SetInsertPoint(DyCkFail);
+  llvm::CallInst *TrapCall = Builder.CreateCall(CGM.getIntrinsic(llvm::Intrinsic::trap));
+  TrapCall->setDoesNotReturn();
+  TrapCall->setDoesNotThrow();
+  Builder.CreateUnreachable();
+
+  Builder.SetInsertPoint(Begin);
+  Zero = Builder.getInt(llvm::APInt::getNullValue(CheckedWidth));
+  IsTrue = Builder.CreateICmpNE(Checked, Zero, "_Dynamic_check_result");
+  Builder.CreateCondBr(IsTrue, DyCkSuccess, DyCkFail);
```
dtarditi: Could this just use Checked instead, instead of comparing against Zero to produce an IsTrue value?

test/CheckedC/dynamic_check-code-gen.c
```diff
@@ -0,0 +1,48 @@
...
+  // CHECK-NEXT: unreachable
+}
+
+// Function with two dynamic checks
+void f2(void) {
+  // CHECK: void @f2()
+
+  int i;
+  _Dynamic_check(i != 3);
+  // CHECK: icmp ne
```
dtarditi: I am confused why this shows only one comparison.   Is the other comparison appearing in the output and just not being checked?  Or are my other comments about two comparisons happening wrong?

docs/checkedc/Dynamic-check.md
lenary: Sure, will update.

docs/checkedc/Dynamic-check.md
lenary: I have no control over this, I provide "_Dynamic_check_succeeded", and the LLVM Builder decides if it needs a number appended or not in order to keep labels unique.

dtarditi: I suggest appending numbers to the labels, just to avoid confusing people.   It doesn't have to be exactly what is produced by clang.

lenary: Sure, done.

test/CheckedC/dynamic_check-code-gen.c
lenary: I decided to only check for the first not-equals comparison, as a way of making sure our check was happening, and not look for the second one, as it felt too much of an implementation detail.

Turns out, there is a way to make sure we don't do two comparisons, buried deep in the code generator (the API surface area is huge), I'm now using it, changes incoming.

lib/CodeGen/CGExpr.cpp
lenary: Turns out that it could.

docs/checkedc/Dynamic-check.md
lenary: Turns out, yes.

dtarditi:
Looks good, thanks!

Issue #116, opened 2017-01-21
Infer bounds for constants, variables, and cast expressions.


dtarditi:
This work item is to infer bounds for constants, variables, and cast expressions.   In clang, casts are both explicit (in the source code) and implicit (inserted by the compiler).  The work item includes writing tests for the inference.   

The testing plan is to have a mode where the compiler walks over an AST and dumps the ASTs for inferred bounds expressions for assignments and declarations.   We'll check that the dumps have the right form.

dtarditi:
Implemented by pull request #119.

PR #119, opened 2017-01-25
Infer bounds for constant, variable, and cast expressions

dtarditi:
Infer bounds for expressions that are constants, variables, or casts.  This addresses issue #116.

To reflect the different kinds of semantics of C expressions,  there are now 3 functions for inferring bounds:
* RValueBounds: for an expression that produces an rvalue, infer bounds for the rvalue.
* LValueTargetBounds: for an expression that produces an lvalue, infer bounds for the target of the lvalue (that is, the bounds for the value stored in the object pointed to by the lvalue).
* LValueBounds, for an expression that produces an lvalue, infer bounds for the lvalue.  The bounds determine whether it is legal to access memory using the lvalue.

To test inference, add a flag to the compiler that dumps information about inferred bounds.  clang does not seem to have a good place to put such debug flags,   Following existing practice, I put the flag in LangOptions.def, where other similar debug dump flags exist.  The dump prints the AST for an assignment expression or declaration for which bounds are being inferred, the expected bounds, and the inferred bounds.

Add feature tests to the clang regression suite for inference of bounds for expressions that are integer or enumeration constants and expressions that are variables.    The tests incorporate tests of cast expressions out of necessity, but further testing is needed for cast expressions.   For expressions that are variables, add tests for variables with pointer type and array type.  The tests check that all the information is correct, including the AST for the expression/declaration being analyzed, the AST for the expected bounds, and the AST for the inferred bounds.  When we add additional tests, we will check less information about the expressions/declarations being analyzed and the expected bounds.

There was a failure in the Checked C PCH tests where the compiler incorrectly produced an error message that compound literal expressions being used as initializers had no bounds.   Fix this by adding one more case to LValueBounds.

Testing:
- Passes new feature tests.
- Passes existing Checked C feature tests.
- Pass clang regression tests.


lib/Sema/SemaBounds.cpp
```diff
@@ -198,43 +198,246 @@ namespace {
...
+      uint64_t Bits = I.getZExtValue();
+      unsigned Width = Context.getIntWidth(Context.UnsignedLongLongTy);
+      llvm::APInt ResultVal(Width, Bits);
+      IntegerLiteral *Lit = IntegerLiteral::Create(Context, ResultVal,
+                                                   Context.UnsignedLongLongTy,
+                                                   SourceLocation());
+      return Lit;
+    }
+
+    BoundsExpr *ExpandToRange(Expr *Base, BoundsExpr *B) {
```
dtarditi: Sure, I'll add more comments in my next commit.

PR #122, opened 2017-01-26
Adds Non-Modifying Expression check to Dynamic Checks

lenary:
Though we have not yet updated the specification, our intention is to require explicit `dynamic_check`s contain non--modifying expressions, in part so we can run them as and when we want to, without fear of side effects. This also partly conveys the expectation that these checks should be lightweight.

I have implemented this as a new `RecursiveASTVisitor` in `SemaBounds.cpp`. If this visitor finishes early, it shows we found something not allowed in a non-modifying expression. The error message tries to clarify what part of the expression exactly we don't like.

### DIFF
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -8877,7 +8877,7 @@ def err_bounds_type_annotation_lost_checking : Error<
     "automatic variable %0 with _Ptr type must have initializer">;
 
   def err_not_non_modifying_expr : Error<
-	"%select{assignment|increment|decrement|call|comma|volatile}0 expression not allowed "
-	"within non-modifying expression in dynamic_check">;
+	"%select{assignment|increment|decrement|call|volatile}0 expression not allowed "
+	"in dynamic check expression">;
 
 } // end of sema component.
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -8876,4 +8876,8 @@ def err_bounds_type_annotation_lost_checking : Error<
   def err_initializer_expected_for_ptr : Error<
     "automatic variable %0 with _Ptr type must have initializer">;
 
+  def err_not_non_modifying_expr : Error<
+	"%select{assignment|increment|decrement|call|comma|volatile}0 expression not allowed "
+	"within non-modifying expression in dynamic_check">;
+
 } // end of sema component.
--- a/lib/CodeGen/CGBuiltin.cpp
+++ b/lib/CodeGen/CGBuiltin.cpp
@@ -2426,6 +2426,9 @@ RValue CodeGenFunction::EmitBuiltinExpr(const FunctionDecl *FD,
     break;
   }
   case Builtin::BI_Dynamic_check: {
+    if (!getLangOpts().CheckedC)
+      break;
+
     const Expr *CheckExpr = E->getArg(0);
     Value *CheckVal = EvaluateExprAsBool(CheckExpr);
     EmitDynamicCheck(CheckVal);
--- a/lib/CodeGen/CGBuiltin.cpp
+++ b/lib/CodeGen/CGBuiltin.cpp
@@ -2426,6 +2426,9 @@ RValue CodeGenFunction::EmitBuiltinExpr(const FunctionDecl *FD,
     break;
   }
   case Builtin::BI_Dynamic_check: {
+    // This disables specific code generation for dynamic checks
+    // if Checked C is not enabled. Code Generation will fall-through
+    // to emitting a "Unknown Builtin" error.
     if (!getLangOpts().CheckedC)
       break;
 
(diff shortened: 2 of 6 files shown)


lib/Sema/SemaBounds.cpp
```diff
@@ -345,3 +345,87 @@ void Sema::CheckTopLevelBoundsDecls(VarDecl *D) {
...
+    // Comma Expressions are defined as modifying
+    bool VisitBinComma(BinaryOperator *E) {
+      addError(E, MEK_Comma);
+      return false;
+    }
+
+
+  private:
+    Sema &S;
+    Expr *Outer;
```
lenary: Ah, I was using this for better diagnostic notes, but it's no longer used. Commit incoming with it removed (once tests have passed)

lenary: Done, ready for a review.

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8876,4 +8876,8 @@ def err_bounds_type_annotation_lost_checking : Error<
   def err_initializer_expected_for_ptr : Error<
     "automatic variable %0 with _Ptr type must have initializer">;
 
+  def err_not_non_modifying_expr : Error<
+	"%select{assignment|increment|decrement|call|comma|volatile}0 expression not allowed "
+	"within non-modifying expression in dynamic_check">;
```
dtarditi: This should use the official name, which is "_Dynamic check".      We should not use the term "non-modifying expression" in an error message without some sort of clarification because it is not a well-known term.

lib/CodeGen/CGBuiltin.cpp
```diff
@@ -2426,6 +2426,9 @@ RValue CodeGenFunction::EmitBuiltinExpr(const FunctionDecl *FD,
     break;
   }
   case Builtin::BI_Dynamic_check: {
+    if (!getLangOpts().CheckedC)
+      break;
```
dtarditi: Can this ever happen?  We should (1) not recognize _Dynamic_check is an intrinsic if Checked C is turned off and (2) have an llvm_unreachable statement before the break, to trigger an internal compiler error if we get here.

lib/Sema/SemaBounds.cpp
```diff
@@ -345,3 +345,86 @@ void Sema::CheckTopLevelBoundsDecls(VarDecl *D) {
...
+      MEK_Volatile
+    };
+
+  public:
+    NonModifiyingExprSema(Sema &S) : S(S) {}
+
+    // Assignments are of course modifying
+    bool VisitBinAssign(BinaryOperator* E) {
+      addError(E, MEK_Assign);
+      return false;
```
dtarditi: We should not terminate the walk early if we find an error.   A design goal for the compiler is to find as many errors as possible in one compilation.  Compilations can be expensive and build systems can be slow, so this improves programmer productivity.    The only reason to cut off traversal would be if we find ourselves in a situation that would later generate spurious errors or in which we can make no further meaningful progress.

lib/Sema/SemaBounds.cpp
```diff
@@ -345,3 +345,86 @@ void Sema::CheckTopLevelBoundsDecls(VarDecl *D) {
...
+
+    // Function Calls are defined as modifying
+    bool VisitCallExpr(CallExpr *E) {
+      addError(E, MEK_Call);
+      return false;
+    }
+
+    // Comma Expressions are defined as modifying
+    bool VisitBinComma(BinaryOperator *E) {
+      addError(E, MEK_Comma);
```
dtarditi: It is OK to allow comma operators within non-modifying expressions.    The restriction on ',' is a syntactic restriction.  It can't appear at the top-level in a non-modifying expression because non-modifying expressions may appear in comma-separated parameter lists.   Plain C expressions have this problem too, so there is a special syntax for this (an extra production it the grammar, from what I recall).

lib/Sema/SemaBounds.cpp
lenary: Ah, sure, I can correct that.

lib/Sema/SemaBounds.cpp
lenary: I believe before semantic analysis happens, that comma expressions in parameter lists are turned into a vector of arguments within a call expression. This is why calling `dynamic_check(I++, I)` will error with "too many arguments", and not with "comma expression in dynamic check expression". I can pull this out and still only prevent it at the top level.

To be honest, to even test this I had to turn off warnings about unused expressions, so maybe it's not so important, as anything truly non-modifying below this comma operator in the binary tree will be caught anyway.

lib/CodeGen/CGBuiltin.cpp
lenary: This can happen, if we're not in Checked C, and we come across a call to `_Dynamic_check`. However, this `break` falls through to some code that will error for "unrecognized builtin", which is our intended behavior I believe. 

I can add a test that compiling something that includes `_Dynamic_check` but without checked c enabled works as the programmer would expect (calling their self-defined `_Dynamic_check` function)

include/clang/Basic/DiagnosticSemaKinds.td
lenary: Sure, I'll correct this.

lenary:
Ready for re-review.

PR #125, opened 2017-01-27
Adds check to ensure function pointer casts are safe

### LINKED ISSUE #78: Check casts of function names to checked function pointer types.
We added new rules to the Checked C specification that allow a function name to be cast from its unchecked function pointer type to the corresponding checked function pointer type, where the two function types are compatible

The support for casting from an unchecked function pointer type to the corresponding checked function pointer type should already be there, but we do need to add tests for that. 

We need to implement a separate test that ensures that the source is actually a function name and not just a variable with an unchecked function pointer type.

lenary:
We check to make sure that any time we're casting to a `_Ptr` to a function, that the thing we are casting is either:
- a top-level function definition (either checked or unchecked)
- a checked local variable/parameter
- a null pointer

Anything else would be unsound. 

Unfortunately, we have to look quite deeply for the thing we're casting to, because you can ref (`&`) and deref (`*`) a function pointer as many times as you want and implicit casts are inserted to make everything work.

Closes #78


lib/Sema/SemaBounds.cpp
```diff
@@ -578,6 +578,79 @@ namespace {
...
+    // Here we're looking for places where the programmer has cast from an
+    // unchecked function pointer type to a checked function pointer type.
+    // We need to descend through all the casts below the top-level bitcast to find
+    // the eventual thing we're referring to. We then check to make sure that it is
+    // either a top-level function declaration (of either checked or unchecked type), or
+    // a local variable that has a checked function type.
+    //
+    // The only other exception we have is that we can always assign the null pointer to
+    // a checked function pointer type.
+    void FindDisallowedFunctionPtrCasts(ImplicitCastExpr *E) {
```
dtarditi: We want this to apply to explicit user casts as well to checked pointer types, not just implicit casts.   So if a user writes:

ptr<int (void)> f = (ptr<int>(void)) (random_value)

this should be rejected.

lib/Sema/SemaBounds.cpp
```diff
@@ -578,6 +578,79 @@ namespace {
...
+      if (!Ty->isCheckedPointerPtrType() ||
+        !Ty->isFunctionPointerType())
+        return;
+
+      Expr *Operand = E;
+      // This is so we can get down through all & and * operators
+      // it is modelled on the Expr::IgnoreParenCasts-like functions
+      while (true) {
+        // This takes us down to the next "interesting" thing
+        Operand = Operand->IgnoreParenCasts();
```
dtarditi: I think this code needs to be more restrictive.  I think an invalid function pointer value could be produced by a combination of explicit casts in the subexpression of the implicit cast,, LValueToRValue casts, and dereference operators.    OperationKinds.def in clang\include\clang\AST lists the different kinds of casts.

lib/Sema/SemaBounds.cpp
lenary: Ah, of course. Added, and added tests.

lenary:
I have updated both this and Microsoft/checked#102 to reflect our discussions, I hope this is closer. I'll probably need to revisit the error message.

lib/Sema/SemaBounds.cpp
```diff
@@ -578,6 +578,180 @@ namespace {
...
+    //    we are casting between incompatible types. Non-value-preserving casts include
+    //    casts that truncate values, and casts that change alignment. An LValueToRValue
+    //    cast is also non-value-preserving because it reads memory.
+    //
+    // 3. Eventually we may get to the end of the chain of casts. This could end in many
+    //    different kinds of expressions and values, but the ones we're allowing are:
+    //  a) 0 (the null pointer)
+    //  b) DeclRefs to unchecked top-level function definitons.
+    //
+    void FindDisallowedFunctionPtrCasts(CastExpr *E) {
```
dtarditi: I think CheckDisallowedFunctionPtrCasts fits better with the naming that I've observed elsewhere in clang.

lib/Sema/SemaBounds.cpp
```diff
@@ -578,6 +578,180 @@ namespace {
...
+      // The type of the outer value
+      QualType ToType = E->getType();
+
+      // We're only looking for ptr<>s to functions
+      if (!ToType->isCheckedPointerPtrType() ||
+        !ToType->isFunctionPointerType())
+        return;
+
+      Expr *Needle = E->getSubExpr();
```
dtarditi: I think Needle should be initialized to E.   This will pick up on illegal/disallowed casts.

lenary: If we initialize the Needle to E, then the first check about finding something of ptr<> type and checking compatibility will automatically pass, and evaluation will stop.

I have inserted a call to `CheckValuePreservingCast` before the loop begins to make sure we check the top-level cast.

lib/Sema/SemaBounds.cpp
```diff
@@ -578,6 +578,180 @@ namespace {
...
+      QualType NeedleTy = Needle->getType();
```
lenary: It could before, but now we do the final type compatibility check after the loop, it's easier for it not to be.

lib/Sema/SemaBounds.cpp
```diff
@@ -578,6 +578,237 @@ namespace {
...
+
+        // If we've not found a cast or a cast-like operator, 
+        // then we stop descending
+        break;
+      }
+
+      // This feels like a terrible hack, but it's the only way that the type compatibility check
+      // immediately below will pass
+      if (NeedleTy->isFunctionType()) {
+        NeedleTy = S.Context.getPointerType(NeedleTy, CheckedPointerKind::Ptr);
```
lenary: This needs further discussion. It is part of the check that the `ToType` matches the thing we've got to.

lenary:
Ready for another review. I think this is now correct according to our discussions today.

lib/Sema/SemaBounds.cpp
```diff
@@ -578,6 +578,243 @@ namespace {
...
+    // addr-of (&) or deref(*).
+    //
+    // We search this chain, starting at E.
+    //
+    // 1. If we come across something of ptr<> function type, then one of two things
+    //    happens:
+    //  a) this type is compatible with ToType, so then the cast is allowed.
+    //  b) this type not compatible, so we add an error about casting between incompatible
+    //     types and stop descending.
+    //    This allows calling functions with checked types, and local varibles
```
dtarditi: Typo in comment here: varibles

lib/Sema/SemaBounds.cpp
```diff
@@ -578,6 +578,243 @@ namespace {
...
+    //    different kinds of expressions and values, but the ones we're allowing are the
+    //    following, assuming they match the type we're looking for:
+    //  a) 0 (the null pointer)
+    //  b) DeclRefs to unchecked top-level function definitons.
+    //
+    void CheckDisallowedFunctionPtrCasts(CastExpr *E) {
+      // The type of the outer value
+      const QualType ToType = E->getType();
+
+      // We're only looking for casts to checked function pointers
```
dtarditi: Nit: need a period at the end of the comment.

lib/Sema/SemaBounds.cpp
```diff
@@ -578,6 +578,243 @@ namespace {
...
+    // This will add the required error messages
+    bool CheckValuePreservingCastLikeOp(const UnaryOperator *E, const QualType ToType) {
+      QualType ETy = E->getType();
+      QualType SETy = E->getSubExpr()->getType();
+
+      bool between_functions = false;
+
+      switch (E->getOpcode()) {
+      case UO_Deref: {
+        // This may be more conservative than necessary
```
dtarditi: Nit: need a period at the end of the comment.

lib/Sema/SemaBounds.cpp
```diff
@@ -578,6 +578,243 @@ namespace {
...
+    //
+    // Other operations might also be, but this algorithm is currently
+    // conservative.
+    //
+    // This will add the required error messages
+    bool CheckValuePreservingCastLikeOp(const UnaryOperator *E, const QualType ToType) {
+      QualType ETy = E->getType();
+      QualType SETy = E->getSubExpr()->getType();
+
+      bool between_functions = false;
```
dtarditi: You can just have two local variables that are scope by the bodies of the case statement.

lib/Sema/SemaBounds.cpp
```diff
@@ -578,6 +578,243 @@ namespace {
...
+    // Let's term the outer value (after the cast), E, of type ToType.
+    // In the values we're examining, ToType is a ptr<> to a function type.
+    //
+    // To produce E, the programmer is performing a sequence of casts,
+    // both implicit and explicit, and perhaps this sequence includes using
+    // addr-of (&) or deref(*).
+    //
+    // We search this chain, starting at E.
+    //
+    // 1. If we come across something of ptr<> function type, then one of two things
```
dtarditi: It would be good to update the comment to match the order of the code below.  In particular, we handle null pointers first.    That way the code doesn't have to cite an out-of-order case.

PR #135, opened 2017-02-03
Explicit dynamic check statistics

lenary:
In order to prepare for adding a lot more code generation for dynamic checks, I have moved some code around and added statistics. 

Statistics of code generation can be viewed with the `-print-stats` argument to `clang -cc1`. 

I have also moved the warning about failing dynamic checks into semantic analysis instead of code generation, because this seems like a much better place for it. It used to only be added when you asked `clang -cc1` to do codegen (via `-emit-llvm` or similar), but now it's run with all the other semantic analysis, so I had to update some tests which now produced warnings.

### DIFF
--- a/lib/CodeGen/CGDynamicCheck.cpp
+++ b/lib/CodeGen/CGDynamicCheck.cpp
@@ -1,0 +1,53 @@
+//===---- CGDynamicCheck.cpp - Emit LLVM Code for Checked C Dynamic Checks -===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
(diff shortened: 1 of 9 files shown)


lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "CodeGenFunction.h"
+#include "llvm/ADT/Statistic.h"
+
+using namespace clang;
+using namespace CodeGen;
+using namespace llvm;
+
+#define DEBUG_TYPE "DynamicCheckCodeGen"
+
+STATISTIC(DynamicChecksFound, "Number of Dynamic Checks Found");
```
dtarditi: Capitalization of the names of statistics may be inconsistent with capitalization for other statistics.   I don't think other names are mostly in upper case.

PR #137, opened 2017-02-04
Warn on invalid use of Checked C extension flag.

dtarditi:
The Checked C clang compiler issues an error message when the
`-fcheckedc-extension` is used with an invalid language target (such as C++
or Objective C).   This change modifies the compiler to issue a warning message
and ignore the flag instead when that happens.  This allows the flag to be
added to the set of compiler flags in a build without breaking the
build.  There will be a warning when compiling C++ programs.

I hit this issue using the LNT --cflags option to pass the -fcheckedc-extension
flag to the compiiler when running LLVM nightly tests.  The flag was passed to
compilations of C++  programs, which caused them to fail to compile.

This change updates the driver tests that use the -fcheckedc-extension
flag while compiling languages other than C to reflect the new behavior.

Testing:
- The nightly tests in the original branch of
checkedc-llvm-test-suite pass when the -fcheckedc-extension
flag is set (2495 tests).  Of course, none of these tests use the
Checked C extension, so this just verifies that we haven't broken any
of the tests with the compiler modifications for Checked C.
- Ran the driver tests using LIT.


lenary:
:+1: LGTM

Issue #61, opened 2016-09-23
Handle types related to variable arguments better in the Checked C rewriter


awruef:
We want to leave those alone. Right now, we do string compares, which is both non-idiomatic and brittle.

dtarditi:
I looked at the handling of variable argument in clang more closely while fixing a bug.  Here is a summary:
- clang does not seem to know about the `va_list` type itself.   According to the C spec, the `va_list` type must be a complete object types.
- clang creates a special built-in type `__builtin_va_list`.    Clang appears to recognize that the type needs to be around when `va_list` is around.  The mechanism for triggering this was unclear.  The use of two underscores in the beginning of the name means that this is an identifier reserved for system use.
- The way that variable argument information  is represented is platform-specific and considered part of the platform ABI.  Clang typedef's `__builtin_va_list` to the platform-specific type.  The function `ASTContext::getBuiltinVaListDecl` returns the typedef declaration.
- The built-in type is used to type variable argument-related primitive operations.

It seems that the rewriter will need to specially recognize the type named `va_list`.   The rewriter should specially recognize the name typedef name `__builtin_va_list`.   There are a wide variety of platform-specific types used to represent variable argument information, including `char *` and `void *`.   It is not practical to avoid rewriting those types because they could be used in ways completely unrelated to variable arguments.

The rewriter also needs to handle source-level typedefs that map to `va_list`.   We should have a function that recognizes type names that should not be rewritten because they are special names related to variable arguments.

There are two places in the rewriter code that try to handle va_list-typed variables: specialCaseVarIntros in ConstraintBuilder.cpp and PointerVariableConstraint in ProgramInfo.cpp.  The former seems to only be applied to local variable definitions and looks for types named "va_list".  It misses parameters.   The latter checks all pointer types, but it only looks for the x86-x64 ABI-specific type to which __builtin_va_list is typedef'ed.  A bug fix added recognition of `va_list`.  It is unclear why the rewriter needed to recognize the the type to which `__builtin_va_list` is typedef'ed for x64.  This suggests that either clang is using the desugared name when typing built-in primitives or that the rewriter is desugaring a name and losing typedef information.   It is also unclear whether all the recognition can be done in PointerVariableConstraint and specialCaseVarIntros can be drooped.

awruef:
[...]

PR #141, opened 2017-02-08
Fix rewriter test vargs.c that is failing on  x86.

dtarditi:
The vargs.c test now fails on x86.    The converter is rewriting a parameter with type va_list to _Ptr<char> on x86.   I debugged the converter on the example and it appears that the converter is never constraining the parameter to wild.    

There are two places in the code that try to handle `va_list`-typed variables: specialCaseVarIntros in ConstraintBuilder.cpp and PointerVariableConstraint in ProgramInfo.cpp.  The former seems to only be applied to local variable definitions and looks for types named `va_list`.  It was not called for the parameter variables in the falling test. The latter checks at all constraints being added for pointer types, but looks for the x86-x64 ABI-specific type to which `__builtin_va_list` is typedef'ed.  

We can fix the failure by changing the check in PointerVariableConstraint to also check for `va_list`.   We'll make a targeted fix for now.  A broader clean up is tracked by issue #61.

Note that there are formatting issues: the source files for the rewriter have tabs in them, which they aren't supposed to have.  The code that I added doesn't use tabs.  I'll fix the formatting issues in a separate commit..

Testing:
- The Checked C rewriter tests now pass on x86.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -30,6 +30,7 @@ list(APPEND CLANG_TEST_DEPS
   c-index-test diagtool
   clang-tblgen
   clang-offload-bundler
+  checked-c-convert
   )
   
 if(CLANG_ENABLE_STATIC_ANALYZER)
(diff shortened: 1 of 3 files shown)


dtarditi:
I was surprised to see testing fail on my machine locally after I had switched to another branch and then switched back to this branch.  I tracked this down to an out-of-date version of the Checked C rewriter binary.  There was a missing dependency in the CMakeLists.txt file for the clang regression tests on the binary.

lenary:
That dependency was probably exactly what was setting off some of the build failures. I can't run the tests on windows until I get back to my desk, I'll try them on my mac but I'm not sure how brittle they are on it.

PR #154, opened 2017-02-14
Check additional expressions for valid bounds.

dtarditi:
Check that assignments through array_ptrs and reads through array_ptrs use array_ptrs with bounds.
Check that casts to ptr type have sources with bounds.  Also check that member base expressions have bounds if the base expression is a pointer dereference of an array_ptr.

In C, a read through a pointer is done by an lvalue-to-rvalue cast and write through a pointer is done by assignment operation.
 - Add a function that determines if an lvalue is the result of a pointer dereference or array subscript operation involving an array_ptr.
- Use this function to check at an assignment whether a bounds check is needed before writing to the LHS of the assignment.  If it is, check that bounds can be inferred for the LHS lvalue.
- Also use this function at lvalue-to-rvalue casts to check if a bounds check is needed before reading memory.  If it is, check that bounds can be inferred for the lvalue.
- Check at a casts to ptr type that the source expressions has bounds.
- Check that a member base expression has bounds, if the member base expression is a pointer dereference involving an array_ptr.  For member reference expression we plan to bounds check base expressions.  This will ensure that the lvalue produced by the member expression is always in bounds.      - Add code to dump bounds for these new cases.

Clang represents parameters with array types with pointer types.  For parameters with checked array types, implicitly add a bounds expression as part of the retyping, so that bounds checking works out.

Add code that infers bounds for a ptr value by treating it as one element array. Don't infer bounds if the pointee type is a function type,because function types do not have sizes.

Testing:
- Add tests of inference for reads and writes through array_ptr pointers.
- Tests for casts to ptr types and for member base expressions are not written yet.
- Add feature tests to ast-dump-bounds that check that checked array parameters have bounds declarations added, and that the implicit bounds declarations do not override a programmer bounds declaration
- Existing Checked C and clang regression tests pass.  The additional checking found places in Checked C tests where bounds declarations were missing or illegal assignments from unchecked pointers to checked pointers were occurring.  The fixes will be committed separately to the Checked C repo.

### DIFF
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -205,6 +205,17 @@ namespace {
                                              SourceLocation());
     }
 
+    BoundsExpr *CreateSingleElementBounds(Expr *LowerBounds) {
+      assert(LowerBounds->isRValue());
+      // Create an unsigned integer 1
+      IntegerLiteral *One =
+        CreateIntegerLiteral(llvm::APInt(1, 1, /*isSigned=*/false));
+      CountBoundsExpr CBE = CountBoundsExpr(BoundsExpr::Kind::ElementCount,
+                                            One, SourceLocation(),
+                                            SourceLocation());
+      return ExpandToRange(LowerBounds, &CBE);
+    }
+
     Expr *CreateImplicitCast(QualType Target, CastKind CK, Expr *E) {
       return ImplicitCastExpr::Create(Context, Target, CK, E, nullptr,
                                        ExprValueKind::VK_RValue);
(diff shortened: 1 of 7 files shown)


lib/Sema/SemaBounds.cpp
```diff
@@ -374,12 +392,26 @@ namespace {
...
       // TODO: handle side effects within E
       E = E->IgnoreParens();
+      QualType QT = E->getType();
+
+      // If the target value v is a Ptr type, it has bounds(v, v + 1), unless
+      // it is a function pointer type, in which case it has no required
+      // bounds.
+      if (QT->isCheckedPointerPtrType()) {
+         if (QT->isFunctionPointerType())
+           return CreateBoundsNone();
```
lenary: Shouldn't function pointers have a count(1) bounds or equivalent, matching any general `ptr<>`. 

My understanding of the semantics is that it is an error to dereference `bounds(none)`, which is arguably what happens when we call a function. 

This also goes for `bounds(none)` below where we apply the deref to a function pointer.

dtarditi: We only allow ptrs to functions in the semantics.   Array_ptrs to functions are not allowed.    A bound for a pointer to a function makes no sense in C because the size of a function is unknown.  If a f is a function pointer, f + 1 has no meaning.   On the other hand, you can have a bound for a pointer to a pointer to a function.

Issue #155, opened 2017-02-16
unexpected bounds error generated by some relational comparisons


dtarditi:
I am enabling checking that the sources of casts to ptr types have bounds.    This in turn is causing unexpected failures in pointer_types.c.    If we do a comparison of ptr type against an unchecked pointer type, there is an error generated that the expression has no bounds.  For example

ptr<int> x;
int *y;
x > y

I am guessing that a spurious implicit cast to ptr type is being inserted for the unchecked type.    It appears that there is a spurious case of y to ptr<int>.  This in turn may cause a complaint from bounds inference about the implicit cast to ptr type.

This appears in the file typechecking\pointer_types.  For now, I'm commenting out the problematic test cases and adding comments citing this issue number.

dtarditi:
The compiler is picking one of the two pointer types to cast the other pointer type to.    Sometimes it picks the unchecked pointer type and other times it picks the ptr type.   In the first case, there is no error message and in the second case there.   Instead of commenting out the tests, I'm disabling emission of the error message until I fix the underlying problem.

PR #156, opened 2017-02-16
Record inferred bounds information.

dtarditi:
This change:
- Records inferred bounds information.
-  Adds tests for inferred bounds information for base expressions in member expressions and for casts to ptr types.
- Fixes a bug where bounds inference on casts to ptr types was not being done.
- This uncovered a new issue where casts to ptr are being inserted for comparisons of checked pointer and unchecked pointers.  This causes spurious error messages about an expression having unknown bounds.  This is tracked by issue #155.   For now, do not issue a error message if the source of a cast to ptr does not have bounds.

Testing:
- Passed Checked  C regression tests.  The check on casts to ptr uncovered tests where source expressions had unknown bounds.  Those fixes will be checked in separately to the Checked C repo.
- Passed clang regression tests.


lenary:
Test failure documented as #158, going to merge so I can progress.

Issue #145, opened 2017-02-10
Add Dynamic Bounds Checks at Array Subscript and Dereference Expressions


lenary:
Specifically inserting dynamic checks where we do `e[e2]` or `*e` on expressions of `_Array_ptr<T>` type.

The aim is to use as much of the already-written dynamic check code as possible, along with david's inference code, to implement these checks. I expect it to take a week or so.

PR #163, opened 2017-02-22
Dynamic bounds check array subscript

lenary:
This is the branch for #145.

I definitely need more tests, which should actually run some compiled programs. That said, I think this is a good first/second pass at implementing this.

Code generation in Clang is really complicated, with 5 different code generators (4 for each of the evaluation kinds, and then an lvalue code generator) each recursively calling each other through visitors and the like. The aim here has been to have as little overall effect on the code as possible, while still inserting the checks. This is why we save BoundsExprs from where they're inferred for until later in the CodeGenFunction instance, because we need them only after we have Emitted the value that calculates the pointer address. 

I think doing it this way preserves an invariant I believe clang has - that we only call Emit on an expression once, which should translate into making sure we don't evaluate any side-effects more than they would be normally, for instance in checking the expression `e[i++]`


lib/CodeGen/CGExprScalar.cpp
```diff
@@ -1503,6 +1503,27 @@ Value *ScalarExprEmitter::VisitCastExpr(CastExpr *CE) {
...
+      }
+      else if (const auto *UO = dyn_cast<UnaryOperator>(E)) {
+        switch (UO->getOpcode())
+        {
+        case UO_Deref:
+          CGF.SetNextBoundsCheckExpr(InferredBounds, CodeGenFunction::BC_Deref);
+        default:
+          break;
+        }
+      }
```
dtarditi: I think there should be an assert if the code falls off the end here.  It means we are dropping a bounds check unexpectedly.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -1503,6 +1503,27 @@ Value *ScalarExprEmitter::VisitCastExpr(CastExpr *CE) {
...
+
+      if (InferredBounds->isInvalid())
+        break;
+
+      if (isa<ArraySubscriptExpr>(E)) {
+        CGF.SetNextBoundsCheckExpr(InferredBounds, CodeGenFunction::BC_ArraySubscript);
+      }
+      else if (const auto *UO = dyn_cast<UnaryOperator>(E)) {
+        switch (UO->getOpcode())
+        {
```
dtarditi: The indenting appears unusual to me.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -1503,6 +1503,27 @@ Value *ScalarExprEmitter::VisitCastExpr(CastExpr *CE) {
...
+        case UO_Deref:
+          CGF.SetNextBoundsCheckExpr(InferredBounds, CodeGenFunction::BC_Deref);
+        default:
```
dtarditi: There should be an assert if we hit the default.  That means that we are missing a case.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -1575,7 +1596,7 @@ Value *ScalarExprEmitter::VisitCastExpr(CastExpr *CE) {
   }
 
   case CK_IntToOCLSampler:
-    return CGF.CGM.createOpenCLIntToSamplerConversion(E, CGF);
+    return CGF.CGM.createOpenCLIntToSamplerConversion(E, CGF);
```
dtarditi: Unnecessary white space change?

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -1503,6 +1503,27 @@ Value *ScalarExprEmitter::VisitCastExpr(CastExpr *CE) {
   case CK_LValueToRValue:
     assert(CGF.getContext().hasSameUnqualifiedType(E->getType(), DestTy));
     assert(E->isGLValue() && "lvalue-to-rvalue applied to r-value!");
+
+    if (CE->hasInferredBoundsExpr()) {
+      BoundsExpr* InferredBounds = CE->getInferredBoundsExpr();
+
+      if (InferredBounds->isInvalid())
```
dtarditi: The InferredBounds could also be Any.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -1503,6 +1503,27 @@ Value *ScalarExprEmitter::VisitCastExpr(CastExpr *CE) {
...
+        break;
+
+      if (isa<ArraySubscriptExpr>(E)) {
```
dtarditi: The subexpression could be a parenthesized expression.   You'll want to ignore parentheses.

lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -20,41 +20,164 @@ using namespace llvm;
...
+
+void CodeGenFunction::EmitDynamicBoundsCheck(const LValue Addr, const BoundsExpr *Bounds) {
+  ++NumDynamicChecksRange;
+
+  // We can only generate the check if we have the bounds as a range.
+  assert(isa<RangeBoundsExpr>(Bounds) && "Can Only Emit Dynamic Bounds Check For RangeBounds Exprs");
+  const RangeBoundsExpr *BoundsRange = dyn_cast<RangeBoundsExpr>(Bounds);
+
+  bool LowerChkNeeded = true, UpperChkNeeded = true;
+  //// Constant Evaluation
```
dtarditi: I think we should avoid committing large blocks of commented-out code.

lib/CodeGen/CodeGenFunction.h
```diff
@@ -1238,6 +1238,48 @@ class CodeGenFunction : public CodeGenTypeCache {
...
+  }
+
+  void SetNextBoundsCheckExpr(BoundsExpr *Bounds, BoundsCheckKind Kind) {
+    if (!getLangOpts().CheckedC)
+      return;
+
+    assert(!Bounds->isInvalid() && "Only valid bounds checks can be checked against");
+    assert(NextBoundsCheckKind == BC_None
+           && !NextBoundsExprForCheck
+           && "Only one bounds check emmision can be in progress at a time");
```
dtarditi: Typo in spelling of emission.

lib/CodeGen/CGExpr.cpp
```diff
@@ -2995,6 +3016,10 @@ LValue CodeGenFunction::EmitArraySubscriptExpr(const ArraySubscriptExpr *E,
 
   // TODO: Preserve/extend path TBAA metadata?
 
+  if (CheckBounds) {
+    EmitCheckedCSubscriptCheck(LV, CheckBounds);
```
dtarditi: Clang style is to generally avoid single line statements wrapped in braces.  They drop the braces in favor of brevity.

lib/CodeGen/CGExprScalar.cpp
lenary: For the moment I am unwilling to put these asserts in, because I am currently explicitly only handling two kinds of inferred bounds. When I am using all the inferred bounds, or expect to be doing so, sure, for the moment it errors when I need it not to.

lib/CodeGen/CodeGenFunction.h
```diff
@@ -1238,6 +1238,48 @@ class CodeGenFunction : public CodeGenTypeCache {
   const TargetInfo &getTarget() const { return Target; }
   llvm::LLVMContext &getLLVMContext() { return CGM.getLLVMContext(); }
 
+public:
```
dtarditi: There needs to be a comment in the code that describes the overall scheme that you are using here during code generation and why you choose to use it.

lib/CodeGen/CodeGenFunction.h
```diff
@@ -1238,6 +1238,48 @@ class CodeGenFunction : public CodeGenTypeCache {
...
+
+    BoundsExpr* result = NextBoundsExprForCheck;
+    NextBoundsExprForCheck = nullptr;
+    NextBoundsCheckKind = BC_None;
+
+    return result;
+  }
+
+  void SetNextBoundsCheckExpr(BoundsExpr *Bounds, BoundsCheckKind Kind) {
+    if (!getLangOpts().CheckedC)
```
dtarditi: I think you can delete this test that this is only CheckedC.    Given that you have to pass in a BoundsExpr, it is unlikely to be accidentally called.   If you want to, you could this into an assert.

lib/CodeGen/CodeGenFunction.h
```diff
@@ -1238,6 +1238,48 @@ class CodeGenFunction : public CodeGenTypeCache {
...
+      return;
+
+    assert(!Bounds->isInvalid() && "Only valid bounds checks can be checked against");
```
dtarditi: You need to handle the case where Bounds is null here - right now if Bounds is null, you'll try to call a method on it.

lib/CodeGen/CodeGenFunction.h
```diff
@@ -1238,6 +1238,48 @@ class CodeGenFunction : public CodeGenTypeCache {
...
+    BC_ArraySubscript
+  };
+
+private:
+  BoundsExpr* NextBoundsExprForCheck;
+  BoundsCheckKind NextBoundsCheckKind;
+
+public:
+  BoundsExpr* GetAndClearNextBoundsCheckExpr(BoundsCheckKind Kind) {
+    if (!getLangOpts().CheckedC)
```
dtarditi: I think you can remove this check or turn it into an assert.

lib/CodeGen/CodeGenFunction.h
```diff
@@ -1238,6 +1238,48 @@ class CodeGenFunction : public CodeGenTypeCache {
...
+      return nullptr;
+
+    if (NextBoundsCheckKind == BC_None)
+      return nullptr;
+
+    if (NextBoundsCheckKind != Kind)
+      return nullptr;
```
lenary: I was in two minds about this. It effectively depends on what expressions are between the LValueToRValue cast and the deref/array subscript expression, and what order Emit methods are called for them. That said, an assert is probably useful because it shows we are breaking a high-level assumption I have about the code.

test/CheckedC/dynamic-checks/array-subscript-code-gen.c
```diff
@@ -0,0 +1,1021 @@
...
+  // CHECK-AST-NEXT: DeclRefExpr {{.*}} 'gma' 'int checked[3][3]'
+  // CHECK-AST-NEXT: IntegerLiteral {{.*}} 'unsigned long long' 3
+  // CHECK-AST-NEXT: ArraySubscriptExpr
+  // CHECK-AST-NEXT: ImplicitCastExpr {{.*}} <ArrayToPointerDecay>
+  // CHECK-AST-NEXT: ArraySubscriptExpr
+
+  // The following line of LLVM IR is offensively long
+  // CHECK-IR: br i1 and
+  // CHECK-IR-SAME: (i1 icmp sle (i32 ptrtoint ([3 x [3 x i32]]* @gma to i32),
+  // CHECK-IR-SAME: i32 ptrtoint (i32* getelementptr inbounds ([3 x [3 x i32]], [3 x [3 x i32]]* @gma, i32 0, i64 1, i32 2) to i32)),
```
lenary: I believe the indexing in `GEP @gma 0 1 2` is correct, because 1*3 + 2 == 5.

test/CheckedC/dynamic-checks/array-subscript-code-gen.c
```diff
@@ -0,0 +1,1021 @@
...
+  // CHECK-AST-NEXT: DeclRefExpr {{.*}} 'gma' 'int checked[3][3]'
+  // CHECK-AST-NEXT: IntegerLiteral {{.*}} 'unsigned long long' 3
+  // CHECK-AST-NEXT: ArraySubscriptExpr
+  // CHECK-AST-NEXT: ImplicitCastExpr {{.*}} <ArrayToPointerDecay>
+  // CHECK-AST-NEXT: ArraySubscriptExpr
+
+  // The following line of LLVM IR is offensively long
+  // CHECK-IR: br i1 and
+  // CHECK-IR-SAME: (i1 icmp sle (i32 ptrtoint ([3 x [3 x i32]]* @gma to i32),
+  // CHECK-IR-SAME: i32 ptrtoint (i32* getelementptr inbounds ([3 x [3 x i32]], [3 x [3 x i32]]* @gma, i32 0, i32 2, i32 -2) to i32)),
```
lenary: This indexing is more clearly "correct" as a translation. That said, I think a test that actually runs this code dynamically would be more useful.

lib/CodeGen/CodeGenFunction.h
```diff
@@ -1238,6 +1238,53 @@ class CodeGenFunction : public CodeGenTypeCache {
...
+    NextBoundsExprForCheck = nullptr;
+    NextBoundsCheckKind = BC_None;
+
+    return result;
+  }
+
+  // This sets the next bounds expression to perform a check against, ready to be retrieved by
+  // GetAndClearNextBoundsCheckExpr.
+  void SetNextBoundsCheckExpr(BoundsExpr *Bounds, BoundsCheckKind Kind) {
+    assert(getLangOpts().CheckedC && "SetNextBoundsCheckExpr requires Checked C to be enabled");
```
dtarditi: I think you can delete the message part of this assert - it is pretty clear what happened.

lib/CodeGen/CodeGenFunction.h
```diff
@@ -1238,6 +1238,53 @@ class CodeGenFunction : public CodeGenTypeCache {
...
+
+  // The kind of expression that NextBoundsExprForCheck is being stored for.
+  // If this is BC_None, NextBoundsExprForCheck should definitely be nullptr.
+  BoundsCheckKind NextBoundsCheckKind;
+
+public:
+  // This retrieves the saved bounds for the next check, zeroing where they were stored
+  // This will give an assertion error if Kind does not match the previously saved kind.
+  BoundsExpr* GetAndClearNextBoundsCheckExpr(BoundsCheckKind Kind) {
+    assert(getLangOpts().CheckedC && "GetAndClearNextBoundsCheckExpr requires Checked C to be enabled");
```
dtarditi: I think you can delete the message part of this assert - it is pretty clear what happened.

Issue #142, opened 2017-02-09
Check additional expressions for valid bounds


dtarditi:
We should check that:
- reads through array_ptr pointers have bounds.
- writes through array_ptr pointers have bounds.
- casts to ptr types have bounds.

This involves 2 pieces of work:
1. Extending the AST walker for checking the presence of bounds to examine the additional expressions.
2. Adding a method to determine whether an lvalue is derived from an array_ptr rvalue (and thus needs bounds check when it is read or written).

PR #166, opened 2017-02-23
Check increment/decrement operators and clean up inferred bounds.

dtarditi:
This change checks that increment/decrement operators are applied to subexpressions with bounds when the subexpressions dereference array_ptr values.   This completes the work for issue #142.

This change also cleans up a bunch of aspects of inferred bounds:
- It updates dumps of inferred bounds to remove redundant information.    The IR includes inferred bounds, so under the -fdump-inferrred-bounds flag, we no longer need to dump the inferred bounds out separately.  This requires updating the inferred-bounds test cases with the new output format.
- It deletes the inferred bounds properties for array subscript expressions, which are not necessary. 
- It updates comments on the remaining inferred bounds properties to describe what the inferred bounds apply to.
- It modifies IR dumps to describe what the inferred bounds apply to as well.

Testing:
- The inferred-bounds tests had to be updated to reflect the new output.
- Passes existing clang and LLVM regression tests.
- Passes Checked C regression tests.


include/clang/AST/Expr.h
```diff
@@ -2662,7 +2655,7 @@ class CompoundLiteralExpr : public Expr {
     return LParenLoc;
   }
   SourceLocation getLocEnd() const LLVM_READONLY {
-    // FIXME: Init should never be null.
+    // FIXME: Init shold never be null.
```
lenary: Typo?

PR #167, opened 2017-02-24
Compute rvalue bounds for more expressions.

dtarditi:
This adds computation of rvalue bounds for the following pointer-typed expressions:
- Pointer addition and subtraction.
- Post/pre-increment and decrement.
- Assignments (assignments produce rvalues in C).
- Compound assiginments involving addition and subtraction.
This completes Github issue #121.

In addition, it adds computation of rvalue bounds for the following integer-typed expressions (section 5.4.2 in the the Checked C version 0.6 specification):
- Addition, subtracation, and multiplication.
- Post/pre-increment and decrement.
- Bitwise logical operations
- Unary +, -, and bitwise complement.

During testing, I found that the code for checking that compound assignment operators (+= and so on) have bounds was not correct and triggered an assert..  The code was just checking the bounds of the rhs expression.  We actually  need to check the rvalue bounds of the entire expression for compound assignment operators. For pointer arithmetic, this is a no-op (there's no possible way it could fail to have bounds), but for integer arithmetic, it is not a no-op.

Testing:
- Added new tests for inference of pointer arithmetic bounds.
- Still need to write tests for bounds-inference for integer-typed expressions.
- Passed Checked C regression tests.
- Passed clang regression tests.

### DIFF
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -453,8 +453,7 @@ namespace {
       }
     }
 
-    // Compute the bounds of a cast operation that produces
-    // an rvalue.
+    // Compute the bounds of a cast operation that produces an rvalue.
     BoundsExpr *RValueCastBounds(CastKind CK, Expr *E) {
       switch (CK) {
         case CastKind::CK_BitCast:
@@ -476,8 +475,7 @@ namespace {
       }
     }
 
-    // Compute the bounds of an expression that produces
-    // an rvalue.
+    // Compute the bounds of an expression that produces an rvalue.
     BoundsExpr *RValueBounds(Expr *E) {
       assert(E->isRValue());
       E = E->IgnoreParens();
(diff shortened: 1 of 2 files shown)


lib/Sema/SemaBounds.cpp
```diff
@@ -520,22 +518,77 @@ namespace {
...
-              return LValueBounds(SubExpr);
-            }
-            default:
-              // TODO: fill in other cases
+          if (Op == UnaryOperatorKind::UO_AddrOf) {
+            Expr *SubExpr = UO->getSubExpr();
+            if (SubExpr->getType()->isFunctionType())
               return CreateBoundsNone();
+
+              return LValueBounds(SubExpr);
```
lenary: Your indentation here is odd, it looks like you have a return directly after a return, but i'm pretty sure the logic is sound.

lib/Sema/SemaBounds.cpp
```diff
@@ -520,22 +518,77 @@ namespace {
...
+            return RValueBounds(UO->getSubExpr());
+          return CreateBoundsNone();
         }
-        // TODO: fill in these cases
         case Expr::BinaryOperatorClass:
-        case Expr::CompoundAssignOperatorClass:
+        case Expr::CompoundAssignOperatorClass: {
+          BinaryOperator *BO = dyn_cast<BinaryOperator>(E);
+          if (!BO) {
+            assert("unexpected cast failure");
```
lenary: `assert("<string>")` will never trigger, as strings are truthy. You probably want `assert(false && "<string>")`. I should have caught this several pull-requests ago.

lib/Sema/SemaBounds.cpp
```diff
@@ -520,22 +518,77 @@ namespace {
...
+
+          // Pointer arithmetic.
+          if (LHS->getType()->isPointerType() &&
+              RHS->getType()->isIntegerType() &&
+              BinaryOperator::isAdditiveOp(Op)) {
+            return IsCompoundAssignment ?
+              LValueTargetBounds(LHS) : RValueBounds(LHS);
+          }
+          if (RHS->getType()->isPointerType() &&
+              LHS->getType()->isIntegerType() &&
```
lenary: these lines confused me for a while, because I didn't notice LHS and RHS had been swapped. Maybe more commentary would be useful here.

lib/Sema/SemaBounds.cpp
```diff
@@ -711,14 +764,17 @@ namespace {
       // Bounds of the right-hand side of the assignment
       BoundsExpr *RHSBounds = nullptr;
 
-      // Check that the RHS of the assignment has bounds if the
+      // Check that the value being assigned has bounds if the
       // target of the LHS lvalue has bounds.
       if (LHSType->isCheckedPointerType() ||
-          LHSType->isIntegralOrEnumerationType()) {
+          LHSType->isIntegerType()) {
```
lenary: What are we doing about EnumerationTypes now? Do we mind that we don't cover them any more?

lib/Sema/SemaBounds.cpp
dtarditi: I'll fix this

lib/Sema/SemaBounds.cpp
dtarditi: I'll convert all of these to llvm_unreachable statements.

lib/Sema/SemaBounds.cpp
dtarditi: If it confused you, it'll confuse other people.  I'll change the code so that the operands are tested in the same order as the prior test.

lib/Sema/SemaBounds.cpp
dtarditi: I read the clang documentation and it says that isIntegralOrEnumerationType() is meant to be used for C++ code.  For C code, you can still with isIntegerType(), which treats enumerations as integers too.

dtarditi:
Thanks for the feedback - I believe that I've addressed it with my commit.

Issue #37, opened 2016-08-25
Add support for relative bounds expressions


dtarditi:
The Checked C specification has relative bounds expressions.  We need to extend the IR to represent relative bounds expressions, add parsing support for that, and add typechecking for that.

jijoongmoon:
I'm trying to implement relative bounds expressions. In order to do that, I have several questions about it.

. Most of the example codes in the specification, I can see the relative expression with types such as rel_align(char). But I'm supposed that type is not the only one for the variable of relative expression. 

. This relative expression could be after the bounds and span expressions. count and byte_count is suppose to have the alignment as type size and one byte for each. But we need to have alignment for the range bounds expression with relative expression. So, I tried to use private TypeSourceInfo * in RangeBoundsExpr class. Do you think which on is ok better to create another class for relative expression or use private variable?

. Handling the exceptions like this  bounds(none + arr, arr + len) rel_align(char) = arr; , When I'm reading the clang code, current checkedc implementation suppose to this as NullaryBoundsExpr and SkipUntil the r_paran of bounds expression. if I implement the rel_align after this bounds expression, the parser also should go to the r_paran of rel_aling expression. Am I right?

. Talking about the ASTDumper for relative alignment, I think it is ok something like " rel_align : char " after the dump Bounds Kind. How about it?

dtarditi:
- Yes, in general, a type name could appear in the `rel_align` clause.  For example, a C programmer might create a `typedef` and usedthe `typedef` name in the `rel_align` clause.
- I would recommend that you create a class to hold the relative alignment information, and then have the `RangeBoundsExpr` point to an instance of this class.    The reason is that relative alignment will be rare, and having the information in a separate class that is created only on "demand" will save memory overall. We also need to worry about `rel_align_value` clauses, which would hold an integer.
- I would implement the parsing as:
-- Try parsing the bounds expression.   If something goes wrong, skip to the rparen the rparen.
-- Then try parsing the rel_align clause.  If something goes wrong, skip to the r_paren of the relative align expression.
- Yes, your suggestion for the ASTDumper sounds good.

jijoongmoon:
I tested with check-clang and check-checkedc.  In case of check-checkedc, there is no error message except the tests in samples directory. 
And there are two errors in check-clang case. 
one is unexpected error in Modules/explicit-build-extra-files.cpp and I confirmed the python configuration is cause of this error. I tested the check-clang with another computer, there is no unexpected error.
[...]

PR #169, opened 2017-02-26
Adds parsing of relative bounds expression (#037)

jijoongmoon:
- Adds RelativeBoundsExpr for relative expression.
- Relative expression is parsed after bounds expression parsing is done and the tokencache for the function parameter is extended to the relative expression
- Adds variable and member functions in RangeBoundsExpr to handle relative bounds expression.
- Currently builtin type variable can be parsed.
- Adds write the type of relative bounds expression in RangeBoundsExpr AST like "rel_align : char"

### DIFF
--- a/include/clang/AST/Expr.h
+++ b/include/clang/AST/Expr.h
@@ -4622,9 +4622,11 @@ public:
     // To save space and for programmng convenience, we store the 
     // ": ptr" interop annotation as a bounds expression.
     InteropTypeAnnotation = 6,
+    //rel_aling(char)
+    Relative = 7,
 
     // Sentinel marker for maximum bounds kind.
-    MaxBoundsKind = InteropTypeAnnotation
+    MaxBoundsKind = Relative
   };
 
   static_assert(MaxBoundsKind < (1 << NumBoundsExprKindBits), "kind field too small");
@@ -4683,6 +4685,8 @@ public:
   bool isRange() const {
     return getKind() == Range;
   }
+
+  bool isRelative() const { return getKind() == Relative; }
 
   bool isInteropTypeAnnotation() const {
     return getKind() == InteropTypeAnnotation;
@@ -4748,11 +4752,40 @@ public:
   }
 };
 
+class RelativeBoundsExpr : public BoundsExpr {
+private:
+  TypeSourceInfo *AlignType;
+
+public:
+  RelativeBoundsExpr(TypeSourceInfo *Align, SourceLocation StartLoc,
+                     SourceLocation RParenLoc)
+      : BoundsExpr(RelativeBoundsExprClass, Relative, StartLoc, RParenLoc),
+        AlignType(Align) {}
+
+  explicit RelativeBoundsExpr(EmptyShell Empty)
+      : BoundsExpr(RelativeBoundsExprClass, Empty), AlignType(nullptr) {}
+
+  TypeSourceInfo *getAlignTypeInfoAsWritten() const { return AlignType; }
+  void setAlignTypeInfoAsWritten(TypeSourceInfo *Ty) { AlignType = Ty; }
+  QualType getTypeAsWritten() const { return AlignType->getType(); }
+  bool hasAlignType() const { return AlignType != nullptr; }
+
+  static bool classof(const Stmt *T) {
+    return T->getStmtClass() == RelativeBoundsExprClass;
+  }
+
+  // Iterators
+  child_range children() {
+    return child_range(child_iterator(), child_iterator());
+  }
+};
+
 /// \brief Represents a Checked C range bounds expression.
 class RangeBoundsExpr : public BoundsExpr {
 private:
   enum { LOWER, UPPER, END_EXPR };
   Stmt *SubExprs[END_EXPR];
+  Stmt *Relative;
 
 public:
   RangeBoundsExpr(Expr *Lower, Expr *Upper, SourceLocation StartLoc,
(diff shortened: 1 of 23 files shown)


include/clang/Serialization/ASTBitCodes.h
```diff
@@ -1322,7 +1322,6 @@ namespace clang {
       EXPR_OBJC_ARRAY_LITERAL,
       EXPR_OBJC_DICTIONARY_LITERAL,
 
-    
       /// \brief An ObjCEncodeExpr record.
```
dtarditi: It looks like there are only white space changes to this file.  Please revert the white space changes.   It will make updating to newer versions of the clang sources more work because we'll have to examine these diffs.

lib/Sema/SemaExpr.cpp
```diff
@@ -12359,6 +12359,26 @@ ExprResult Sema::ActOnRangeBoundsExpr(SourceLocation BoundsKWLoc,
...
+  ExprResult Result =
+      ActOnRangeBoundsExpr(BoundsKWLoc, LowerBound, UpperBound, RParenLoc);
+  Range = cast<RangeBoundsExpr>(Result.get());
+  Range->setRelativeBoundsExpr(Relative);
+  return Result;
+}
+
+ExprResult Sema::ActOnRelativeBoundsExpr(SourceLocation BoundsKWLoc,
+                                         ParsedType Ty,
+                                         SourceLocation RParneLoc) {
```
dtarditi: There is a typo: RParneLoc should be RParenLoc.

include/clang/AST/Expr.h
```diff
@@ -4748,27 +4752,68 @@ class CountBoundsExpr : public BoundsExpr {
   }
 };
 
+class RelativeBoundsExpr : public BoundsExpr {
```
dtarditi: Could you rename to this class to RelativeBoundsClause?

I don't think  it needs to be subclass of BoundsExpr or an expression at all.  It's a helper class and doesn't make much sense as an expression.

It would be helpful to add a comment describing why we have it.  There will be two different ways of specifying relative alignment, so we have a class to encapsulate it.  In addition, we expect bounds declarations with relative alignment clauses to be rare, so having the data out-of-line reduces memory usage.

include/clang/Parse/Parser.h
```diff
@@ -1664,7 +1667,12 @@ class Parser : public CodeCompletionHandler {
   /// type annotation.
   bool StartsInteropTypeAnnotation(Token &tok);
 
+  bool StartsRelativeBoundsExpression(Token &tok);
```
dtarditi: A better name might be "StartRelativeBoundsClause".

include/clang/AST/Expr.h
```diff
@@ -4748,27 +4752,68 @@ class CountBoundsExpr : public BoundsExpr {
...
 
   explicit RangeBoundsExpr(EmptyShell Empty)
-    : BoundsExpr(RangeBoundsExprClass, Empty) {}
+    : BoundsExpr(RangeBoundsExprClass, Empty), Relative(nullptr) {}
 
   Expr *getLowerExpr() const { return cast<Expr>(SubExprs[LOWER]); }
   void setLowerExpr(Expr *E) { SubExprs[LOWER] = E; }
   Expr *getUpperExpr() const { return cast<Expr>(SubExprs[UPPER]); }
   void setUpperExpr(Expr *E) { SubExprs[UPPER] = E; }
+  Expr *getRelative() const { return cast<Expr>(Relative); }
```
dtarditi: I'd suggest using the names getRelativeAlignment and hasRelativeAlignment.

include/clang/AST/Expr.h
```diff
@@ -4748,27 +4752,68 @@ class CountBoundsExpr : public BoundsExpr {
...
+      : BoundsExpr(RelativeBoundsExprClass, Relative, StartLoc, RParenLoc),
+        AlignType(Align) {}
+
+  explicit RelativeBoundsExpr(EmptyShell Empty)
+      : BoundsExpr(RelativeBoundsExprClass, Empty), AlignType(nullptr) {}
+
+  TypeSourceInfo *getAlignTypeInfoAsWritten() const { return AlignType; }
+  void setAlignTypeInfoAsWritten(TypeSourceInfo *Ty) { AlignType = Ty; }
+  QualType getTypeAsWritten() const { return AlignType->getType(); }
+  bool hasAlignType() const { return AlignType != nullptr; }
```
dtarditi: I think it is expensive to use the TypeSourceInfo to compute the type.   All expressions store their QualType, so you can just call getType() to get the type of expression.  This includes expressions that have TypeSourceInfo.

Could you add a new field that stores that QualType, as well as getter and setter properties for that?

lib/Parse/ParseExpr.cpp
```diff
@@ -2949,6 +3002,10 @@ ExprResult Parser::DeferredParseBoundsExpression(std::unique_ptr<CachedTokens> T
   ConsumeAnyToken();   // Skip past the current token to the new tokens.
   ExprResult Result = ParseBoundsExpression();
 
+  if (StartsRelativeBoundsExpression(Tok)) {
```
dtarditi: This is not the right place to be parsing the relative alignment clause.  It needs to be parsed as part of ParseBoundsExpression, and only if comes after a 'bounds' expression.

I think a better place to parse the relative BoundsExpression is at line 2912 in ParseExpr.cpp (in the function ParseBoundsExpression).

lib/Parse/ParseExpr.cpp
```diff
@@ -2908,6 +2921,37 @@ ExprResult Parser::ParseBoundsExpression() {
   return Result;
 }
 
+ExprResult Parser::ParseRelativeBoundsExpression(Expr *Expr) {
+  ExprResult Result;
+  tok::TokenKind TokKind = Tok.getKind();
+  if (TokKind != tok::identifier) {
```
dtarditi: I think you should delete this if-statement.    It would be incorrect say that we are expecting a bounds expression here.  We are expecting a relative bounds expression, which is covered by the next test.

include/clang/Serialization/ASTBitCodes.h
jijoongmoon: Thanks, I will revert the white space. And I suppose that there is more place like this. I will revert them too.!

lib/Sema/SemaExpr.cpp
jijoongmoon: I will change this.

include/clang/AST/Expr.h
jijoongmoon: I see. I will try to make relative class as a helper class and write a comment about this class.

include/clang/AST/Expr.h
jijoongmoon: I see. I will change to use QualType and add getter and setter.

lib/Parse/ParseExpr.cpp
jijoongmoon: I see. I will put this in the function Parse BoundsExpression.

include/clang/Parse/Parser.h
jijoongmoon: I will change this. Thanks.

include/clang/AST/Expr.h
jijoongmoon: I see. I will change it.

lib/Parse/ParseExpr.cpp
jijoongmoon: I see. I will delete it. Thanks.

PR #171, opened 2017-02-28
Move bounds expressions to lvalue-producing expressions with bounds checks.

dtarditi:
Modify the generated AST so that bounds expressions are on lvalue-producing expressions that need bounds checks.   This simplifies code generation and makes the IR easier to understand. 

**Note: you will have to a clean build of the compiler after this change is committed or if you intend to test this locally.**   This change modifies the IR layout for BinaryOperators by deleting a field.  For some reason, this causes mysterious memory corruptions, even though we are only deleting a field..  My guess is that there is some dependency that is missing in the build system.   The problem goes away with a clean build.

Currently, we are storing the bounds information on parent expressions of expressions that needs bounds checks.   It is easier to store the information on the expression that produces the lvalue that needs a bounds check.    The bounds checks end up on pointer dereference and array subscript expressions, where you'd expect them to be.  They also end up on member expressions that use the -> operator.  

We determine whether an lvalue expression that produces an lvalue from an array_ptr needs a bounds check based on its enclosing context.  Lvalue expressions that are used to read or write memory or as the base operand of a member reference need bounds checks..   The contexts where lvalue expressions need bounds checks include:
- The LHS of an assignment expression (writes memory) or compound assignment expression (reads/writes memory).
- As an operand of an increment/decrement operator (reads/writes memory).
- As the base operand of member reference expression using the  `.` operator (ensure the resulting lvalue of the member reference expression is in range).
 
This change deletes bounds expressions from binary expressions and adds them back to array subscript expressions.   It also deletes the code in code generation for propagating bounds information from a parent expression to a child expression.

Testing:
- Updated clang Checked C tests to reflect the new location of bounds information 
- Passed clang Checked C tests.
- Passed Checked C regression tests.
- Clang test suite run in progress now.


lib/CodeGen/CGExprScalar.cpp
```diff
@@ -1597,7 +1575,7 @@ Value *ScalarExprEmitter::VisitCastExpr(CastExpr *CE) {
   }
 
   case CK_IntToOCLSampler:
-    return CGF.CGM.createOpenCLIntToSamplerConversion(E, CGF);
+    return CGF.CGM.createOpenCLIntToSamplerConversion(E, CGF);
```
lenary: So, I worked out what this change is. It's not whitespace exactly, it's the wrong kind of line ending (this changes the ending from windows style to unix style). Let's leave it in this fix so it doesn't appear every time we change this file.

PR #173, opened 2017-03-02
Adds Parsing relative alignment (#037)

jijoongmoon:
. Adds RelativeBoundsClause class as a subclass of RangeBoundsExpr.
  . Adds RelativeBoundsClause class, RelativeTypeBoundsClause and RelativeConstExprBoundsClause.
  . Adds RelativeBoundsClause and getter/setter in RangeBoundsExpr class to point relative alignment.
  . Adds Parsing and Creating RelativeBoundsClause obj in ParseBoundsExpression(). Depending on Type
    of relative alingment, it parse QualType or Const Expression.
    Set the RelativeBoundsClause After ParseBoundsExpression if the Bounds Expressiong is Range Bounds Expression.
  . For the Function Prototype, adds the relative expression into the Tokencache if there is relative alignment keywords.
  . Adds Sementic to create RelativeBoundsClause object and Modify RebuildRangeBoundsExpr to take RelativeBoudsClause.
  . Modify AST Printer for the RangeBoundsExpr to print relative clause. Depending on the it's type,
    it prints rel_align or rel_align_value.


lib/AST/StmtPrinter.cpp
```diff
@@ -1873,6 +1873,17 @@ void StmtPrinter::VisitRangeBoundsExpr(RangeBoundsExpr *Node) {
   OS << ", ";
   PrintExpr(Node->getUpperExpr());
   OS << ")";
+  if (Node->hasRelativeBoundsClause()) {
+    RelativeBoundsClause *Expr = cast<RelativeBoundsClause>(Node->getRelativeBoundsClause());
+    if(Expr->getClauseKind() == RelativeBoundsClause::Kind::Type){
```
dtarditi: Please add a space after the if.

lib/AST/StmtPrinter.cpp
```diff
@@ -1873,6 +1873,17 @@ void StmtPrinter::VisitRangeBoundsExpr(RangeBoundsExpr *Node) {
...
+      OS << " rel_align(";
+      (cast<RelativeTypeBoundsClause>(Expr)->getType()).print(OS, Policy);
+    }else{
```
dtarditi: Please add a space after the `}`.

lib/Sema/SemaExpr.cpp
```diff
@@ -12359,6 +12359,50 @@ ExprResult Sema::ActOnRangeBoundsExpr(SourceLocation BoundsKWLoc,
...
 }
 
+ExprResult Sema::CreateRangeBoundsExpr(SourceLocation BoundsKWLoc,
+                                       Expr *LowerBound, Expr *UpperBound,
+                                       RelativeBoundsClause *Relative,
+                                       SourceLocation RParenLoc) {
+  RangeBoundsExpr *Range = nullptr;
+  ExprResult Result =
+      ActOnRangeBoundsExpr(BoundsKWLoc, LowerBound, UpperBound, RParenLoc);
+  Range = cast<RangeBoundsExpr>(Result.get());
```
dtarditi: ExprResult can be invalid or carry a value with it that is valid.  You need to check if Result.isInvalid() and return ExprError() if it is.   You can't call Result.get on an invalid result.

lib/Sema/SemaExpr.cpp
```diff
@@ -12359,6 +12359,50 @@ ExprResult Sema::ActOnRangeBoundsExpr(SourceLocation BoundsKWLoc,
                                        RParenLoc);
 }
 
+ExprResult Sema::CreateRangeBoundsExpr(SourceLocation BoundsKWLoc,
```
dtarditi: Please rename this to ActOnRangeBoundsExpr.   Create* functions are used to allocate data and not do any semantic checking.

lib/Parse/ParseExpr.cpp
```diff
@@ -2905,9 +2913,49 @@ ExprResult Parser::ParseBoundsExpression() {
     SkipUntil(tok::r_paren, StopAtSemi | StopBeforeMatch);
 
   PT.consumeClose();
+
+  if (StartsRelativeBoundsClause(Tok)) {
```
dtarditi: I think this needs to be moved to the body of ParseBoundsExpression().     We should parse a relative bounds clause if it follows a range bounds expression.

lib/Parse/ParseExpr.cpp
```diff
@@ -2806,8 +2811,11 @@ ExprResult Parser::ParseInteropTypeAnnotation(const Declarator &D, bool IsReturn
 
 ExprResult Parser::ParseBoundsExpressionOrInteropType(const Declarator &D,
                                                       bool IsReturn) {
-  if (StartsBoundsExpression(Tok))
-    return ParseBoundsExpression();
+
+  if (StartsBoundsExpression(Tok)) {
+    ExprResult Results = ParseBoundsExpression();
```
dtarditi: If you follow my suggestion below, I don't think you need to make this change.

lib/Parse/ParseExpr.cpp
```diff
@@ -2905,9 +2913,49 @@ ExprResult Parser::ParseBoundsExpression() {
...
+  }
+
+  IdentifierInfo *Ident = Tok.getIdentifierInfo();
+  SourceLocation BoundsKWLoc = Tok.getLocation();
+  ConsumeToken();
+  BalancedDelimiterTracker PT(*this, tok::l_paren);
+  if (PT.expectAndConsume(diag::err_expected_lparen_after,
+                          Ident->getNameStart()))
+    return false;
+  if( Ident == Ident_rel_align){
```
dtarditi: Please add a space after the `if` and remove the space after the `(`.  Also please add a space after the `)'

include/clang/AST/Expr.h
```diff
@@ -4741,16 +4741,86 @@ class CountBoundsExpr : public BoundsExpr {
   }
 };
 
+class RelativeBoundsClause {
```
dtarditi: This class looks good.  Thank you for making this change.

lib/Parse/ParseExpr.cpp
```diff
@@ -2905,9 +2913,49 @@ ExprResult Parser::ParseBoundsExpression() {
...
+  if (StartsRelativeBoundsClause(Tok)) {
+    ParseRelativeBoundsClause(Result.get());
+  }
+
   return Result;
 }
 
+bool Parser::ParseRelativeBoundsClause(Expr *Expr) {
+  RelativeBoundsClause *Result;
+  ExprResult Results;
```
dtarditi: I think you need to replace this with a Boolean variable that indicates an error, instead of using ExprResult.

lib/Parse/ParseExpr.cpp
```diff
@@ -2905,9 +2913,49 @@ ExprResult Parser::ParseBoundsExpression() {
...
+
+  IdentifierInfo *Ident = Tok.getIdentifierInfo();
+  SourceLocation BoundsKWLoc = Tok.getLocation();
+  ConsumeToken();
+  BalancedDelimiterTracker PT(*this, tok::l_paren);
+  if (PT.expectAndConsume(diag::err_expected_lparen_after,
+                          Ident->getNameStart()))
+    return false;
+  if( Ident == Ident_rel_align){
+    TypeResult ResultTy = ParseTypeName();
```
dtarditi: ResultTy could be invalid - you need to check if it is before calling ResultTy.get().

lib/Parse/ParseExpr.cpp
```diff
@@ -2905,9 +2913,49 @@ ExprResult Parser::ParseBoundsExpression() {
...
+    Result = Actions.ActOnRelativeTypeBoundsClause(BoundsKWLoc, ResultTy.get(),
+					     Tok.getLocation());
+  } else {
+    Results = Actions.CorrectDelayedTyposInExpr(ParseAssignmentExpression());
```
dtarditi: You'll need to create an ExprResult variable to hold the result of Actions.CorrectDelayedTyposInExpr().

You'll need to check if this valid is before calling get().

lib/AST/StmtPrinter.cpp
jijoongmoon: I will check clang-format again.

lib/AST/StmtPrinter.cpp
jijoongmoon: I will check clang-format.

lib/Sema/SemaExpr.cpp
jijoongmoon: The reason I wrote is that it is used in RebuildRangeBoundsExpr() in TreeTransform.h. and ActOnRangeBoundsExpr is called in ParseBoundsExpression. When I looked into the source code, I thought the CreateBoudsInteropType is kind of CerateRangeBoundsExpr. In the CreateRangeBoundsExpr, ActOnRangeBoundsExpr is called and the RelativeBoundsClause is setted.

lib/Sema/SemaExpr.cpp
jijoongmoon: I see. I will add Error check.

lib/Parse/ParseExpr.cpp
jijoongmoon: Yes. It does not need. I will revert it.

lib/Parse/ParseExpr.cpp
jijoongmoon: I think this is not changed code during branch merge. I will delete it.

jijoongmoon: The reason I put this at the end of ParseBoundsExpression() is that PT.consumeClose() at the previous line. The PT is affected all the Bounds Expression and I think if it moves in the parsing of RangeBoundsExpr, it might not be natural (something like ... bounds(a, a+1 rel_align(char)). and have to add PT.consumeClose() for every BoundsExpression. So I adds check the Ident of Tok is kind of relative alignment caluase in StartRelativeBoundsClause() to do the parsing of relativeboundsclause.

lib/Parse/ParseExpr.cpp
jijoongmoon: I will change it with a Boolean variable and handle the error after ParseRelativeBoundsClause.

lib/Parse/ParseExpr.cpp
jijoongmoon: I will change it.

lib/Parse/ParseExpr.cpp
jijoongmoon: I will add check invalidation of ResultTy.

lib/Parse/ParseExpr.cpp
jijoongmoon: I will add the check.

include/clang/AST/Expr.h
jijoongmoon: Thanks, I will change the code following your comment and do PR again.

lib/Parse/ParseExpr.cpp
dtarditi: Ah, yes, this is the correct place.   The Github pull request omits lines of code and I got confused about where this had been placed.

I think this if-statement should only be executed only if the code started parsing a range bounds expression before.  You may need to add a flag to record this.

lib/Parse/ParseExpr.cpp
dtarditi: The Result.get() could fail if the Result is invalid.   You may need to check if the Result is invalid and pass a null pointer into ParseRelativeBoundsClause.   Then in ParseRelativeBoundsClause you can check that the pointer is non-null before trying to attach the bounds expression.

lib/Parse/ParseExpr.cpp
```diff
@@ -2905,9 +2913,49 @@ ExprResult Parser::ParseBoundsExpression() {
...
+
   return Result;
 }
 
+bool Parser::ParseRelativeBoundsClause(Expr *Expr) {
+  RelativeBoundsClause *Result;
+  ExprResult Results;
+  tok::TokenKind TokKind = Tok.getKind();
+  if (TokKind != tok::identifier) {
+    Diag(Tok, diag::err_expected_bounds_expr);
```
dtarditi: I don't think this case can ever occur because the caller checks that the token is an identifier that starts a relative bounds clause.  I think you should replace the Diag message with an assert.

lib/Sema/SemaExpr.cpp
dtarditi: OK, that makes sense.   I think it is fine to leave it "as is".

lib/Parse/ParseExpr.cpp
jijoongmoon: Yes. I will delete it.

lib/Parse/ParseExpr.cpp
```diff
@@ -2905,9 +2910,67 @@ ExprResult Parser::ParseBoundsExpression() {
...
+  RelativeBoundsClause *RelativeClause = nullptr;
+  ExprResult ConstExpr;
+
+  IdentifierInfo *Ident = Tok.getIdentifierInfo();
+  SourceLocation BoundsKWLoc = Tok.getLocation();
+  ConsumeToken();
+
+  // Skip until right paren and do nothing if it is not range bounds expr.
+  // for example bounds(none, none + 5). Currently, it is NullaryBoundsExpr
+  if (!(cast<BoundsExpr>(Expr)->isRange())) {
```
dtarditi: You'll be able to delete this error recover  code if you make the change to ParseBoundsExpression that I recommend.

include/clang/Basic/DiagnosticParseKinds.td
```diff
@@ -1066,6 +1066,15 @@ def err_for_co_await_not_range_for : Error<
 def err_expected_bounds_expr : Error<
   "expected bounds expression">;
 
+def err_expected_range_bounds_expr : Error<
```
dtarditi: Is this new error message still used?

lib/Parse/ParseExpr.cpp
```diff
@@ -2905,9 +2910,67 @@ ExprResult Parser::ParseBoundsExpression() {
...
+    TypeResult ResultTy = ParseTypeName();
+    if (ResultTy.isInvalid()) {
+      SkipUntil(tok::r_paren, StopAtSemi | StopBeforeMatch);
+      IsError = true;
+    } else
+      RelativeClause = Actions.ActOnRelativeTypeBoundsClause(
+          BoundsKWLoc, ResultTy.get(), Tok.getLocation());
+  } else {
+    ConstExpr = Actions.VerifyIntegerConstantExpression(
+        ParseConstantExpression().get());
```
dtarditi: ParseConstantExpression could return an invalid result.   You need code of the form
```
ConstExpr = ParseConstantExpression();
if (!ConstExpr.isInvalid()) {
  ConstExpr = Actions.VerifyIntegerConstantExpression(ConstExpr.get());
}

File: lib/Parse/ParseExpr.cpp [hunk 16]
```diff
@@ -2905,9 +2910,67 @@ ExprResult Parser::ParseBoundsExpression() {
     SkipUntil(tok::r_paren, StopAtSemi | StopBeforeMatch);
 
   PT.consumeClose();
+
+  if (StartsRelativeBoundsClause(Tok) && !Result.isInvalid()) {
```
jijoongmoon: I think invalidate check of Result should be if(StartRelativeBoundsClause(Tok)). Cause the input parameter of ParseRelativeBoundsClause() is Result.get(). So... I will change this as below.

  if (StartsRelativeBoundsClause(Tok) ) {    // check next token is relative bounds clause.
       if(Result.isInvalid()) { // skip relative bounds clause even if it is invalid result.
       } else { //if it is valid, then
           if( Range Bounds Expression) { // check if it is range bounds expression.
              if( ParseRelativeBoundsClause(Result.get())) // if it is range bounds expression do parsing relative bounds clause
                   Diag(error);
            }else { //if it is not the range bounds expression 
               SkipUntile( r_paren); // skip relative bounds clause
            }
        }

dtarditi: It is  problematic to just call `SkipUntil` if Result is invalid because of a typechecking error.  `SkipUntil` does not handle nested parentheses correctly and could produce an incorrect syntax error.   I think this is a general issue with `SkipUntil` in clang.

Consider code of the form:
`bounds(e1, e2) rel_align_value((sizeof(T))`, where the programmer put an extra parenthesis around the body of the rel_align_value constant and `e1` or `e2` has a typechecking error.

lib/Parse/ParseExpr.cpp
jijoongmoon: I will delete this code.

lib/Parse/ParseExpr.cpp
jijoongmoon: I will change the code.

lib/Parse/ParseExpr.cpp
```diff
@@ -2905,9 +2918,101 @@ ExprResult Parser::ParseBoundsExpression() {
...
+      if (!ConstExpr.isInvalid())
+        RelativeClause = Actions.ActOnRelativeConstExprClause(
+            ConstExpr.get(), BoundsKWLoc, Tok.getLocation());
+    }
+  } else {
+    llvm_unreachable("unexpected relative alignment clause kind");
+  }
+
+  if (Expr.isInvalid()) {
+    SkipUntil(tok::r_paren, StopAtSemi | StopBeforeMatch);
```
dtarditi: I think you should just return 'true' here. We don't know why the Expr that was passed in was invalid, so we should not emit an error message. The programmer will need to fix the error that caused the invalid expression first.

lib/Parse/ParseExpr.cpp
```diff
@@ -2806,14 +2811,22 @@ ExprResult Parser::ParseInteropTypeAnnotation(const Declarator &D, bool IsReturn
...
-  if (StartsInteropTypeAnnotation(Tok))
-    return ParseInteropTypeAnnotation(D, IsReturn);
+  if (StartsBoundsExpression(Tok)) {
+    Result = ParseBoundsExpression();
+    isBoundsorItype = true;
+  }
 
-  Diag(Tok, diag::err_expected_bounds_expr_or_interop_type);
-  return ExprError();
+  if (StartsInteropTypeAnnotation(Tok)) {
```
dtarditi: I think you want this code to run only if `StartsBoundsExpression(Tok)` was false.

Otherwise we will parse something like:
bounds(e1, e2) itype(t)

Line 2820 can have the form:
```
} else if (StartsInteropTypeAnnotation(Tok))
```

lib/Parse/ParseExpr.cpp
dtarditi: (sorry for the duplicate comments about this line). I think you should just return 'true' here. We don't know why the Expr that was passed in was invalid, so we should not emit an error message. The programmer will need to fix the error that caused the invalid expression first.

lib/Parse/ParseExpr.cpp
```diff
@@ -2928,7 +3034,15 @@ bool Parser::ConsumeAndStoreBoundsExpression(CachedTokens &Toks) {
     return false;
   }
   ConsumeParen();
-  return ConsumeAndStoreUntil(tok::r_paren, Toks, /*StopAtSemi=*/true);
+  result = ConsumeAndStoreUntil(tok::r_paren, Toks, /*StopAtSemi=*/true);
+  if (Tok.getKind() == tok::identifier) {
+    Toks.push_back(Tok);
+    ConsumeToken();
+    Toks.push_back(Tok);
+    ConsumeParen();
```
dtarditi: Please guard the `ConsumeParen` before calling it.  It should be guarded the same way the prior `ConsumeParen` call is guarded:
```
if (Tok.getKind() != tok::l_paren) {
  return false;
}
```
ConsumeParen emits an error message if it is called when the token is not a parentheses.   We are storing tokens now to be parsed later, so we don't want error messages here.

lib/Parse/ParseExpr.cpp
```diff
@@ -2949,6 +3063,9 @@ ExprResult Parser::DeferredParseBoundsExpression(std::unique_ptr<CachedTokens> T
   ConsumeAnyToken();   // Skip past the current token to the new tokens.
   ExprResult Result = ParseBoundsExpression();
 
+  if (StartsRelativeBoundsClause(Tok))
```
dtarditi: I think this code needs to be changed  a bit.    We have a stream of tokens have the same "shape" as a bounds expression followed by relative alignment clause, but that may not actually be valid bounds expressions or relative alignment clauses.   There are four possible cases here:

1.  Result is invalid.
2.  The tok is not EOF and starts a relative bounds clause
3.  The tok is not EOF and does not start a relative bounds clause.
4. The tok is EOF.

For case 1, we want to consume the remaining tokens.
For case 2, we want to parse the relative alignment clause.  If there are any tokens left after parsing the relative alignment clause, we want to consume them.    We should have generated an error message while parsing the relative alignment clause.
For case 3, we need to issue an error message and then consume the rest of the tokens.
For case 4, we do not need to do anything.

dtarditi: Actually, I have figured out a simpler approach.   At line 3064, instead of calling ParseBoundsExpression, call ParseBoundsExpressionOrInteropType.    The root of the problem is that it is ParseBoundsExpressionOrInteropType that handle rel_align clause.

lib/Parse/ParseExpr.cpp
```diff
@@ -2806,14 +2811,22 @@ ExprResult Parser::ParseInteropTypeAnnotation(const Declarator &D, bool IsReturn
...
+  }
 
-  Diag(Tok, diag::err_expected_bounds_expr_or_interop_type);
-  return ExprError();
+  if (StartsInteropTypeAnnotation(Tok)) {
+    Result = ParseInteropTypeAnnotation(D, IsReturn);
+    isBoundsorItype = true;
+  }
+
+  SkipInvalidExprAndParseRelativeBoundsClause(Result, isBoundsorItype);
```
dtarditi: I think improving error recovery for bounds expressions is outside the scope of this pull request.  We should just be focused on parsing rel_align in this pull request.  Yes, a parsing error in a bounds expression can lead to poor handling of rel_align.   That is OK, we can fix that in a future check-in. 

 I'd like you to change the code as follows:
- Please remove SkipInvalidAndParseRelativeBoundsClause from this pull request.
- Instead, if the next token starts a relative bounds clause, call ParseRelativeBoundsClause.
- Otherwise,
-- if isBoundOrItype is false, emit an error message stating that we expected a bounds expression or interop type here.
-- if isBoundsOrIType is true, emit a different error message stating the token is unexpected; we expected rel_align or rel_align_value.
- if the token starts a relative .

lib/Parse/ParseExpr.cpp
```diff
@@ -2905,9 +2918,101 @@ ExprResult Parser::ParseBoundsExpression() {
     SkipUntil(tok::r_paren, StopAtSemi | StopBeforeMatch);
 
   PT.consumeClose();
+
   return Result;
 }
 
+void Parser::SkipInvalidExprAndParseRelativeBoundsClause(ExprResult &Result,
```
dtarditi: Please remove this function from this pull request.   I think improving error recovery for invalid bounds expressions should be in a separate pull request.

lib/Parse/ParseExpr.cpp
jijoongmoon: I will change it.

lib/Parse/ParseExpr.cpp
jijoongmoon: I will return true without skipping.

lib/Parse/ParseExpr.cpp
jijoongmoon: I will guide it before called.

jijoongmoon:
I think this pull request is ready for further review. I changed the error message with Token. I'm really appreciate for your review as always.

PR #178, opened 2017-03-07
Adds Linux & Mac Build Automation

lenary:
This should allow you to build on linux or mac. 

I put mac as a subdirectory of linux, so that it's obvious that mac just has one or two changes away from the linux install.

If you install cmake 3.7 on the machine, it won't download it for every build. Though, in fact, as long as you're always building in the same directory, everything should stick around between builds and just work.

When it comes time to extend where we pull from to other repos, `checkout.sh` is the file to change.

Also, I do no managing of build dirs etc, which I really should do better. We should talk about this.


lenary:
Did a build using this on Linux, it worked and ran all the tests (though some of the LLVM Test Suite tests didn't pass, as has been the case before). Merging

PR #179, opened 2017-03-08
Adds Tests for Dynamic Bounds Checks at Assignment

lenary:
This test checks both the AST and the IR for tests involving using `=` to assign into a checked array/checked `array_ptr`.

It effectively echoes the test for reading out of these arrays, only the check happens before the store. The pointer calculations remain exactly the same.


lenary:
Testing:
- Passes on Mac (64-bit)
- Passes on Windows, 32-bit and 64-bit. (A different test is failing, but this is the type problem I solved in a different PR)
- Not yet tested on linux, will do that after our meeting.

PR #186, opened 2017-03-10
Clears up Dynamic Check generation and adds methods for next steps

lenary:
We had lots of stats we weren't using, and some don't make any sense any more.

We can no longer tell the expression origin of a given check beyond its immediate expression such as `*e1`, `e1[e2]`, `e1.f`, `e1->f` or `dynamic_check()`, which means there's no way to record if the check came from an assignment, update, increment or read.

I've also added two stub methods for Member expressions (`e.f`) and Arrow expressions (`e->f`). Though these are not particularly different in the AST (both represented by MemberExprs), I think the logic may be different so I've split them apart.


lenary:
Tested on Mac, this gets rid of some Variable Unused warnings, and everything continues to pass.

Issue #189, opened 2017-03-13
Add checked scope


wonsubkim:
add checked scope
checked scope is program scope for checked pointer type (refer to ch 2.7)
we will implement first checked scope not unchecked scope

checked scope syntax
1. **checked** _compound_stmt_ ({... })
2. **checked** _function_declaration_

In checked scope, it is allowed to have checked pointer types, checked array types or unchecked pointer types, unchecked array types with bounds-safe interface

wonsubkim:
Requirement

- In checked scope, code is limited to use pointer types that are checked pointer types or unchecked pointer types with bounds-safe interface

-  In checked function, function declaration can have use checked pointer types for return type and function parameter types

- In checked function, function body is defined as checked scope by default

- In checked scope, function without prototype is not allowed to use

wonsubkim:
In checked function, checked keyword acts as new function specifier
checked function specifier is added in Declarator and it enforces to check if
function declarator has proper type or function definitions has proper statement type

In checked scope, checked keyword insert additional property into related scope
checked property is added Scope/CompoundScope

According to this checked property, Sema (Semantic) will check if 
declaration statement has proper checked pointer type
or function declarator has proper return/parameter type

dtarditi:
I have created work items for this feature: issue #199, issue #200, issue #201, and issue #202.

wonsubkim:
I have a question about work item management

The new created work items have different status in my thought
Some work items such as #199, #200 are related to current PR (#193)
I am now working on parsing of checked scope and declaration type restriction in #193 PR
and testing those with simple test code in checkedc
So, if #193 PR is completed, those two work items can be closed

Other work items such as #201, #202 are not yet implemented but works to do
How about your idea?

wonsubkim:
Related issues (#199, #200, #201, #202, #247, #251)

dtarditi:
The feature is complete.

PR #193, opened 2017-03-16
Add initial version of checked scope/function implementation

wonsubkim:
Implemented checked scope/function
checked function
 - checked keyword is before function definition
 - checked keyword is regarded as function specifier in implementation
 - checked function specifier affects function declaration scope & function definition scope
 - in function declaration, it checks if function parameter type & return type is proper for checked scope
 - in function definition, it checks if function definition statement body is regarded as checked compound scope

checked scope
 - checked keyword is before l_brace
 - this compound statment is regarded as checked block
 - in checked compound statments, all declaration can have checked pointer type or unchecked pointer type with bounds-safe interface
 - otherwise, compiler emits compilation error

Also relative alignment bug fix is added in this pull request


wonsubkim:
I also have added test code for simple version of checked scope test
Should I do pull request for this checked scope test code?

dtarditi:
Yes, please do a pull request for the test code.    Which repo do you plan to put the test code in?

wonsubkim:
Similar to relative alignment test code, I plan to put in checked repository (checkedc/test/typechecking/checked_scope.c)
If this PR is merged, I will do PR for test code

dtarditi:
Yes, that is a good place to put it.  It would be helpful to do the pull request for the test code now.   I like to review compiler changes and test for the changes at the same time.   It helps me understand if there is good testing of the compiler change.

We have to commit the compiler changes before the tests so that testing does not break.

wonsubkim:
I plan to start checked scope implementation with simple test code having checked function & scope
Okay, I will do PR for checkedc test code example with this PR
Thanks

wonsubkim:
I have done PR for test code for #189 
It is simple test code for testing checked function and checked scope
I have checked those simple test code

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8904,6 +8904,11 @@ def err_bounds_type_annotation_lost_checking : Error<
 	"%select{assignment|increment|decrement|call|volatile}0 expression not allowed in "
 	"%select{expression|dynamic check expression|count expression|byte count expression|bounds expression}1">;
 
+  def err_checked_scope_type : Error<
+    "invalid %select{parameter type|return type|variable type|member type}0 in checked scope or function">;
+
+  def err_checked_scope_no_prototype_func : Error<
```
dtarditi: I would reword this as "function without a prototype cannot be used or declared in a checked scope."

include/clang/Sema/DeclSpec.h
```diff
@@ -393,6 +395,8 @@ class DeclSpec {
       TQ_unalignedLoc;
   SourceLocation FS_inlineLoc, FS_virtualLoc, FS_explicitLoc, FS_noreturnLoc;
   SourceLocation FS_forceinlineLoc;
+  // Checked C - checked function
```
dtarditi: I would reword this comment as "//Checked C- checked keyword location"

lib/Parse/ParseDecl.cpp
```diff
@@ -2829,6 +2828,32 @@ void Parser::ParseDeclarationSpecifiers(DeclSpec &DS,
       AttrsLastTime = true;
       continue;
 
+    case tok::kw__Checked:
+      // Checked C - it is parsed as checked array or checked function keyword
+      // checked array type, checked []
+      if (NextToken().is(tok::l_square)) {
+        if (!getLangOpts().CPlusPlus11 || !isCXX11AttributeSpecifier())
```
dtarditi: The "checked" keyword is only recognized when the compiler is parsing C.  This means that !getLangOpts().CPlusPlus11() will always be true.  I suggest simplifying this block of code.  I think the code beneath the goto can be deleted, and the if-statement can be removed.

lib/Parse/ParseDecl.cpp
```diff
@@ -6236,7 +6274,8 @@ void Parser::ParseParameterDeclarationClause(
 
       // Inform the actions module about the parameter declarator, so it gets
       // added to the current scope.
-      ParmVarDecl *Param = Actions.ActOnParamDeclarator(getCurScope(), ParmDeclarator);
+      ParmVarDecl *Param =
```
dtarditi: This looks like only a formatting change not related to parsing checked scopes.  Please undo this change because it might make merging updated sources for clang harder.

lib/Parse/ParseStmt.cpp
```diff
@@ -854,7 +865,7 @@ StmtResult Parser::ParseCompoundStatement(bool isStmtExpr) {
 ///
 StmtResult Parser::ParseCompoundStatement(bool isStmtExpr,
                                           unsigned ScopeFlags) {
-  assert(Tok.is(tok::l_brace) && "Not a compount stmt!");
+  assert(Tok.is(tok::l_brace) && "Not a compound stmt!");
```
dtarditi: I agree that there is a typographical error here.   However, this change is unrelated to parsing checked keywords, so please undo it.  It might make future merges of code harder.

lib/Parse/ParseStmt.cpp
```diff
@@ -21,6 +21,7 @@
 #include "clang/Sema/PrettyDeclStackTrace.h"
 #include "clang/Sema/Scope.h"
 #include "clang/Sema/TypoCorrection.h"
+#include "clang/Sema/SemaDiagnostic.h"
```
dtarditi: Is this new include needed in this file?   I am not seeing changes to this file that seem to use the result of the include.

lib/Parse/ParseStmt.cpp
```diff
@@ -441,7 +450,7 @@ StmtResult Parser::ParseSEHTryBlock() {
     return StmtError(Diag(Tok, diag::err_expected) << tok::l_brace);
 
   StmtResult TryBlock(ParseCompoundStatement(/*isStmtExpr=*/false,
-                      Scope::DeclScope | Scope::SEHTryScope));
+                      (unsigned)(Scope::DeclScope | Scope::SEHTryScope)));
```
dtarditi: This change seems unrelated to parsing checked scopes.   Please undo it.

lib/Parse/ParseStmt.cpp
```diff
@@ -2068,8 +2086,8 @@ StmtResult Parser::ParseCXXTryBlockCommon(SourceLocation TryLoc, bool FnTry) {
     return StmtError(Diag(Tok, diag::err_expected) << tok::l_brace);
 
   StmtResult TryBlock(ParseCompoundStatement(/*isStmtExpr=*/false,
-                      Scope::DeclScope | Scope::TryScope |
-                        (FnTry ? Scope::FnTryCatchScope : 0)));
+                      (unsigned)(Scope::DeclScope | Scope::TryScope |
```
dtarditi: This change seems unrelated to parsing checked scopes. Please undo it.

lib/Parse/Parser.cpp
```diff
@@ -1096,8 +1107,12 @@ Decl *Parser::ParseFunctionDefinition(ParsingDeclarator &D,
       TemplateInfo.Kind == ParsedTemplateInfo::Template &&
       Actions.canDelayFunctionBody(D)) {
     MultiTemplateParamsArg TemplateParameterLists(*TemplateInfo.TemplateParams);
-    
-    ParseScope BodyScope(this, Scope::FnScope|Scope::DeclScope);
+    // Checked C - consider checked function definition
+    ParseScope BodyScope(this,
+                         Scope::FnScope | Scope::DeclScope |
+                             (D.getDeclSpec().isCheckedSpecified() || isChecked
+                                  ? Scope::CheckedScope
```
dtarditi: It looks this computation is repeated several times below (I am reviewing this via the GitHub interface which hides lines of code, so I might be wrong).   Instead of repeating the computation, please compute it once and store the result in a local variable.

lib/Sema/SemaChecking.cpp
```diff
@@ -10621,6 +10621,64 @@ void Sema::DiagnoseSelfMove(const Expr *LHSExpr, const Expr *RHSExpr,
...
+    TypeKind = 3; // member type
+    Ty = Field->getType();
+  }
+
+  bool checkedScopeError = false;
+  if (TargetDecl) {
+    // type is unchecked pointer w/o bounds-safe interface
+    // type is unchecked pointer & its bounds-safe interface is also unchecked
+    QualType InterOpTy = GetCheckedCInteropType(TargetDecl);
+    if ((Ty->isUncheckedPointerType() || Ty->isUncheckedArrayType()) &&
```
dtarditi: We will need a more complex way to determine whether a type uses an unchecked type.   We have to check what the pointer points to and whether that is an unchecked type.  We have a routine in Type.h that determines whether a type mentions a checked type.  We will need a similar routine for a checked type..

I think what you have here is a great first step.   I think adding more complex checking is beyond the scope of this pull request.   I have opened work item #200 to track this additional work.   I do not think you should extend this pull request to add this functionality - it recommend doing it separately.

include/clang/Basic/DiagnosticSemaKinds.td
wonsubkim: Okay, I will modify err_checked_scope_no_prototype_func error message
How about err_checked_scope_type error message?

include/clang/Sema/DeclSpec.h
wonsubkim: Okay, I will do it

lib/Parse/ParseStmt.cpp
wonsubkim: As I added additional argument into ParseCompoundStatement method to pass checked scope flag, it is necessary to distinguish from those two methods (ParseCompoundStatement(bool, bool) and ParseCompoundStatement(bool, unsigned).
So, for compilation purpose, I have added above casting

ParseCompoundStatement(bool isStmtExpr) & ParseCompoundSatatment(bool isStmtExpr, unsigned flag)

lib/Parse/ParseStmt.cpp
wonsubkim: Okay I will undo this

lib/Parse/Parser.cpp
wonsubkim: Okay, I will apply this

lib/Sema/SemaChecking.cpp
wonsubkim: Thanks for your suggestion
I will check this more

lib/Parse/ParseStmt.cpp
wonsubkim: Maybe it was related to previous work
I will check this. If possible, I will undo this

lib/Parse/ParseDecl.cpp
wonsubkim: In checked array, checked keyword is used before l_square (checked []).
In checked scope, checked keyword is used before l_brace (checked { ).
In checked function, checked keyword is used before another function specifier or type
To distinguish above two situations, I have used if-statement
In checked array case, it was same as before
In checked scope case, it is impossible case for this function declaration specifier
Otherwise, it is used as checked function specifier. In this case, I have set chekced function specifier in declaration

In checked array case, I will remove unnecessary code beneath the goto statement

dtarditi:
I did not realize that ParseCompoundStatement was being overloaded by this change.  Thank you for point that out.  I do not recommend using overloading in subtle ways like this.   It can lead to errors and problems later on.  I recommend adding isChecked as a 3rd argument to ParseCompoundStatement.  We can set the default argument value for "isChecked" to false.   However, we will still need to examine each call to ParseCompoundStatement to set whether isChecked needs to be propagated or not.  In the case where isChecked needs to be propagated and default arguments are being relied upon for other arguments, it is better to just make the other arguments explicit.  At least the code will be clear.

wonsubkim:
As you have mentioned, I have overloaded `ParseCompoundStatement(bool isStmtExpr)` function to pass **isChecked** argument.
After parsing checked scope keyword, I have passed isChecked result as argument of `ParseCompoundStatement(bool isStmtExpr, bool isChecked)
`And then, if isChecked is true, it calls `ParseCompoundStatement(bool isStmtExpr, unsigned scopeFlag) `call after inserting **CheckedScope** flag into scope flag
For this, I have added isChecked parameter into that function

According to your feedback, I will add isChecked parameter into ParseCompoundStatement function and then adjust scope flag within `ParseCompoundStatement(bool isStmtExpr, unsigned ScopeFlags, bool isChecked) `function

wonsubkim:
I have applied your feedback and slightly extended implementation to handle checked keyword for struct/union
According to your advice, I will improve further expression checking including C-style casting in a checked scope

include/clang/Basic/DiagnosticSemaKinds.td
dtarditi: Yes, I agree that the err_checked_scope_type need should be revised.   I think it would be useful to say something like "parameter|return|variable|member cannot have an unchecked pointer|array type in a checked scope or function"

lib/Parse/ParseStmt.cpp
```diff
@@ -228,8 +229,16 @@ Parser::ParseStatementOrDeclarationAfterAttributes(StmtVector &Stmts,
   case tok::kw_default:             // C99 6.8.1: labeled-statement
     return ParseDefaultStatement();
 
+  case tok::kw__Checked:
+    // Checked C - check if checked scope keyword
+    assert(NextToken().is(tok::l_brace));
```
dtarditi: How do we know that the next token will be an l_brace?

lib/Parse/ParseDecl.cpp
dtarditi: I think !getLangOpts().CPlusPlus11() will always be true at line 2835.

lib/Parse/ParseStmt.cpp
wonsubkim: In checked c specification, checked keyword can be used as
checked array, checked function, checked scope (compound or struct/union)

Since checked array & checked function is used in declaration, 
checked scope keyword is parsed in declaration
In this case, checked scope keyword is used before l_brace and function 

function (ParseStatementOrDeclarationAfterAttributes) is used to parse statement
so, I have expected that checked keyword is used only before l_brace
In that case, we prepare to parse next l_brace by calling parse compound statement
Otherwise, it is an error case

I agree your opinion. It is not always correct
To handle error case, I will change it

include/clang/AST/Expr.h
```diff
@@ -4814,13 +4814,24 @@ class RangeBoundsExpr : public BoundsExpr {
 private:
   enum { LOWER, UPPER, END_EXPR };
   Stmt *SubExprs[END_EXPR];
+  RelativeBoundsClause *DefaultClause;
   RelativeBoundsClause *RelativeClause;
+  /// Relative alignment base : bounds(lb, ub) rel_align(T)
+  /// compiler checkes whether sizeof(referent-type(base)) is common factor of
+  /// sizeof(T)
+  /// \brief BALIGN(base align) is sizeof(referent-type(base)))
+  /// \breif RALIGN(relative align) is from relative alignment clause
```
dtarditi: There is a typo: this should be \brief.

include/clang/AST/Expr.h
```diff
@@ -4814,13 +4814,24 @@ class RangeBoundsExpr : public BoundsExpr {
 private:
   enum { LOWER, UPPER, END_EXPR };
   Stmt *SubExprs[END_EXPR];
+  RelativeBoundsClause *DefaultClause;
   RelativeBoundsClause *RelativeClause;
+  /// Relative alignment base : bounds(lb, ub) rel_align(T)
+  /// compiler checkes whether sizeof(referent-type(base)) is common factor of
```
dtarditi: The semantics is slightly different: 
Normally, the compiler assumes that ub-lb is a multiple of sizeof(referentype(base)).
This lets the compiler generate more efficient code for bounds checks.  With
a rel_align clause, the compiler assume that ub-lb is a multiple of sizeof(T).

lib/Sema/SemaChecking.cpp
```diff
@@ -10621,6 +10621,73 @@ void Sema::DiagnoseSelfMove(const Expr *LHSExpr, const Expr *RHSExpr,
...
+  else if (const FieldDecl *Field = dyn_cast<FieldDecl>(Decl)) {
+    TargetDecl = Field;
+    DeclKind = 3; // member
+    Ty = Field->getType();
+  }
+
+  bool checkedScopeError = false;
+  int TypeKind = 0;
+  if (TargetDecl) {
+    // type is unchecked pointer w/o bounds-safe interface
```
dtarditi: The type could be an unchecked array too.  This can happen for global variables.  We will add a pragma at the top-level that makes everything be in a checked scope.

lib/Parse/ParseDecl.cpp
```diff
@@ -2829,6 +2828,22 @@ void Parser::ParseDeclarationSpecifiers(DeclSpec &DS,
       AttrsLastTime = true;
       continue;
 
+    case tok::kw__Checked:
+      // Checked C - it is parsed as checked array or checked function keyword
+      // checked array type, checked []
+      if (NextToken().is(tok::l_square)) {
+        if (!isCXX11AttributeSpecifier())
```
dtarditi: `!getLangOpts().CPlusPlus11` was true, so the if-statement 'if (!getLangOpts().CPlusPlus11 '' !isCXX11AttributeSpecifier()) goto DoneWithDeclSpec` should just be replaced with `goto DoneWithDeclSpec`.

lib/Sema/SemaChecking.cpp
```diff
@@ -10621,6 +10621,73 @@ void Sema::DiagnoseSelfMove(const Expr *LHSExpr, const Expr *RHSExpr,
...
+    // type is unchecked pointer & its bounds-safe interface is also unchecked
+    QualType InterOpTy = GetCheckedCInteropType(TargetDecl);
+    if ((Ty->isUncheckedPointerType() || Ty->isUncheckedArrayType()) &&
+        (InterOpTy.isNull() || (InterOpTy->isUncheckedPointerType() ||
+                                InterOpTy->isUncheckedArrayType()))) {
+      checkedScopeError = true;
+      TypeKind =
```
dtarditi: I think this should just be:

TypeKind = (Ty->isUncheckedPointerType()) ? 0 : 1

lib/Parse/ParseStmt.cpp
```diff
@@ -826,6 +831,22 @@ StmtResult Parser::ParseDefaultStatement() {
                                   SubStmt.get(), getCurScope());
 }
 
+/// ParseCheckedScopeStatement - Parse a 'checked {' statement
```
dtarditi: This looks great!  Thanks for the change.

include/clang/AST/Expr.h
wonsubkim: Sorry for mistake
This update related to BoundsClause is wrongly merged into current branch
By merging branches, I had some mistakes
I will exlucde this updates for this commit

include/clang/AST/Expr.h
wonsubkim: I will check it

lib/Sema/SemaChecking.cpp
wonsubkim: You mean that interop type SHOULD be checked type.
Okay I will change this

wonsubkim:
As I have commented about this issue, I apologize mistakes about wrong merge
To follow remote update, I have merged master branch into current branch but wrong commit is accidently merged into current branch
I will revert this update from current commit and apply your feedback into this commit
I did not realize this mistake :(
Sorry for bothering you

lib/Sema/SemaChecking.cpp
dtarditi: I think there is some confusion here.   I thought in the error message that we need to print whether the type was an unchecked pointer or unchecked array type.   I don't think the additional clause `(InteropType.isNull ...` is needed.   My suggestion is to simplify the code.

dtarditi:
It is not a problem about the merge issues.   Please let me know when the code is ready to review again.

lib/Sema/SemaChecking.cpp
wonsubkim: I have missed that interoperation type (itype) SHOULD require checked type
If interop type is not null, it is definitely checked type.
So, it is not necessary to check if interop type is checked or unchecked
I will simplify above codes

wonsubkim:
I am ready to review again

Issue #195, opened 2017-03-16
Avoid generating error message 'expected bounds expression' when there is a typechecking error


dtarditi:
Tests for rel_align show that sometimes we generate the error message "expected bounds expression before relative bounds clause" when there is a typechecking error in the bounds expression.   We should not generate this message because it will confuse programmers.   We should avoid generating this error message when the bounds expression is invalid.

dtarditi:
This is fixed.

PR #197, opened 2017-03-17
Avoid generating error message 'expeced bounds expression' when there…

jijoongmoon:
… is a typechecking error (#195)


jijoongmoon:
I think it is ready for the review! Thanks always.

Issue #188, opened 2017-03-11
Improve error recovery for bounds expressions parsing


dtarditi:
We do a poor job of error recovery when the keyword that starts a bounds expression is misspelled.   This leads to problematic error messages, particularly when rel_align is involved.   We should improve the error recover for bounds expressions.

jijoongmoon:
I also found that there is assert error after ParseBoundsExpressionOrInteropType() in DeferredParseBoundsExpression(). This assert is happened when the relative bounds clause is typo like below,
bounds(arr, arr+2) rel_alve(char).
After parsing the bounds expression, the ResultEpxr is not invalid but rel_alive is not the right.
This is because I push back without checking if the token is relative alignment or not.
I will fix it including this pull request.

jijoongmoon:
We also found the clang error situation in the code below:

void a(int c, int d) **_xxxxx_**{}
extern int cLan**_;_**
void f(){
  int b = **_cLan_**;
}

In this case, we can see that "undeclared identifier 'cLan' error, although we define it as extern.
This is because clang parser (SkipUntil) goes to the semi after the cLan when there is an error at function definition such as '**_xxxxx_**'.
There is same situation in rel_align as in code below.

array_ptr<int> f1(array_ptr<int>arr : count(5)) : bounds( arr, arr+1 ) **_rel_alive_**(char) {}
extern int cLan**_;_**
extern array_ptr<int> f : byte_count(**_cLan_** * sizeof(int));

We can see the "undefined error of cLan". Because parser skip to the nearest semi which is the one after cLan. The typo of rel_align "**_rel_alive_**" cause of it. That's why we can see undefined error. 
I wonder if we have to handle this kind of situation? I think it looks like clang parsing error rather than our parsing error situation.

dtarditi:
I believe we do better job now.  I looked at a variant of the code snippet that @jijoongmoon described:
```
_Array_ptr<int> f1(_Array_ptr<int> arr : count(5)) : bounds(arr, arr + 1) rel_alive(char) { return arr; }
extern int cLan;
extern _Array_ptr<int> f : byte_count(cLan * sizeof(int));
```
Here is the error message:
```
tmp.c:1:75: error: expected function body after function declarator
_Array_ptr<int> f1(_Array_ptr<int> arr : count(5)) : bounds(arr, arr + 1) rel_alive(char) { return arr; }
                                                                          ^
tmp.c:3:39: error: use of undeclared identifier 'cLan'
[...]

PR #226, opened 2017-03-21
Improve error recovery for bounds expression parsing (#188)

jijoongmoon:
. Add SkipInvalidBoundsExpr() to skip invalid bounds Expr such as bo0unds(), b00nd, bounnds(e1,e2), etc.
 . Fix generate Token Cache if there is relative bounds clause. Token is pushed back when there is rel_align or rel_align_value only.

### DIFF
--- a/include/clang/Parse/Parser.h
+++ b/include/clang/Parse/Parser.h
@@ -1674,6 +1674,8 @@ private:
 
   bool ParseRelativeBoundsClause(ExprResult &Expr);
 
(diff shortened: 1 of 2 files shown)


jijoongmoon:
I think it is ready to review. Thanks always.!

dtarditi:
Could you create a pull request for the Checked C repo that updates the error messages there?

jijoongmoon:
Yes. I will update the test code in Checked C repo.

PR #233, opened 2017-03-24
Adds dynamic non-null checks

lenary:
This commit adds non-null checks to Checked C. 

They are inserted during code generation of the following expressions:
- `*e` (checking `e` is non-null)
- `e->f` (checking `e` is non-null)
- `e1[e2]` (checking `e1` is non-null)

In particular, we call the original pointer that we calculate the offset from (in the case of subscript/arrow) the "base" pointer, and check it is non-null before performing any pointer offset calculations. These checks trap in exactly the same way that bounds checks trap, all containing their own trap block in LLVM IR. 

Testing:
- Not tested at all yet. Don't merge until there are some tests.

### DIFF
--- a/test/CheckedC/dynamic-checks/array-increment-code-gen.c
+++ b/test/CheckedC/dynamic-checks/array-increment-code-gen.c
@@ -41,6 +41,9 @@ void f1(void) {
   // CHECK-AST-NEXT: IntegerLiteral {{.*}} 'int' 0
 
   // CHECK-IR: [[REG1:%[a-zA-Z0-9.]*]] = load i32*, i32** @gp1
+  // CHECK-IR-NEXT: [[REG_DYNN:%_Dynamic_check.non_null[a-zA-Z0-9.]*]] = icmp ne i32* [[REG1]], null
+  // CHECK-IR-NEXT: br i1 [[REG_DYNN]], label %[[LAB_DYSUC:_Dynamic_check.succeeded[a-zA-Z0-9.]*]], label %{{_Dynamic_check.failed[a-zA-Z0-9.]*}}
+  // CHECK-IR: [[LAB_DYSUC]]:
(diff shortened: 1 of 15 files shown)


lenary:
I've finally written some tests. This should be ready for review, I'm going to add a note or two, but they're reflected in the source code.

test/CheckedC/dynamic-checks/array-increment-code-gen.c
```diff
@@ -400,6 +440,7 @@ void f3(void) {
   // CHECK-AST-NEXT: DeclRefExpr {{.*}} 'ga1' 'int checked[1]'
   // CHECK-AST-NEXT: IntegerLiteral {{.*}} 'int' 0
 
+  // NOTE: No Non-null check inserted
```
lenary: These are all constant-folded out.

test/CheckedC/dynamic-checks/struct-increment-code-gen.c
```diff
@@ -109,7 +115,11 @@ void f1(void) {
   // CHECK-AST-NEXT: IntegerLiteral {{.*}} 2
 
   // CHECK-IR: [[REG1:%[a-zA-Z0-9.]*]] = load %struct.S1*, %struct.S1** @gp3
+  // NOTE: no check here, because we don't yet do checks at pointer arithmetic
```
lenary: There's no non-null check here because `gp3 + 2` doesn't currently generate a non-null check of `gp3`. This is still an outstanding TODO, but will probably be clearer once we know where to insert the pointer arithmetic checks. 

I could look at adding them in this PR if you would prefer though.

lenary: Ok, I've had a look. They should go here: https://github.com/Microsoft/checkedc-clang/blob/master/lib/CodeGen/CGExprScalar.cpp#L2454-L2456

The reason this isn't trivial is that it turns out this is hit when generating the upper and lower bounds for the bounds check, which we don't want. A full fix is a different PR (which would prevent emitting non-null checks within a bounds check, as at any time we emit the bounds check, we have done a non-null check of the base ptr just before).

Issue #202, opened 2017-03-17
Implement restrictions on types in expressions in checked blocks.


dtarditi:
- No uses of variables or members that have unchecked pointer type and no bounds-safe interface.
- No C-style casts to unchecked pointer type.

wonsubkim:
when parsing expression statement, check if each element is unchecked type or not in checked scope
Also, unchecked pointer type casting is not allowed in checked scope

wonsubkim:
For constructed type checking, we have added constraints as follows:
- checked pointer to unchecked pointer is not allowed in checked (No use, No C-style type cast)
- variable arguments type is not allowed in checked block (No use, No C-style type cast)

dtarditi:
This work is complete.

Issue #201, opened 2017-03-17
Change types produced by address-of operator in checked blocks.


dtarditi:
This is describe in Section 2.4.1 of the Checked C specification.

wonsubkim:
I have a question about address-of operator
Let's look at below example to verify generated expression

```
int a checked[10];
array_ptr<int> b;
int i;

ptr<int> x = &a[i];
ptr<int> y = &b[0];
ptr<int> z = &i;
```
For current implementation, the address-of operator always produce unchecked pointer type
However, checked scope will change types according to checked property 
For example, `ptr<int> x = &a[i];`
Although a is checked array type, &a[i] produces unchecked pointer type and it generates default implicit cast operation after that
As a result of that, generated expression is as follows
**ImplicitCastExpr _Ptr<int> <BitCast> (UnaryOperator int * prefix &)**
If checked scope changes type as proper checked pointer type, expected result is as follows
**UnaryOperator _Ptr<int> prefix &**

Therefore, default implicit cast operator can be removed if possible
How about your opinion?

wonsubkim:
In specification (ch2.4.1), it describes a rule for the address-of operator as follows
_In checked block, if an address-of operator (&) applied to an lvalue expression, the address-of operator produces a value of type ptr<T>
Otherwise, the address-of operator produces a value of type (T *)_

I have some question about this
```
checked {
int a checked[10];
array_ptr<int> b;
int i;

ptr<int> x = &a[i]; 
ptr<int> y = &b[0];
ptr<int> z = &i;

array_ptr<int> ax = &a[i]; // wonder if address-of operator output is _Ptr or _Array_ptr
array_ptr<int> ay = &b[0]; // wonder if address-of operator output is _Ptr or _Array_ptr
array_ptr<int> az = &i; // think that address-of operator output is _Ptr
}
```
In assignment of ptr<int> x, y, z, since the result of address-of operator of lvalue is _Ptr<int>, there is no error
However, in next 3 assignments of array_ptr<int> ax, ay, az, there are incompatible errors
because address-of operator result is _Ptr<int> type but LHS is _Array_ptr<int>

However, a is checked array type and b is checked array pointer type
since a is checked array type, I think that address-of operator result is checked array pointer type
since b is checked pointer type, I think that address-of operator result is also checked array pointer type
In consequence, when address-of operator is applied to Lvalue, 

Can you explain this?

dtarditi:
[...]

PR #235, opened 2017-03-28
Add checked scope implementation for work-item (#201, #202)

wonsubkim:
PR for Initial version of checked scope/function implementation (#193) is already commited
PR for extended (#201, #202)  implementaton of checked scope/function 

1. Implicit cast rule has changed for address-of operation in checked scope
In checked scope, address-of operator produces checked array pointer (array_ptr<T>) instead of unchecked pointer type (array_ptr<T> not ptr<T>, this is also commented in related issue #201)
Therefore, implicit cast rule between checked pointer types is added
This changes is related to change of checkedc test code in checkedc project (checkedc project PR is also done)

2. implemented restrictions on type in expression in checked block (#202)
No uses of variable or members that have unchecked pointer type
For all uses of declaration, I have checked if used expression has unchecked pointer type in checked scope.

To prevent C-style casts to unchecked pointer type, I have modified related portion of clang

3. related error message is slightly revised


include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8874,6 +8874,15 @@ def err_bounds_type_annotation_lost_checking : Error<
   def err_expected_bounds : Error<
     "expression has no bounds">;
 
+  def err_expected_bounds_for_ptr_cast : Error<
+    "expression has no bounds, cast to ptr<T> require a source /w bounds">;
```
dtarditi: There's a typo in the error messages to correct: "require" should be "requires".    

Please don't abbreviate in the error messages: decl should be expanded to declaration the word "with" should be expanded and so on.

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8874,6 +8874,15 @@ def err_bounds_type_annotation_lost_checking : Error<
   def err_expected_bounds : Error<
     "expression has no bounds">;
 
+  def err_expected_bounds_for_ptr_cast : Error<
+    "expression has no bounds, cast to ptr<T> require a source /w bounds">; 
+
+  def err_expected_bounds_for_assignment : Error<
+    "expression has no bounds, lvalue /w bounds require a rvalue /w bounds">;
```
dtarditi: I worry that some C programmers won't understand this error message because they have an imprecise grasp of lvalue vs. rvalue terminology.  Is there to reword this?   For example, "right-hand side of assignment expected to have bounds because the left-hand side lvalue has bounds."

lib/Sema/SemaBounds.cpp
```diff
@@ -825,7 +825,7 @@ namespace {
           S.InferRValueBounds(S.getASTContext(), E->getSubExpr());
         if (SrcBounds->isNone()) {
           // TODO: produce more informative error message.
```
dtarditi: You can remove the TODO.

lib/Sema/SemaExpr.cpp
```diff
@@ -7401,7 +7410,19 @@ checkPointerTypesForAssignment(Sema &S, QualType LHSType, QualType RHSType) {
   // between checked and unchecked pointers or between different kinds of checked
   // pointers. Only implicit conversions between unchecked pointers or from
   // unchecked to checked pointers are allowed.
-  if (lhkind != rhkind && rhkind != CheckedPointerKind::Unchecked)
+
+  // Checked C - add rules for implicit casts between checked pointer types
+  // implicit casts between checked pointer type are not allowed
```
dtarditi: This comment is confusing.     Please describe what casts are allowed now and don't worry about describing what the code used to do.

lib/Sema/SemaExpr.cpp
```diff
@@ -10739,6 +10760,7 @@ QualType Sema::CheckAddressOfOperand(ExprResult &OrigOp, SourceLocation OpLoc) {
     OrigOp = op =
         CreateMaterializeTemporaryExpr(op->getType(), OrigOp.get(), true);
   } else if (isa<ObjCSelectorExpr>(op)) {
+    // Checked C - objective C is not related to checked c
```
dtarditi: You can remove this comment.

lib/Sema/SemaExpr.cpp
```diff
@@ -10866,7 +10888,19 @@ QualType Sema::CheckAddressOfOperand(ExprResult &OrigOp, SourceLocation OpLoc) {
 
   // In Checked C, the address-of operator always produces an unchecked pointer
   // type.
-  return Context.getPointerType(op->getType(), CheckedPointerKind::Unchecked);
+
+  // checked scope change types produced by unary address-of operator
```
dtarditi: Here is some rewording for the comment:

// Checked scopes change the types the address-of (&) operator.  In a checked scope, the
// operator poduces an array_ptr<T>.  In an unchecked scope, it continues to produce T *.

lib/Sema/SemaDecl.cpp
```diff
@@ -8527,10 +8527,16 @@ Sema::ActOnFunctionDeclarator(Scope *S, Declarator &D, DeclContext *DC,
     // checked function or checked scope
     // Checked C - in checked function, check return/param type
     // In checked scope, no prototype function is not allowed to use
-    DiagnoseCheckedDecl(NewFD);
+
+    // Checked C - check type restriction on declaration in checked scope
+    // if failed at function declaration, set invalid
```
dtarditi: I think you can remove the comments about setting things to be invalid.

lib/Sema/SemaChecking.cpp
```diff
@@ -10666,22 +10669,32 @@ void Sema::DiagnoseCheckedDecl(const DeclaratorDecl *Decl) {
     Ty = Field->getType();
   }
 
-  bool checkedScopeError = false;
+  bool Res = true;
```
dtarditi: Please expand this to "Result".

lib/Sema/SemaDecl.cpp
```diff
@@ -14205,8 +14211,10 @@ FieldDecl *Sema::ActOnField(Scope *S, Decl *TagD, SourceLocation DeclStart,
   FieldDecl *Res = HandleField(S, cast_or_null<RecordDecl>(TagD),
                                DeclStart, D, static_cast<Expr*>(BitfieldWidth),
                                /*InitStyle=*/ICIS_NoInit, AS_public);
-  if (S->isCheckedScope())
-    DiagnoseCheckedDecl(Res);
+  // Checked C - check type restriction on declaration in checked scope
+  // if failed at member declaration, set invalid
```
dtarditi: You can remove the comment about setting invalid.

dtarditi:
To clarify my review feedback: you *don't* need say things that are apparent from the code.

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8904,10 +8913,17 @@ def err_bounds_type_annotation_lost_checking : Error<
 	"%select{assignment|increment|decrement|call|volatile}0 expression not allowed in "
 	"%select{expression|dynamic check expression|count expression|byte count expression|bounds expression}1">;
 
-  def err_checked_scope_type : Error<
+  def err_checked_scope_type_for_declaration : Error<
     "%select{parameter|return|variable|member}0 cannot have an unchecked "
     "%select{pointer|array}1 type in a checked scope or function">;
 
+  def err_checked_scope_type_for_expression : Error<
```
dtarditi: There are some grammatical errors in this error message.  I suggest: "cannot use a %select{parameter|return|variable|member}0 with an unchecked type in a checked scope or function".

lib/Sema/SemaChecking.cpp
```diff
@@ -10666,22 +10669,32 @@ void Sema::DiagnoseCheckedDecl(const DeclaratorDecl *Decl) {
...
+      QualType InterOpTy = GetCheckedCInteropType(TargetDecl);
+      if (!Ty->hasCheckedType() && InterOpTy.isNull()) {
+        Res = false;
+        TypeKind = (Ty->isUncheckedPointerType() ? 0 : 1);
+      }
     }
   }
-  if (checkedScopeError) {
-    Diag(TargetDecl->getLocStart(), diag::err_checked_scope_type) << DeclKind
+  if (!Res) {
```
dtarditi: I think it is redundant to generate an error message at the declaration of a variable and the use of the variable, in the case where the variable was incorrectly declared to have an unchecked pointer type in a checked scope.  The additional error messages do not really give additional information to the programmer (they need to fix the original definition), and they definitely cause the programmer more work.

I'd suggest only generating an error message flagging a use of variable with unchecked pointer or array type in a checked scope if the variable was defined in a scope that was unchecked.

wonsubkim: Thanks for your suggestion
I consider how to handle error message for both use and declaration more efficiently.
If definition has an unchecked type error in a checked block, it produces error message for only declaration
Otherwise, it produces error message for each uses

wonsubkim:
If variable declaration has an unchecked type error in checked block, I have regarded that declaration as invalid declaration
For invalid declaration, clang parse & sema skips next semantic checking
Therefore, another expected error message is not generated

How about your opinion?

dtarditi:
I agree with marking the declaration invalid.  I think that is the right approach.

wonsubkim:
I am ready for review
This change is as follows

- I  have modified diagnostic error messages and comments

- I  have generated error message for use of variable/member only if it is declared correctly

- I have moved type restriction on declaration into ActOnVariableDeclarator where Sema actions for variable declaration

This change works well for checkedc test code (checked_scope.c) and check-clang

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8874,6 +8874,18 @@ def err_bounds_type_annotation_lost_checking : Error<
...
+    "expression has no bounds"
+    ", cast to ptr<T> expects source to have bounds">;
+
+  def err_expected_bounds_for_assignment : Error<
+    "expression has no bounds"
+    ", right-hand side of assignment expected to have bounds because the left-hand side has bounds">;
+
+  def err_expected_bounds_for_initializer : Error<
+    "expression has no bounds"
+    ", an initializer expected to have bounds because variable declaration has bounds">;
```
dtarditi: This error message is slightly grammatically incorrect.  I would reword it as "initializer expected to have bounds because the variable being declared has bounds."

lib/Sema/SemaCast.cpp
```diff
@@ -2557,6 +2557,17 @@ void CastOperation::CheckCStyleCast() {
       return;
     }
   }
+
+  // Checked C - No c-style casts to unchecked pointer/array type
```
dtarditi: Pleaase captitalize C in "C-style casts".

lib/Sema/SemaDecl.cpp
```diff
@@ -8521,16 +8526,20 @@ Sema::ActOnFunctionDeclarator(Scope *S, Declarator &D, DeclContext *DC,
     }
   }
 
-  // Checked C - checked function return type checking
+  // Checked C - type restriction on declarations in checked block
```
dtarditi: Please add periods 

// Checked C - type restriction on declarations in checked block.
// Function parameter & return are not allowed to use unchecked type
// in checked blocks.

lib/Sema/SemaDecl.cpp
```diff
@@ -8521,16 +8526,20 @@ Sema::ActOnFunctionDeclarator(Scope *S, Declarator &D, DeclContext *DC,
...
+  // function parameter & return type is not allowed to use unchecked type
+  // in checked blocks
   if (D.getDeclSpec().isCheckedSpecified()) {
-    // current scope - function definition/body scope, checked propery is from
-    // checked function or checked scope
-    // Checked C - in checked function, check return/param type
-    // In checked scope, no prototype function is not allowed to use
-    DiagnoseCheckedDecl(NewFD);
+    // current scope - function definition/body scope
+    // checked propery is from checked function or checked scope keyword
```
dtarditi: Typo: propery->property.   Also please add some periods here.  
I think there should be periods here:
// Current scope checked property is from checked function or
// checked scope keyword.
// In checked functions, check function/param types.
// In checked scopes, no prototypes functions are not allowed.

lib/Sema/SemaDecl.cpp
```diff
@@ -14205,8 +14214,10 @@ FieldDecl *Sema::ActOnField(Scope *S, Decl *TagD, SourceLocation DeclStart,
   FieldDecl *Res = HandleField(S, cast_or_null<RecordDecl>(TagD),
                                DeclStart, D, static_cast<Expr*>(BitfieldWidth),
                                /*InitStyle=*/ICIS_NoInit, AS_public);
-  if (S->isCheckedScope())
-    DiagnoseCheckedDecl(Res);
+  // Checked C - type restriction on declarations in checked block
```
dtarditi: Some typos:
// Checked C - type restrictions on declarations in checked block.
// Member declaration is not allowed to use unchecked type in checked block.

lib/Sema/SemaExpr.cpp
```diff
@@ -388,6 +388,19 @@ bool Sema::DiagnoseUseOfDecl(NamedDecl *D, SourceLocation Loc,
     Diag(D->getLocation(), diag::note_entity_declared_at) << D;
     return true;
   }
+
```
dtarditi: Please add periods at the ends of sentences in comments.

lib/Sema/SemaExpr.cpp
```diff
@@ -7401,7 +7414,12 @@ checkPointerTypesForAssignment(Sema &S, QualType LHSType, QualType RHSType) {
   // between checked and unchecked pointers or between different kinds of checked
   // pointers. Only implicit conversions between unchecked pointers or from
   // unchecked to checked pointers are allowed.
-  if (lhkind != rhkind && rhkind != CheckedPointerKind::Unchecked)
+  // Also, implicit conversion between checked pointers is added
+  // to handle properly address-of (&) operator in checked block
+  // Except for implicit conversion from checked to unchecked pointers,
+  // all other implicit conversions are allowed
+  if (lhkind != rhkind && rhkind != CheckedPointerKind::Unchecked &&
```
dtarditi: Can this just be simplified to rhkind != CheckedPointerKind::Unchecked &&
lhKind == CheckedPointerKind::Unchecked?   I think those two conditions imply that lhkind != rhkind, so that clause can be deleted.

include/clang/Basic/DiagnosticSemaKinds.td
wonsubkim: Actually, this modification makes changes in expected-error message in make check-clang
So, I just have added the reason why this expected bounds error occurrs at the end of "expression has no bounds"
After modifying this error message, I will also modify related expected error message

lib/Sema/SemaExpr.cpp
wonsubkim: Okay, you are right.
I did not check that. I will fix it

wonsubkim:
I am ready for review
I try to consider how to write informative/correct comments for changes

Issue #203, opened 2017-03-17
Add unchecked scopes


dtarditi:
The Checked C specification allow a programmer to declare that a block or scope allows the use of unchecked pointers:
```
unchecked {  ...
}
```
This is the default behavior for blocks, so it is not usually necessary to declare this.     This declaration is meant to be used in a checked scope where a programmer needs to drop into an unchecked scope.

This feature will have 2 work items:
- Parsing, including testing.
- Changes to semantic processing: clearing “checked” bit in scope.   Most of the work will be writing tests.

We will open the work items when work begins on this feature.

wonsubkim:
I will start to work this feature
unchecked scope parsing will be done in same manner with checked scope already implemented
unchecked scope works as follows:
- unchecked keyword is already defined, use it
- unchecked scope parsing is same as checked scope parsing
- clear checked property if parent scope has checked property, prevent inheritance of checked property

wonsubkim:
unchecked scope implementation is just to clear checked property of scope if parent scope has checked property
In unchecked scope test, I will check if parent's checked scope is properly cleared or not
- In checked function, clear checked property
- In checked block, clear checked property of parent scope
I will do PR both implementation and test

dtarditi:
This feature is complete.

PR #240, opened 2017-03-31
Add unchecked scope (#203) implementation

wonsubkim:
- unchecked scope keyword is already defined, use it

- unchecked scope parsing is same as checked scope parsing (syntax : unchecked '{')

- unchecked  scope keyword clears checked property of scope

- unchecked scope keyword overrides impact of checked function specifier for function body

To handle unchecked scope property, checked boolean value is replaced with enum value which represents checked scope kind

To clear checked property inherited from parent scope, unchecked scope flag is also added


include/clang/Parse/Parser.h
```diff
@@ -449,7 +458,7 @@ class Parser : public CodeCompletionHandler {
     if (Tok.getKind() == tok::l_brace)
       ++BraceCount;
     else if (BraceCount)
-      --BraceCount;     // Don't let unbalanced }'s drive the count negative.
+      --BraceCount;     // Don't let unbalanced '}'s drive the count negative.
```
dtarditi: Could you remove this change?   It is unrelated to implementing unchecked scopes?

wonsubkim:
I did not implement unchecked function specifier because I thought that function definition is always in unchecked scope.
Is unchecked function specifier necessary? Can you show example code?

dtarditi:
We intend to add a pragma at the top-level scope that makes everything afterward be considered checked by default (see Section 2.7 of version 0.6 of the Checked C specification).   The unchecked function specifier might be needed in that situation.

dtarditi:
If you did not implemented support for unchecked function declarations, I would recommend that we finish this pull request first.   I would suggest that you add the support in a separate pull request.

wonsubkim:
Okay, I have missed that pragma at the top-level scope
After this PR, I will do separate work-item. 
Thanks for your suggestion

wonsubkim:
I am ready for review

wonsubkim:
If possible, can you create work-item for this feature
- unchecked function specifier
- BOUNDS_CHECKED pragma to control top level scope checked property

I am now working on above work-items
As soon as possible, I will do PR for above items

dtarditi:
I have created the additional work items and assigned them to you.

wonsubkim:
I have implemented #246, #247 work-item on this PR
I have also added test code (checked_scope_pragma.c) in checkedc repos for #247 work-item
and related commit is also requested

To confirm unchecked function specifier, top-level checked scope pragma is also required
For unchecked function specifier test, pragma implementation is added for this PR.

I have also found a problem related to type restrictions on declaration by checking top-level translation unit checked property
Since interop type is defined after variable declaration, I have deferred checking of declaration after parsing bounds expression

dtarditi:
It is great to see that you have implemented #246 and #247.     Next time, though, please open a new pull request for that functionality.   I was about to merge your prior change and now I can't.    What you do is create a branch off your existing branch, and then issue a pull request for that.   When I commit the prior pull request, you merge main into the new branch, so that all the other differences go away.   The only issue with doing this is that if I request changes for the prior pull request, there could be conflict when you merge.

I will review this now with the additional changes.

Issue #204, opened 2017-03-17
Implement parsing and typechecking of new cast operators


dtarditi:
Section 5.1 of the Checked C specification describes new cast operators for converting between pointers with different kinds of checking.   It also describes changes to the existing C cast operator for checked pointer types.

Here are the proposed work items for this feature:
- Design for extending IR with new cast information.
- Implement IR extension for `dynamic_bounds_cast` and parse `dynamic_bounds_cast<T>(e1)`, `dynamic_bounds_cast<T>(e1, e2)` , and `dynamic_bounds_cast<T>(e1, e2, e3)`.
- Same as above but for `assume_bounds_cast`.
- Implement type checking rules for C-style casts involving checked pointer types.  Work includes writing tests.
- Implement type checking rules for `dynamic_bounds_cast`.  Work includes writing tests.
- Implement type checking rules for `assume_bounds_cast`.  Work includes writing tests.

jijoongmoon:
I have some questions about dynamic and assume bounds cast in spec.
According to spec, it says " dynamic bounds cast<T >(e1 ) converts e1 to either a ptr or *** type.**"
I wonder what might be the result after the dynamic bounds cast Done according to the e1. I thought the result of Dynamic/Assume bounds cast is Checked Type. But it says converts e1 to unchecked pointer type. 
 How can we handle these cases differently?
   `int *p;`
    `ptr<int> q = 0;`
    `ptr<int> r = 0;`
    `r = assume_bounds_cast<int>(q);`
    `p = assume_bounds_cast<int>(q);`

dtarditi:
The spec is unclear.   The type argument should be the "full pointer type".   In your example, it would be:
```
r = assume_bounds_cast<ptr<int>>(q);
p = assume_bounds_cast<int *>(q);`
```
This is the same way that the various C++-style cast operators work (they take the full target type).
It would be good to add some examples to the spec.   The need to improve the description of the bounds cast operators is  tracked by https://github.com/Microsoft/checkedc/issues/4.

jijoongmoon:
so, in the r case, assume_bounds_cast should generate _Ptr<_Ptr<int>> type,  and in case of p, assume_bounds_cast should generate _Ptr<int*>.  Am I right?
And r is OK and p supposed to be Incompatible type error. In case of r, assume bounds cast generate _Ptr<int> type, and talking about p, it also make _Ptr<int> and try to assign unchecked variable. That's why generate incompatible type error.
`int *p;`
[...]

PR #248, opened 2017-04-03
Add BoundsCastExpr to hold information to make Bounds Expression in S…

jijoongmoon:
…ema.(#204)

. assume_bounds_cast<T>(e1) => ptr<T> e1
. assume_bounds_cast<T>(e1, e2) => array_ptr<int> e1 : count(e2)
. assuem_bounds_cast<T>(e1, e2, e3) => array_ptr<int> e1 : bounds(e2,e3)

  - Add BoundCastExpr Class with BoundsCastKind (Invalid, Dynamic, Assume) and Expr e1, e2, e3.
  - Add Parser::ParseBoundsCastExpr in Parser::ParseCastExpr for the dynamic and assume cast expr.
  - Add ActOnBoundsCastExpr for the BoundsCastExpr in Sema/SemaCast.
  - Add "CK_UnCheckedToChecked" Casting Kind to identify bounds casting.
  - Bounds Expression is made and set as subExpr of CastExpr in SemaBounds.cpp.


jijoongmoon:
This is the initial commit for the Bounds Cast Expression 
 . dynamic_bounds_cast and assume_bounds_cast.
 . currently basic of assume_bounds_cast is working. 
I think it is ready to first review. Thanks always.

include/clang/AST/Expr.h
```diff
@@ -2962,6 +2962,77 @@ class CStyleCastExpr final
   friend class CastExpr;
 };
 
+class BoundsCastExpr final
+    : public ExplicitCastExpr,
+      private llvm::TrailingObjects<BoundsCastExpr, CXXBaseSpecifier *> {
+  SourceLocation LPLoc;
+  SourceLocation RPLoc;
+  Expr *BaseExpr = nullptr;
+  Expr *CountExpr = nullptr;
```
dtarditi: Instead of having separate fields for CountExpr and RangeExpr, I'd recommend one field for a BoundsExpr.

include/clang/AST/OperationKinds.def
```diff
@@ -327,6 +327,10 @@ CAST_OPERATION(AddressSpaceConversion)
 // Convert an integer initializer to an OpenCL sampler.
 CAST_OPERATION(IntToOCLSampler)
 
+// CK_UnCheckedToChecked - A conversion from a Unchecked pointer to checkedc
+// pointer
+CAST_OPERATION(UnCheckedToChecked)
```
dtarditi: This should be renamed to something else, such as PointerBounds.   The source operand could actually be a checked pointer type.  We might be casting from array_ptr<char> to array_ptr<int>, for example.

include/clang/AST/Expr.h
```diff
@@ -2962,6 +2962,77 @@ class CStyleCastExpr final
   friend class CastExpr;
 };
 
+class BoundsCastExpr final
+    : public ExplicitCastExpr,
+      private llvm::TrailingObjects<BoundsCastExpr, CXXBaseSpecifier *> {
+  SourceLocation LPLoc;
+  SourceLocation RPLoc;
+  Expr *BaseExpr = nullptr;
```
dtarditi: I don't think BaseExpr is necessary: BaseExpr is stored in the base class CastExpr.

include/clang/AST/RecursiveASTVisitor.h
```diff
@@ -2426,6 +2426,9 @@ DEF_TRAVERSE_STMT(CountBoundsExpr, {})
 DEF_TRAVERSE_STMT(NullaryBoundsExpr, {})
 DEF_TRAVERSE_STMT(RangeBoundsExpr, {})
 DEF_TRAVERSE_STMT(InteropTypeBoundsAnnotation, {})
+DEF_TRAVERSE_STMT(BoundsCastExpr, {
```
dtarditi: You'll need to define the children() iterator in BoundsCastExpr for this work.  This will be a little complicated to implement in C++ because one child is stored in the CastExpr parent and the other child will be the BoundsExpr stored in the cast operation.  You might want to concatenate two iterators to do this.

lib/AST/StmtPrinter.cpp
```diff
@@ -1903,6 +1903,13 @@ void StmtPrinter::VisitInteropTypeBoundsAnnotation(
   }
 }
 
+void StmtPrinter::VisitBoundsCastExpr(BoundsCastExpr *Node) {
```
dtarditi: I don't think this will print the correct output.    It won't be what was originally written.

lib/Parse/Parser.cpp
```diff
@@ -505,6 +505,8 @@ void Parser::Initialize() {
     Ident_itype = &PP.getIdentifierTable().get("itype");
     Ident_rel_align = &PP.getIdentifierTable().get("rel_align");
     Ident_rel_align_value = &PP.getIdentifierTable().get("rel_align_value");
+    Ident_dynamic_bounds_cast = &PP.getIdentifierTable().get("dynamic_bounds_cast");
```
dtarditi: This should start with an underscore and a capital letter, to avoid possible name clashes with existing names.    The spec needs to be updated to reflect this.

lib/Sema/SemaExpr.cpp
```diff
@@ -12449,6 +12449,39 @@ ExprResult Sema::CreatePositionalParameterExpr(unsigned Index, QualType QT) {
...
+  BaseExpr = Res.get();
+
+  checkUnusedDeclAttributes(D);
+
+  QualType castType = castTInfo->getType();
+  Ty = CreateParsedType(castType, castTInfo);
+
+  CheckTollFreeBridgeCast(castType, BaseExpr);
+
+  CheckObjCBridgeRelatedCast(castType, BaseExpr);
```
dtarditi: We don't support objective C with the Checked C extension.  This should not be necessary.

lib/Serialization/ASTReaderStmt.cpp
```diff
@@ -741,6 +741,14 @@ void ASTStmtReader::VisitCStyleCastExpr(CStyleCastExpr *E) {
   E->setRParenLoc(ReadSourceLocation(Record, Idx));
 }
 
+void ASTStmtReader::VisitBoundsCastExpr(BoundsCastExpr *E) {
```
dtarditi: This is missing the kind field.

lib/Serialization/ASTWriterStmt.cpp
```diff
@@ -716,6 +716,15 @@ void ASTStmtWriter::VisitCStyleCastExpr(CStyleCastExpr *E) {
   Code = serialization::EXPR_CSTYLE_CAST;
 }
 
+void ASTStmtWriter::VisitBoundsCastExpr(BoundsCastExpr *E) {
+  VisitExplicitCastExpr(E);
+  Record.AddSourceLocation(E->getLParenLoc());
+  Record.AddSourceLocation(E->getRParenLoc());
+  Record.AddStmt(E->getCountExpr());
+  Record.AddStmt(E->getRangeExpr());  
+  Code = serialization::EXPR_BOUNDS_CAST;
```
dtarditi: This is missing the kind field.

include/clang/AST/Expr.h
jijoongmoon: I think this means a lot. When I though I have to choices to implement this castings. One is build bounds expression after the parsing and the other is build bounds expression in CheckFunctionBodyBoundsDecls. and depending on how it is implemented, It would be decided if I make new Bounds Cast Class or just add bounds expression pointer in CStyleCast. Currently I implemented the first way, and that's why there are separate fields for CountExpr and RangeExpr. May be I should have chosen second option. I will change the code and update this pull request.

include/clang/AST/OperationKinds.def
jijoongmoon: I will change it

include/clang/AST/Expr.h
dtarditi: I think it is a good idea to have a separate class and to not combine this with CStyleCast.

lib/Parse/Parser.cpp
jijoongmoon: I see. I will add underscore ana a capital letter.

include/clang/AST/Expr.h
jijoongmoon: Ok. I will update code properly and ask review again. Thanks!

jijoongmoon:
I think it's ready to review. I will update the conflicts next round review. Thanks.

lib/Parse/ParseExpr.cpp
```diff
@@ -3007,6 +3016,107 @@ bool Parser::ParseRelativeBoundsClause(ExprResult &Expr) {
...
+    SkipUntil(tok::greater, StopAtSemi);
+
+  // if there is rel, parsing it.
+  // if(Tok.is(tok::comma)){
+  //   ConsumeToken();
+  //   Parsing Alignment
+  //   .....
+  // }
+
+  Declarator DeclaratorInfo(DS, Declarator::TypeNameContext);
```
dtarditi: I don't think you should be parsing a declarator here.   ParseTypeName already parses a declarator.

You can look at ParseInteropTypeAnnotation (line 2780, ParseExpr.cpp) to see an example of how to ParseTypeName can be used.

lib/Parse/ParseExpr.cpp
```diff
@@ -3007,6 +3016,107 @@ bool Parser::ParseRelativeBoundsClause(ExprResult &Expr) {
...
+
+  SourceLocation LParenLoc, RParenLoc;
+  BalancedDelimiterTracker T(*this, tok::l_paren);
+
+  if (T.expectAndConsume(diag::err_expected_lparen_after, Ident.getNameStart()))
+    return ExprError();
+  LParenLoc = T.getOpenLocation();
+
+  // Parsing e1, e2, e3
+  ExprResult CastExpr(true), CountExpr(true), RangeExpr(true);
```
dtarditi: I think it would be clearer to use E1, E2, and E3 for the variable names.   I think CastExpr, CountExpr, and RangeExpr could confuse people later.

lib/Sema/SemaExpr.cpp
```diff
@@ -12449,6 +12449,63 @@ ExprResult Sema::CreatePositionalParameterExpr(unsigned Index, QualType QT) {
...
+
+  checkUnusedDeclAttributes(D);
+
+  BoundsExpr *bounds = nullptr;
+  ExprResult Result(true);
+
+  ExprResult ResCount = CorrectDelayedTyposInExpr(CountExpr);
+  ExprResult ResRange = CorrectDelayedTyposInExpr(RangeExpr);
+
+  if (!ResCount.isUsable() && !ResRange.isUsable()) {
```
dtarditi: I think you should have the caller build the bounds expression and pass it in as an argument, instead of building it here.

lib/Sema/SemaExpr.cpp
```diff
@@ -12449,6 +12449,63 @@ ExprResult Sema::CreatePositionalParameterExpr(unsigned Index, QualType QT) {
...
+                             ResCount.get(), SourceLocation());
+
+  } else if (ResCount.isUsable() && ResRange.isUsable()) {
+    Result = ActOnRangeBoundsExpr(SourceLocation(), ResCount.get(),
+                                  ResRange.get(), SourceLocation());
+  } else {
+    Result = ActOnNullaryBoundsExpr(SourceLocation(), BoundsExpr::Kind::None,
+                                    SourceLocation());
+  }
+  if (Result.isInvalid())
```
dtarditi: You should build an invalid bounds expression.   There is a helper method in Sema to do this:  CreateInvalidBoundsExpr().   This will suppress downstream errors in bounds checking etc.

jijoongmoon:
I think it's ready to further review. Thanks always.

jijoongmoon:
Without it, there is error in the case of assume_bounds_cast< int *>. such as "expected '>' and expected expression.

dtarditi:
Calling ParseDeclarator is not the right thing to do.    The error suggests that ParseTypeName() is consuming the '>' after the actual type name.   Sometimes clang has error handling code that does the wrong thing when given the slight novel syntax for Checked C.   I would suggest that you try to figure out whether the '>' is being consumed.   A first step is to see what the current token is after calling ParseTypeName.

lib/Parse/ParseExpr.cpp
```diff
@@ -3007,6 +3016,90 @@ bool Parser::ParseRelativeBoundsClause(ExprResult &Expr) {
...
+  else
+    kind = BoundsCastExpr::Kind::Invalid;
+
+  DeclSpec DS(AttrFactory);
+
+  if (ExpectAndConsume(tok::less, diag::err_expected_less_after,
+                       Ident.getNameStart()))
+    return ExprError();
+
+  ParseSpecifierQualifierList(DS);
```
dtarditi: You should be calling ParseTypeName here.

dtarditi: A type name in C is not what you might expect it to be.   It is actually the name of a type, not just an identifier.  See Section 6.7.7 of the C11 specification.    You should be calling ParseTypeName and then checking for a comma.  You should be calling ParseSpecifierQualifierList, checking for a comma, and then parsing a declarator.

dtarditi: I just updated the prior comment to insert a missing "not".

jijoongmoon: I see. I will update the code. Thanks.

jijoongmoon:
I think it is ready to review. Thanks always!

Issue #247, opened 2017-04-03
Implement BOUNDS_CHECKED pragma


dtarditi:
This allows us to make the top-level scope of a compilation unit checked.

wonsubkim:
#pragma BOUNDS_CHECKED on/off/default
If value is on, top level scope of compilation unit is defined as checked scope
If value is off or default, top level scope is defined as unchecked scope.

I think that by default top level scope is unchecked scope. Therefore, value off is same as value default

Is there any different between default and off?

dtarditi:
Yes, it makes sense for the default state to be "off".   Could you update the spec in the Checked C repo to describe this?

wonsubkim:
Yes, I can

dtarditi:
This work is complete.

PR #250, opened 2017-04-04
Implement BOUNDS_CHECKED pragma (#247)

wonsubkim:
+ fixed typo
  + #pragma BOUNDS_CHECKED [on|off], default value is off
  + fix bugs related multiple setting of BOUNDS_CHECKED pragma value
  + release related test code (checked_scope_pragma.c), passed


wonsubkim:
Not only top-level scope but also normal scope is influenced by #pragma BOUNDS_CHECKED on/off
Corresponding test code is also released

Issue #251, opened 2017-04-04
Do not allow calls to variable argument functions in checked blocks


dtarditi:
We currently have no way to enforce that arguments with correct types are passed to variable argument functions such as printf.   For now, we should not allow calls to variable argument functions in checked blocks.

wonsubkim:
In checked blocks, a variable argument function call is not allowed.
How about a variable argument function declaration?
A variable argument function declaration SHOULD be prohibited?
A variable argument standard library function prototype SHOULD be prohibited if #pragma BOUNDS_CHECKED on ?

wonsubkim:
I am little bit confused that variable argument function is prohibited in checked blocks.
For a variable arguments, what is correct type of argument?

In checked blocks, only checked type argument can be used as actual argument.
If function expects a checked pointer type, it is not a problem.
Although function expects an unchecked pointer type, I think that it is not problem if actual argument is a checked type.

dtarditi:
This work is complete.

PR #252, opened 2017-04-05
Implement variable argument function call issue in checked blocks (#251) & modify BOUNDS_CHECKED pragma (#247)

wonsubkim:
+ Prevent variable argument function call in checked blocks (#251)
  + In checked block, it does not allow call to variable argument function

+ Modify implementation BOUNDS_CHECKED pragma (#247)
  + For elaborate scope handling, implementation is changed to use annotation token
  + In previous implementation, BOUNDS_CHECKED pragma is handled in incorrect scope
  When parsing '#pragma BOUNDS_CHECKED', it is processed with both '{' and '}'
  When entering/exiting scope, it is not combined with proper scope
  To bind BOUNDS_CHECKED pragma with accurate scope, it generates annotation token for pragma
  When consuming annotation token(tok::annot_pragma_bounds_checked), it can recognize scope corresponding to pragma
  Annotation token pragma is also used in some other pragma handling
  + #pragma BOUNDS_CHECKED [ON/OFF/DEFAULT]
  + corresponding tests are also commited


wonsubkim:
BOUNDS_CHECKED pragma (#247) error case was found by extending checked_scope_pragma.c
checked scope property is sometimes incorrectly set because BOUNDS_CHECKED pragma is processed with brace token. Before parsing brace token, it is processed together.
For elaborate process of pragma, I have generated annotation token after parsing pragma
This annotation token is independently parsed after l_brace or r_brace
Therefore, we can handle accurate scope property

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8928,6 +8928,9 @@ def err_bounds_type_annotation_lost_checking : Error<
 
   def err_checked_scope_no_prototype_func : Error<
     "function without a prototype cannot be used or declared in a checked scope">;
+
+  def err_checked_scope_no_variadic_func : Error<
+    "call to variable argument function cannot be used in a checked scope">;
```
dtarditi: I suggest changing the wording from "cannot be used" to "cannot be made"

lib/Parse/ParsePragma.cpp
```diff
@@ -2204,39 +2214,22 @@ void PragmaMSIntrinsicHandler::HandlePragma(Preprocessor &PP,
 }
 
 // Handle the checked-c top level scope checked property.
-// #pragma BOUNDS_CHECKED on
-// #pragma BOUNDS_CHECKED off
+// #pragma BOUNDS_CHECKED [on-off-swtich]
```
dtarditi: Typo: swtich -> switch.

lib/Sema/SemaExpr.cpp
```diff
@@ -394,11 +394,18 @@ bool Sema::DiagnoseUseOfDecl(NamedDecl *D, SourceLocation Loc,
   // type in checked block.
   // Otherwise, it does not emit redundant error message at the use of variable
   // since it already produced an error message at the declaration of variable.
+  // Calls to variable argument functions are not allowed in a checked block.
```
dtarditi: You can declare a function pointer that takes variable arguments and has a variable argument type.   I think you need check this at the type level too, just as you a check for uses of unchecked pointer types.

This makes me wonder about what we should do when you have constructed types in a checked block that use unchecked pointer types.  For example, you could have a pointer to an unchecked pointer.    Checked code could never use the unchecked pointer, but it could pass it around.

dtarditi: I opened a Checked C spec issue about constructed types that use unchecked pointer types: https://github.com/Microsoft/checkedc/issues/142.   Feel free to comment on it there.

wonsubkim:
For constructed type & variable arguments type, I have modified
I have implemented recursive unchecked type checking & variadic type checking
Related test code is also released

wonsubkim:
I am ready for review  :)

PR #255, opened 2017-04-06
Elevated execution bug

lenary:
This bug is totally my fault for not rigorously testing clang on windows, especially windows x86.

It turns out windows hates 32-bit executables which contain the words "install", "setup" or "update" in their filename, and requires they run with elevated permissions. https://msdn.microsoft.com/en-us/enus/library/aa905330.aspx http://stackoverflow.com/questions/11573444/why-is-windows-asking-for-system-administrator-privileges-for-running-executable

This means when we compile and run some tests with "update" in the name, through Visual Studio, the tests fail as it can't execute the files concerned. This PR just renames the files and adds some documentation.


lenary:
The `-opt.c` files are in the other repo, which are where the main failures were. These changes were just to have some parity and documentation.

Issue #249, opened 2017-04-04
Finish implementing bounds inference


dtarditi:
To do:
- When we have a bounds expression on a field member, it is a member bounds expression.  It needs to be transformed into a regular bounds expression:
```
struct S {
   array_ptr<int> f : count(len);
   int len;
}

struct S a;
 ... a.f ...  // needs to transform this to have bounds count(a.len)
```
- Infer bounds for literal expressions.

lenary:
Here's the code that doesn't ICE, but is generated with only incidentally correct bounds: https://github.com/Microsoft/checkedc-llvm-test-suite/blob/master/MultiSource/Benchmarks/Ptrdist/anagram/anagram.c#L526-L527

This is the AST of an expression that ICE's in Olden/mst:
```
ArraySubscriptExpr 0x10408be60 'HashEntry':'_Ptr<struct hash_entry>' lvalue
|-Bounds
| `-RangeBoundsExpr 0x10408c4a8 'NULL TYPE'
|   |-ImplicitCastExpr 0x10408c460 '_Array_ptr<HashEntry>' <LValueToRValue>
|   | `-MemberExpr 0x10408bdb8 '_Array_ptr<HashEntry>' lvalue ->array 0x104088c80
|   |   `-ImplicitCastExpr 0x10408bd98 'Hash':'_Ptr<struct hash>' <LValueToRValue>uite/MultiSource/Benchmarks/Olden/mst/hash.c
|   |     `-DeclRefExpr 0x10408bd70 'Hash':'_Ptr<struct hash>' lvalue Var 0x10408af00 'retval' 'Hash':'_Ptr<struct hash>'
|   `-BinaryOperator 0x10408c480 '_Array_ptr<HashEntry>' '+'
|     |-ImplicitCastExpr 0x10408c460 '_Array_ptr<HashEntry>' <LValueToRValue>
|     | `-MemberExpr 0x10408bdb8 '_Array_ptr<HashEntry>' lvalue ->array 0x104088c80
|     |   `-ImplicitCastExpr 0x10408bd98 'Hash':'_Ptr<struct hash>' <LValueToRValue>
|     |     `-DeclRefExpr 0x10408bd70 'Hash':'_Ptr<struct hash>' lvalue Var 0x10408af00 'retval' 'Hash':'_Ptr<struct hash>'
|     `-ImplicitCastExpr 0x104088f08 'int' <LValueToRValue>
[...]

PR #258, opened 2017-04-07
Infer bounds for struct member accesses.

dtarditi:
Infer bounds for uses of struct members that have member bounds declarations or that are checked arrays.  The main implementation challenge is that the member bounds have to be transformed from using members to using member accesses.  We use a tree transform class to do this.  The address issue #249.

This change implements bounds inference for rvalues produced by member references and bounds inference for lvalues produced by member inferences.

-  Struct member accesses in contexts that expect rvalues will produce rvalues. These rvalues need bounds.   An example of this is:
```
struct S {
  array_ptr<int> p : count(len);
  int len;
}

struct S tmp;
array_ptr<int> p : count(tmp.len) = tmp.p;  // need bounds for tmp.p
```
- In other cases, we need the bounds for the lvalues produced by struct member accesses.   Examples of this are:
```
struct  S {
   int a checked[5];
}

struct S tmp;
tmp.a[3] = ...  (we need the bounds for tmp.a).
```
or
```
struct S { 
  int f;
}

struct S tmp;
  ...   &tmp.f  ...
```
In the first case, the reference to `a` produces an lvalue that is converted from array type to pointer type, but not dereferenced.

Testing:
- Add clang test of inference of bounds for reads of structure members and assignments to struct members.
- Update existing clang tests that use structure members to expand the information from 'Any' to
  the actual bounds.
- Add dynamic test to the Checked C repo that reads through a pointer stored in a structure member.
- More comprehensive testing is needed.    This changes unblocks other people, so committing it now.
main implementation work is that the


lenary:
I don't see any glaring issues, let's get it merged and I'll file bugs if I find issues in it.

PR #265, opened 2017-04-11
Switch to x64-hosted tools for building clang on Windows.

dtarditi:
During automated testing, the MSVC linker is running out of memory while
compiling clang or other programs that link with clang on Windows.  Switch to
the x64 toolset for Visual Studio to fix this problem.
- Update the documentation for how to set up a build environment for Checked C.
- Update the Windows automation scripts to pass the `-T "host=x64"` option to
   CMake.
  This is a new feature available with version 3.8.

Clang doesn't use DLLs on Windows. All programs are instead compiled with the
LLVM library statically linked as part of the executable.  By default the
32-bit MSVC toolset is used when compiling clang. This appears to be putting too
much memory pressure on the linker, perhaps in combination with ASLR causing
fragmentation of the 32-bit address space. The result is a fairly regular
pattern of automated testing failures.

Testing:
- Locally ran automation script on Windows to verify that the the new
  cmake option works.
- Passed the Checked C regression tests with this option.


lenary:
Great! We should probably update the linux builders to clang 3.8 so that we don't get any odd errors from the difference in the two make systems.

Issue #262, opened 2017-04-11
Implement Parsing Relative Bounds Clause in Bounds Cast Expression


jijoongmoon:
According to the specification about Assume and Dynamic bounds cast, Relative Bounds Clause can be used after Type.

dtarditi:
This work is complete.

PR #266, opened 2017-04-11
Add Parsing Relative Bounds Clauses in Bounds Cast Expression.(#262)

jijoongmoon:
In order to parse relative bounds clause in bounds cast expression, change ParseRelativeBoundsClause().


jijoongmoon:
I think it is ready to review! Thanks always.

lib/Parse/ParseExpr.cpp
```diff
@@ -2958,8 +2958,8 @@ ExprResult Parser::ParseBoundsExpression() {
   return Result;
 }
 
-bool Parser::ParseRelativeBoundsClause(ExprResult &Expr) {
-  bool IsError = false;
+RelativeBoundsClause *Parser::ParseRelativeBoundsClause(ExprResult &Expr,
```
dtarditi: I think this interface is too complicated now.   The function is doing too many things. 

I  suggest that you remove ExprResult as an argument and replace it with a Boolean flag that indicates whether this method should print an error message.   You can move the code that sets the bounds property of the ExprResult to the caller.

Then the function becomes easier to explain:
- It tries to parse a relative bounds expression.  It returns the relative bounds expression if successful and null if there was an error.
- If it cannot parse a relative bounds expression, it optionally prints an error message that it expected a relative bounds expression, if the new flag says it should.

lib/Parse/ParseExpr.cpp
```diff
@@ -2974,23 +2974,20 @@ bool Parser::ParseRelativeBoundsClause(ExprResult &Expr) {
   if (PT.expectAndConsume(diag::err_expected_lparen_after,
                           Ident->getNameStart())) {
     SkipUntil(tok::r_paren, StopAtSemi | StopBeforeMatch);
-    IsError = true;
-    return IsError;
+    return RelativeClause;
```
dtarditi: I think error recovery in this situation should be left to the caller.

jijoongmoon:
I think it is ready to further review. Thanks always.

lib/Parse/ParseExpr.cpp
```diff
@@ -3041,7 +3050,24 @@ ExprResult Parser::ParseBoundsCastExpression(IdentifierInfo &Ident,
...
     ConsumeToken();
-    //...
+    if (StartsRelativeBoundsClause(Tok)) {
+      IdentifierInfo *Ident = Tok.getIdentifierInfo();
+      bool isError = false;
+      ConsumeToken();
+      BalancedDelimiterTracker PT(*this, tok::l_paren);
+      if (PT.expectAndConsume(diag::err_expected_lparen_after,
+                              Ident->getNameStart())) {
+        SkipUntil(tok::r_paren, StopAtSemi | StopBeforeMatch);
```
dtarditi: I think you want to skip until you hit tok::greater, not a tok::r_paren.

lib/Parse/ParseExpr.cpp
```diff
@@ -3041,7 +3050,24 @@ ExprResult Parser::ParseBoundsCastExpression(IdentifierInfo &Ident,
...
+      } else {
+        RelativeClause =
+            CreateRelativeBoundsClause(isError, Ident, Tok.getLocation());
+        if (isError)
+          SkipUntil(tok::r_paren, StopAtSemi | StopBeforeMatch);
+      }
+      PT.consumeClose();
+    } else {
```
dtarditi: Generally the clang code avoid putting braces around single statements in else branches.   The code elsewhere has the from:
```
else 
   SkipUntil (tok::greater, ...
```

lib/Parse/ParseExpr.cpp
```diff
@@ -2959,38 +2959,57 @@ ExprResult Parser::ParseBoundsExpression() {
...
+  } else {
+    Diag(TempTok, diag::err_expected_range_bounds_expr);
   }
 
+  PT.consumeClose();
+  return isError;
+}
+
+RelativeBoundsClause *
+Parser::CreateRelativeBoundsClause(bool &isError, IdentifierInfo *Ident,
```
dtarditi: This should probably be called ParseRelativeBoundsClause, as this is what this method does.

lib/Parse/ParseExpr.cpp
```diff
@@ -2959,38 +2959,57 @@ ExprResult Parser::ParseBoundsExpression() {
 }
 
 bool Parser::ParseRelativeBoundsClause(ExprResult &Expr) {
```
dtarditi: I'd recommend renaming this method to avoid confusion.  It could be called ParseRelativeBoundsClauseForDecl (or something like that).

jijoongmoon:
I think it is read for further review. Thanks always.

Issue #260, opened 2017-04-10
Follow-up issues for parsing *_bounds_cast operators.


dtarditi:
It is important to be defensive when programming in a large code base.  People can introduce things like memory corruptions that cause programs to operate incorrectly.  By having defaults and cover all cases (even if we think they are logically impossible), we catch problems quicker and itis easier to debug problems when they do occur.  That means covering all the possible cases for enums in switch statements or conditional tests.  In addition, we should also have `default` cases for switch statements and avoid silently falling off the end of `if`-chains..

- In ASTDumper.cpp around line 1966, I think you should introduce an else statement that handles the situation where Kind does not have an expected value.
```
if (kind == .. Assume)
else if (kind ==  ... Dymaic)
else {
  OS << "Ilegal_bounds_cast_kind"
}
```
It's possible that there would be an error inside the compiler that results in an invalid kind.  We would want to know that when we dump the AST.
- I think a similar check for an unexpected case in StmtPrint.cpp is needed (around line 1907).

I think the `children()` method for the `BoundsCastExpr`class needs to include the base expression (Op, which is returned by CastExpr.getSubExpr()).   Children is supposed to return an iterator over all the child nodes in the AST of the current node.  The definition of the clang IR is somewhat unclear about this.  If you look at all the places where the method is called, it is clear that they expected all AST child nodes to be included

dtarditi:
Also, we should check that the type arguments to bounds_cast operators are pointer types.

jijoongmoon:
When I looked into the VisitCastExpr in ASTDumper, there is also print out about the bounds expression followed by "Inferred" string and if the CastExpr has bounds. and we also have children in CastExpr as Stmt* SubExprs[2], one for the op and one for the bounds. So I wonder that we may do not want to dump for the bounds in ASTDumper.  Unless it prints the bounds twice, one in the ASTDumper, and one in the children iterator.

dtarditi:
Yes, I agree that we should dump the bounds only once for a BoundsCastExpr in ASTDumper.  Given your observation, I would suggest (1) not dumping the inferred bounds for a bound cast expr in VisitCastExpr (2) not doing anything special to dump the bounds for in BoundsCastExpr.   The bounds will be included as one of the children.   Does this seem good?

dtarditi:
I fixed the issue with the children method in BoundsCastExpr with PR #263.

dtarditi:
Addressed by commit 944b36 in PR #269.

PR #269, opened 2017-04-13
Follow-up issues for parsing *_bounds_cast operators (#260)

jijoongmoon:
- Add CheckBoundsCast() in CastOperation for the static checking of Bounds Cast Expression.
  . Add checking whether the SrcExpr is pointer type and if it is not the pointer type then assign invalid bounds expression.
  . Assign invalid bounds expression if DestType is unchecked pointer type.
  . Check and convert ValueKind of SrcExpr.
  . Set Kind as CK_PointerBounds.
  . will be added more check.


jijoongmoon:
I think it is ready to review. Thanks always.

lib/Sema/SemaCast.cpp
```diff
@@ -2612,6 +2613,24 @@ void CastOperation::CheckCStyleCast() {
   }
 }
 
+void CastOperation::CheckBoundsCast(BoundsExpr *&Bounds,
```
dtarditi: - Bounds does not need to be by-ref parameter.  I think you can just return it.
- The source type can be an integral type or pointer type if this is an assume_bounds_cast.

include/clang/AST/Expr.h
```diff
@@ -2884,7 +2884,7 @@ class CastExpr : public Expr {
     // in the AST because they are part of the source program.  For other
     // cast expressions, the bounds are not included because they are
     // inferred by the compiler only when needed.
-    if (getStmtClass() == BoundsCastExprClass)
+    if (SubExprs[BOUNDS])
```
dtarditi: Please undo this change.    This could incorrect behavior if this is an implicit or explicit cast expression and we've inferred the bounds.  They inferred bounds (which were computed by the compiler) should not be included in the AST children.

lib/Sema/SemaBounds.cpp
```diff
@@ -932,6 +938,14 @@ namespace {
         return true;
       }
 
+      if (CK == CK_PointerBounds) {
```
dtarditi: This is only needed if this is a dynamic_bounds_cast.   It is OK for the soure expression to not have a bound if it is an assume_bounds_cast operation.

jijoongmoon:
I think it is ready for further review!. thanks always!

jijoongmoon:
I think it is ready for further review. Thanks.

lib/Sema/SemaBounds.cpp
```diff
@@ -613,14 +614,21 @@ namespace {
...
             return CreateBoundsNone();
           }
+
           Expr *subExpr = CE->getSubExpr();
           BoundsExpr *Bounds = CE->getBoundsExpr();
-          QualType QT = subExpr->getType();
-	  Expr *Base = subExpr;
-          if (subExpr->getStmtClass() != clang::Stmt::CallExprClass)
-	    Base=CreateImplicitCast(QT, CastKind::CK_LValueToRValue, subExpr);
+
```
dtarditi: You don't need lines 621-629.  The bounds are just the declared bounds of the bounds-cast expression (the value of Bounds).

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8945,6 +8945,23 @@ def err_bounds_type_annotation_lost_checking : Error<
...
+  def err_bounds_cast_array_pointer_type_with_single_syntax
+      : Error<"invalid bounds casting : cast array pointer type with single "
+              "bounds casting syntax :  %0">;
+
+  def err_bounds_cast_integral_with_single_syntax
+      : Error<"invalid bounds casting : cast integral type with single bounds "
+              "casting syntax :  %0">;
+
+  def err_bounds_cast_single_pointer_type_with_array_syntax
+      : Error<"invalid bounds casting : cast single pointer type with array "
```
dtarditi: I would simplify this error message to be: `invalid bounds cast: expected Array_ptr type`.  The error message should point directly at the problem target type.

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8945,6 +8945,23 @@ def err_bounds_type_annotation_lost_checking : Error<
 
   def err_checked_scope_no_variadic_func_for_expression : Error<
     "cannot use a variable arguments function in a checked scope or function">;
+
+  def err_bounds_cast_array_pointer_type_with_single_syntax
```
dtarditi: I would simplify this error message to  `invalid bounds cast: expected _Ptr or * type`.  The error message should point directly at the problem target type.

lib/Sema/SemaBounds.cpp
```diff
@@ -909,6 +911,15 @@ namespace {
         }
       }
 
+      if (LHSType->isPointerType() &&
```
jijoongmoon: I do not know why this code is there. I thought I deleted it. I checked my local repository and there is no this code. I will make sure that there is no this code in next PR.

lib/Sema/SemaBounds.cpp
```diff
@@ -932,6 +943,19 @@ namespace {
         return true;
       }
 
+      if (CK == CK_PointerBounds &&
+          (dyn_cast<BoundsCastExpr>(E)->getBoundsCastKind() ==
+           BoundsCastExpr::Kind::Dynamic)) {
+        BoundsExpr *SrcBounds = S.InferRValueBounds(E);
+        if (SrcBounds->isNone()) {
```
jijoongmoon: I'm really sorry that this is mixed with previous code. I changed CK_PointerBounds to CK_DynamicPtrBounds and CK_AssumePtrBounds. I think it is not updated. I will Again.

lib/Sema/SemaBounds.cpp
jijoongmoon: Bounds and bounds are different as in this code. Bounds is the CastExpr's bounds expression and bounds is the bounds expression of e1. When the developer write code properly, the bounds expression of cast expression is generated during parsing ( GenerateBoundsExpr() in SemaExpr.cpp ). During parsing and generating bounds expression for cast expression, checking for the existence of e1 is not done. this checking is delayed until SemaBounds and this code is responsible for the check. Compare the Bounds for the Cast Expression and bounds for the e1. In the cast that there is bounds expression for cast expression and if there is no bounds for e1, I think this bounds expression for Cast Expression should be none.

lib/Sema/SemaBounds.cpp
```diff
@@ -907,12 +915,12 @@ namespace {
           else
             RHSBounds = S.InferRValueBounds(RHS);
           if (RHSBounds->isNone()) {
-             S.Diag(RHS->getLocStart(), diag::err_expected_bounds_for_assignment) << RHS->getSourceRange();
-             RHSBounds = S.CreateInvalidBoundsExpr();
+	    S.Diag(RHS->getLocStart(), diag::err_expected_bounds_for_assignment) << RHS->getSourceRange();
```
jijoongmoon: Sorry. I will.

lib/Sema/SemaBounds.cpp
```diff
@@ -907,12 +915,12 @@ namespace {
...
+	    RHSBounds = S.CreateInvalidBoundsExpr();
           }
         }
       }
-
+
```
jijoongmoon: Sorry. I will.

lib/Sema/SemaBounds.cpp
```diff
@@ -945,7 +974,7 @@ namespace {
         if (SrcBounds->isNone()) {
           S.Diag(E->getSubExpr()->getLocStart(), diag::err_expected_bounds_for_ptr_cast)  << E->getSubExpr()->getSourceRange();
           SrcBounds = S.CreateInvalidBoundsExpr();
-        }
+	}
```
jijoongmoon: I will. Thanks.

lib/Sema/SemaExpr.cpp
```diff
@@ -12490,36 +12493,77 @@ ExprResult Sema::ActOnBoundsCastExpr(Scope *S, SourceLocation LParenLoc,
...
+    if (DestTy->isCheckedPointerArrayType()) {
+      Diag(E1->getLocStart(),
+           diag::err_bounds_cast_array_pointer_type_with_single_syntax)
+          << DestTy;
+      Result = CreateInvalidBoundsExpr();
+    } else if (DestTy->isIntegralType(Context)) {
+      Diag(E1->getLocStart(), diag::err_bounds_cast_integral_with_single_syntax)
+          << DestTy;
+      Result = CreateInvalidBoundsExpr();
+    } else if (DestTy->isUncheckedPointerType()&&(Kind == tok::kw__Assume_bounds_cast)){
```
jijoongmoon: You mentioned at the last email, saying that the specification about this is error if the destination type is unchecked with single assume casting syntax. But I think have to follow the spec for now and if there is change, then change the code.

lib/Sema/SemaExpr.cpp
```diff
@@ -12490,36 +12493,77 @@ ExprResult Sema::ActOnBoundsCastExpr(Scope *S, SourceLocation LParenLoc,
...
+      Diag(E1->getLocStart(),
+           diag::
+               err_bounds_cast_unchecked_pointer_type_with_assume_single_syntax)
+          << DestTy;
+      Result = CreateInvalidBoundsExpr();
+    } else {
```
jijoongmoon: I see. I will check Ptr and * for the Destination type first and then fall into cases.

lib/Sema/SemaBounds.cpp
jijoongmoon: I will. Sorry.

lib/Sema/SemaBounds.cpp
```diff
@@ -936,6 +944,27 @@ namespace {
...
+        QualType DestTy = E->getType();
+        QualType SrcTy = subExpr->getType();
+
+        if (!Bounds->isInvalid() && bounds->isNone()) {
+          if (!(DestTy->isUncheckedPointerType() &&
+                SrcTy->isUncheckedPointerType()))
+            S.Diag(subExpr->getLocStart(), diag::err_expected_bounds);
+        }
+
+        if (SrcBounds->isNone()) {
```
jijoongmoon: I will check it and modify.

include/clang/Basic/DiagnosticSemaKinds.td
jijoongmoon: I will simplify this error message including others.

jijoongmoon:
I think it is ready for further review. Thanks.!

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8945,6 +8945,16 @@ def err_bounds_type_annotation_lost_checking : Error<
...
+      : Error<"invalid bounds cast : expected _Ptr or * type (%0)">;
+
+  def err_bounds_cast_integral_with_single_syntax
```
dtarditi: Could you remove the space before the `:`?    You can also delete the `(%0)` because the error message will point directly at the problematic type.   The error messages for integral and array_pointer types are now identical.  I think you can just have one error message.

lib/Sema/SemaBounds.cpp
```diff
@@ -936,6 +934,19 @@ namespace {
         return true;
       }
 
+      if (CK == CK_DynamicPtrBounds || CK == CK_AssumePtrBounds) {
+        BoundsExpr *SrcBounds = S.InferRValueBounds(E);
+        Expr *subExpr = E->getSubExpr();
+        BoundsExpr *subExprBounds = S.InferRValueBounds(subExpr);
+
+	if(subExprBounds->isNone()){
```
dtarditi: It looks like there needs to be more indentation here.

lib/Sema/SemaCast.cpp
```diff
@@ -2612,6 +2613,16 @@ void CastOperation::CheckCStyleCast() {
   }
 }
 
+void CastOperation::CheckBoundsCast(tok::TokenKind kind) {
+
+  SrcExpr = Self.DefaultFunctionArrayLvalueConversion(SrcExpr.get());
+  
+  if(kind == tok::kw__Assume_bounds_cast)
```
dtarditi: Please ads a space after the `if'.

lib/Sema/SemaExpr.cpp
```diff
@@ -12490,36 +12495,74 @@ ExprResult Sema::ActOnBoundsCastExpr(Scope *S, SourceLocation LParenLoc,
...
+      llvm::APInt I = llvm::APInt(1, 1, false);
+      uint64_t Bits = I.getZExtValue();
+      unsigned Width = Context.getIntWidth(Context.UnsignedLongLongTy);
+      llvm::APInt ResultVal(Width, Bits);
+      IntegerLiteral *One = IntegerLiteral::Create(
+          Context, ResultVal, Context.UnsignedLongLongTy, SourceLocation());
+      Result =
+          ActOnCountBoundsExpr(SourceLocation(), BoundsExpr::Kind::ElementCount,
+                               One, SourceLocation());
+    } else if (DestTy->isCheckedPointerArrayType()) {
```
dtarditi: If you merge the error messages, I think you can just replace this with an `else` clause that prints the error message.

lib/Sema/SemaExpr.cpp
```diff
@@ -12490,36 +12495,74 @@ ExprResult Sema::ActOnBoundsCastExpr(Scope *S, SourceLocation LParenLoc,
...
+      Diag(TypeLoc,
+           diag::err_bounds_cast_array_pointer_type_with_single_syntax)
+          << DestTy;
+      Result = CreateInvalidBoundsExpr();
+    } else if (DestTy->isIntegralType(Context)) {
+      Diag(TypeLoc, diag::err_bounds_cast_integral_with_single_syntax)
+          << DestTy;
+      Result = CreateInvalidBoundsExpr();
+    } else
```
dtarditi: It would be incorrect to fall into this case and not produce an error message.   The compiler would generate code in that case, even though the bounds would be invalid.

If you change the `else` clause that is above as I suggest, this is no longer an issue.

lib/Sema/SemaExpr.cpp
```diff
@@ -12490,36 +12495,74 @@ ExprResult Sema::ActOnBoundsCastExpr(Scope *S, SourceLocation LParenLoc,
...
+           diag::err_bounds_cast_single_pointer_type_with_array_syntax)
+          << DestTy;
+      Result = CreateInvalidBoundsExpr();
+    } else {
+      if (ResE2.isUsable() && ResE3.isUsable())
+        Result = ActOnRangeBoundsExpr(SourceLocation(), ResE2.get(),
+                                      ResE3.get(), SourceLocation());
+    }
+    break;
+  default:
```
dtarditi: The default should have an llvm_unreachable call.   It is not expected for the code to reach this point.

lib/Sema/SemaBounds.cpp
dtarditi: I wanted to follow up on your question about the code that I asked you to delete from RValueBounds.   

The code that was deleted was handling an error case.  The error case was that the subexpression did not have bounds.  

The RValueBounds function does not need to check for errors in bounds.  That checking is done elsewhere and error messages will be generated if there is a problem.   Because error messages are generated elsewhere, no code will be generated.  The fact that RValueBounds may produce bounds that are invalid for incorrect input does not matter.   The RValueBounds function only needs to not crash if it is given IR for which an error message was generated.   

We specifically do not want to make the bounds depend on whether the subexpression bounds are correct or not.   Using the bounds of None in the case of an error involving the subexpression could cause confusing error messages for programmers.   The compiler might complain that it expected the entire dynamic_bounds_cast or assume_bounds_cast expression to have bounds.   That is a confusing error message for a programmer to see when the error involves only the subexpression.   

For this case,  the code in the class CheckBoundsDeclarations does the check that the subexpression has bounds as expected.  It generates an error message, if it does not.  This means that compilation will not succeed and no code will be generated.

The purpose of the RValueBounds function is to infer bounds for an expression.   Dynamic_bounds_cast and assume_bound_cast expression affect inference of bounds differently than C-style cast expressions.  For C-style cast expression, the bounds just pass through.  For dynamic_bounds_cast/assume_bounds_cast, these declare new bounds for the expression.

The following example illustrates the difference in the operators.  In the first case, a C-style cast (which is unnecessary) preserves the bounds of `a`, so it is OK for `b` to have bounds 10.  In the second case, a `dynamic_bounds_cast` declares new bounds for `a` that are smaller than the original bounds.  It is not OK for `c` to have bounds 10 because the bounds have been changed to be smaller by the `dynamic_bounds_cast`
```
int a checked[10];
array_ptr<int> b: count(10) = (array_ptr<int>) a;
array_ptr<int> c : count(10) = dynamic_bounds_cast<array_ptr<int>>(a, 5); // error
```

include/clang/Basic/DiagnosticSemaKinds.td
jijoongmoon: I see. I will change it.

lib/Sema/SemaCast.cpp
jijoongmoon: I will fix it. Thanks.

lib/Sema/SemaExpr.cpp
jijoongmoon: I will merge this else clause.

lib/Sema/SemaExpr.cpp
jijoongmoon: I see.

lib/Sema/SemaExpr.cpp
jijoongmoon: I will change it. Thanks.

lib/Sema/SemaBounds.cpp
jijoongmoon: I'm not sure about it. Could you give me more specific? 
The syntax error is handled during generating bounds expression in SemaExpr and e1 expression has no bounds error is generated at this point.

jijoongmoon:
I think it is ready for further review. Thanks.

lib/Sema/SemaExpr.cpp
```diff
@@ -12490,36 +12495,69 @@ ExprResult Sema::ActOnBoundsCastExpr(Scope *S, SourceLocation LParenLoc,
...
+      llvm::APInt I = llvm::APInt(1, 1, false);
+      uint64_t Bits = I.getZExtValue();
+      unsigned Width = Context.getIntWidth(Context.UnsignedLongLongTy);
+      llvm::APInt ResultVal(Width, Bits);
+      IntegerLiteral *One = IntegerLiteral::Create(
+          Context, ResultVal, Context.UnsignedLongLongTy, SourceLocation());
+      Result =
+          ActOnCountBoundsExpr(SourceLocation(), BoundsExpr::Kind::ElementCount,
+                               One, SourceLocation());
+    } else if (DestTy->isCheckedPointerArrayType() ||
```
dtarditi: You should just fall into this else case.  There should be no check.    The problem is that if DestTy isn't a checked pointer array type or an integral type, the code just fall into the last case without an error.  As I explained in my prior review, for any code that creates an invalid bounds expression, there must always be an error message issued so that we don't generate code.

lib/Sema/SemaBounds.cpp
```diff
@@ -936,6 +934,20 @@ namespace {
         return true;
       }
 
+      if (CK == CK_DynamicPtrBounds || CK == CK_AssumePtrBounds) {
```
dtarditi: Sorry, I just noticed this.   We should only be requiring bounds for dynamic_bounds_casts.  We should not require bounds for assume ptr bounds.

assume_bound_cast is meant to be used for things like code that has to cast a hardware address (just an integer) to a checked pointer type:
```
// memory-mapped hardware buffer that is 128 bytes in size begins at 0x4000
array_ptr<char> buf : count(128) = assume_bounds_cast<array_ptr<char>>(0x4000, 128);
```

We will want to require that the subexpression for both dynamic_bound_cast and assume_bounds_cast opertors have a type suitable for casting to a pointer. It should have a pointer or an integeral type.  I will open a separate work item for that.   I don't want that check added to this pull request.  I checked the specification and noticed that it was silent on the type of the first operand.   I will file a work item for that too.

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8945,6 +8945,13 @@ def err_bounds_type_annotation_lost_checking : Error<
 
   def err_checked_scope_no_variadic_func_for_expression : Error<
     "cannot use a variable arguments function in a checked scope or function">;
+
+  def err_bounds_cast_error_with_single_syntax
+      : Error<"invalid bounds cast: expected _Ptr or * type">;
+
+  def err_bounds_cast_error_with_array_syntax
+      : Error<"invalid bounds cast: expected Array_ptr type">;
```
dtarditi: There's a type: the underscore in front of the type name is missing. It should `_Array_ptr`, not `Array_ptr`.

jijoongmoon:
I think it is ready for further review. Thanks always.

lib/Sema/SemaExpr.cpp
```diff
@@ -12490,36 +12497,66 @@ ExprResult Sema::ActOnBoundsCastExpr(Scope *S, SourceLocation LParenLoc,
...
+    break;
+  case BoundsCastExpr::SyntaxType::Count:
+    if (!DestTy->isCheckedPointerArrayType()) {
+      Diag(TypeLoc, diag::err_bounds_cast_error_with_array_syntax);
+      Result = CreateInvalidBoundsExpr();
+    } else if ((E1->getType())->isVoidPointerType()) {
+      Diag(E1->getLocStart(),
+             diag::err_typecheck_void_pointer_count_return_bounds);
+    } else {
+      if (ResE2.isUsable())
```
dtarditi: You should check that ResE2 is not invalid.  It is clang idiom that they do not have a method for check if a valid expression, you have to negate "IsInvalid" (a double negation).
```
if  (!ResE2.isInvalid()) { ...
```

lib/Sema/SemaExpr.cpp
```diff
@@ -12490,36 +12497,66 @@ ExprResult Sema::ActOnBoundsCastExpr(Scope *S, SourceLocation LParenLoc,
...
+                                      BoundsExpr::Kind::ElementCount,
+                                      ResE2.get(), SourceLocation());
+    }
+    break;
+  case BoundsCastExpr::SyntaxType::Range:
+    if (!DestTy->isCheckedPointerArrayType()) {
+      Diag(TypeLoc, diag::err_bounds_cast_error_with_array_syntax);
+      Result = CreateInvalidBoundsExpr();
+    } else {
+      if (ResE2.isUsable() && ResE3.isUsable())
```
dtarditi: You should use `!isInvalid()` here instead of `isUsable()`.

lib/Sema/SemaExpr.cpp
```diff
@@ -12490,36 +12497,66 @@ ExprResult Sema::ActOnBoundsCastExpr(Scope *S, SourceLocation LParenLoc,
...
+        Result = ActOnRangeBoundsExpr(SourceLocation(), ResE2.get(),
+                                      ResE3.get(), SourceLocation());
+    }
+    break;
+  default:
+    llvm_unreachable("unexpected expression kind");    
   }
+
```
dtarditi: You can delete lines 12558 and 12559.  I think we should just return an invalid expression here.  

As I mentioned, I think having an "invalid bounds expression" was a design mistake.   Clang follows the design choice that if an expression is invalid, any expression or declaration using it should be invalid too.

lib/Parse/ParseExpr.cpp
```diff
@@ -3064,34 +3066,38 @@ ExprResult Parser::ParseBoundsCastExpression(IdentifierInfo &Ident,
             ParseRelativeBoundsClause(isError, Ident, Tok.getLocation());
         if (isError)
           SkipUntil(tok::r_paren, StopAtSemi | StopBeforeMatch);
-	PT.consumeClose();
+        PT.consumeClose();
```
jijoongmoon: I see. I will change to return ExprError() if there is an error.

lib/Parse/ParseExpr.cpp
```diff
@@ -3064,34 +3066,38 @@ ExprResult Parser::ParseBoundsCastExpression(IdentifierInfo &Ident,
...
+
   LParenLoc = T.getOpenLocation();
 
   // Parsing e1, e2, e3
   ExprResult E1(true), E2(true), E3(true);
 
-  E1 = ParseCastExpression(true);
-  if (E1.isInvalid()) {
+  E1 = ParseAssignmentExpression();
+  if (!E1.isInvalid())
```
jijoongmoon: I will change code. Thanks.

lib/Parse/ParseExpr.cpp
```diff
@@ -3064,34 +3066,38 @@ ExprResult Parser::ParseBoundsCastExpression(IdentifierInfo &Ident,
...
+    E1 = Actions.CorrectDelayedTyposInExpr(E1.get());
+  else {
     SkipUntil(tok::r_paren, StopAtSemi | StopBeforeMatch);
     return ExprError();
   }
 
   if (Tok.is(tok::comma)) {
     ConsumeToken();
+    syntax = BoundsCastExpr::SyntaxType::Count;
```
jijoongmoon: I will make sure to stop parsing and try skip tokens and return ExprError() if there is an error about E1, E2 and E3.

lib/Parse/ParseExpr.cpp
```diff
@@ -3103,6 +3109,7 @@ ExprResult Parser::ParseBoundsCastExpression(IdentifierInfo &Ident,
 
   if (Tok.is(tok::comma)) {
     ConsumeToken();
+    syntax = BoundsCastExpr::SyntaxType::Range;
```
jijoongmoon: I will make sure. Thanks.

lib/Parse/ParseExpr.cpp
```diff
@@ -3115,10 +3122,12 @@ ExprResult Parser::ParseBoundsCastExpression(IdentifierInfo &Ident,
...
-  if (!E1.isInvalid())
-    Result = Actions.ActOnBoundsCastExpr(getCurScope(), ILoc, Ty.get(),
-                                         RelativeClause, RParenLoc, E1.get(),
-                                         E2.get(), E3.get(), kind);
+
+  Result = Actions.ActOnBoundsCastExpr(
+      getCurScope(), OpLoc, Kind, LAngleBracketLoc, Ty.get(), RAngleBracketLoc,
+      RelativeClause, LParenLoc, RParenLoc, E1.get(), E2.get(), E3.get(),
+      syntax);
+
```
jijoongmoon: I think if there is an error about E1, E2, E3 and TypeResult, then the code return ExprError(). code is not reached at this point if one of them is invalid. If want me to check them again, I will modify the code to check again here.

lib/Sema/SemaExpr.cpp
```diff
@@ -12465,23 +12466,29 @@ ExprResult Sema::CreatePositionalParameterExpr(unsigned Index, QualType QT) {
...
+ExprResult Sema::ActOnBoundsCastExpr(
+    Scope *S, SourceLocation OpLoc, tok::TokenKind Kind,
+    SourceLocation LAngleBracketLoc, ParsedType D,
+    SourceLocation RAngleBracketLoc, RelativeBoundsClause *RelativeClause,
+    SourceLocation LParenLoc, SourceLocation RParenLoc, Expr *E1, Expr *E2,
+    Expr *E3, BoundsCastExpr::SyntaxType SyntaxType) {
   RangeBoundsExpr *Range = nullptr;
   TypeSourceInfo *castTInfo;
-  GetTypeFromParser(D, &castTInfo);
+  QualType DestTy = GetTypeFromParser(D, &castTInfo);
```
jijoongmoon: yes. I should remove this code. I was needed previously but it's not now. Thanks.

lib/Sema/SemaExpr.cpp
```diff
@@ -12465,23 +12466,29 @@ ExprResult Sema::CreatePositionalParameterExpr(unsigned Index, QualType QT) {
...
     return ExprError();
   E1 = ResE1.get();
 
-  ExprResult bounds = GenerateBoundsExpr(E2, E3);
-  if (RelativeClause != nullptr) {
+  ExprResult bounds =
+      GenerateBoundsExpr(E1, E2, E3, DestTy, SyntaxType, Kind,
+                         (castTInfo->getTypeLoc()).getBeginLoc());
+
+  if(bounds.isInvalid())
```
jijoongmoon: Yes. I will add space.

lib/Sema/SemaExpr.cpp
```diff
@@ -12490,36 +12497,66 @@ ExprResult Sema::ActOnBoundsCastExpr(Scope *S, SourceLocation LParenLoc,
...
+          ActOnCountBoundsExpr(SourceLocation(), BoundsExpr::Kind::ElementCount,
+                               One, SourceLocation());
+    } else {
+      Diag(TypeLoc, diag::err_bounds_cast_error_with_single_syntax);
+    }
+    break;
+  case BoundsCastExpr::SyntaxType::Count:
+    if (!DestTy->isCheckedPointerArrayType()) {
+      Diag(TypeLoc, diag::err_bounds_cast_error_with_array_syntax);
+      Result = CreateInvalidBoundsExpr();
```
jijoongmoon: I will delete it.

lib/Sema/SemaExpr.cpp
jijoongmoon: I thought E2, E3 and E1 are not invalid after they comes here. I will add invalid check.

lib/Sema/SemaExpr.cpp
```diff
@@ -12490,36 +12497,66 @@ ExprResult Sema::ActOnBoundsCastExpr(Scope *S, SourceLocation LParenLoc,
...
+      if (ResE2.isUsable())
+        Result = ActOnCountBoundsExpr(SourceLocation(),
+                                      BoundsExpr::Kind::ElementCount,
+                                      ResE2.get(), SourceLocation());
+    }
+    break;
+  case BoundsCastExpr::SyntaxType::Range:
+    if (!DestTy->isCheckedPointerArrayType()) {
+      Diag(TypeLoc, diag::err_bounds_cast_error_with_array_syntax);
+      Result = CreateInvalidBoundsExpr();
```
jijoongmoon: Yes. I will delete it.

lib/Sema/SemaExpr.cpp
jijoongmoon: I will change them.

lib/Sema/SemaExpr.cpp
jijoongmoon: I will delete them.

jijoongmoon:
I think it is ready for further review. Thanks always.

lib/AST/ASTDumper.cpp
```diff
@@ -1965,23 +1965,13 @@ void ASTDumper::VisitCastExpr(const CastExpr *Node) {
...
-    OS << ">";
-  }
-
-  if (const BoundsExpr *Bounds = Node->getBoundsExpr()) {
-    dumpChild([=] {
-      OS << "Inferred Bounds";
-      dumpStmt(Bounds);
-    });
-  }
+  if(Node->getStmtClass() != Expr::BoundsCastExprClass)
```
dtarditi: There needs to be a space after the `if'.

lib/Sema/SemaBounds.cpp
```diff
@@ -936,6 +934,20 @@ namespace {
         return true;
       }
 
+      if (CK == CK_DynamicPtrBounds) {
+        BoundsExpr *SrcBounds = S.InferRValueBounds(E);
```
dtarditi: I am confused by this code.   Why are you inferring the rvalue bounds for the entire cast expression?   We also should not be modifying the bounds expression with E->setBoundsExpr().  

I think it is enough here to just check that the subexpression has bounds and issue an error message if it does not.

lib/AST/ASTDumper.cpp
jijoongmoon: I will make space after the if. Thanks.

lib/Sema/SemaBounds.cpp
jijoongmoon: Yes. You are right. I will change it.

PR #271, opened 2017-04-13
Support for automated LNT testing using VSTS.

dtarditi:
- Initial support; I'll add the LNT cmake support later after this is working reliably.
- Follows the same pattern as used for other repos: you can control the branch and commit in the repo to use for testing.
- Assumes that the LNT system has been pre-installed on the agent in a known location.   My conclusion was that is too messy to be installing LNT as part of the automated test, especially given the packages required by LNT.   It'd be fine to have a separate script for LNT.


automation/UNIX/test-lnt.sh
```diff
@@ -0,0 +1,36 @@
+# Test clang on UNIX using Visual Studio Team Services
+
+set -ue
+set -o pipefail
+set -x
```
lenary: You probably want to remove this line unless you want bash to output all its debug information.

automation/UNIX/test-lnt.sh
```diff
@@ -0,0 +1,36 @@
...
+CLANG=${LLVM_OBJ_DIR}/bin/clang
+RESULT_LOG="${LNT_RESULTS_DIR}/result.log"
+
+
+if [ ! -e "$CLANG" ]; then
+  echo "clang compiler not found at $CLANG"
+  exit 1
+fi
+
+"$LNT_SCRIPT" runtest nt --sandbox "$LNT_RESULTS_DIR" --no-timestamp \
```
lenary: I'd abandon the `--no-timestamp` argument so that it versions each benchmark build folder. I don't think want prior build artefacts skewing test results. If you want caching, there is definitely a way of making that work with cmake, for the `nt` suite I'm not so sure.

dtarditi: Thanks - the script deletes the results directory each time it runs, so there will be interaction with prior runs.  

I didn't want to use a timestamped directory name because it means that we the script does not know the precise directory name.  That in turn would make it harder find files later to upload for retention.  IN VSTS, agents cache files, but are not used for long-term retention of files.

automation/UNIX/test-lnt.sh
dtarditi: It's helpful to have the detailed trace of commands when there are automation failures.   It does make the logs harder to read because they are more verbose, but I don't think it is excessive in this case (the results are at the end of the log file).

automation/UNIX/test-lnt.sh
dtarditi: I just corrected some typos in the prior comment.

Issue #273, opened 2017-04-13
checkedc nme_bounds.c compiler test failing


dtarditi:
The checked nme_bounds.c test (in the Checked C repo in tests/static_checking/nme_bounds.c) is failing using a Release X64 Linux build of the checked clang compiler.   The failure from the log file is:
```
2017-04-13T18:11:56.5126530Z error: 'error' diagnostics expected but not seen: 
2017-04-13T18:11:56.5157620Z   File /mnt/work/3/s/llvm/projects/checkedc-wrapper/checkedc/tests/static_checking/nme_bounds.c Line 48: assignment expression not allowed in bounds expression
2017-04-13T18:11:56.5185130Z 1 error generated.
```

PR #274, opened 2017-04-13
Fix uninitialized field in non-modifying expression analysis.

dtarditi:
The field ReportError was not initialized by the constructor in NonModifyingExprSema.
This caused the Checked C test nme_bounds.c to fail when compiled using a Release X64
Linux version of the Checked C clang compiler (checked-clang Github issue #273).

Testing:
- Checked C tests now pass using the compiler configuration that was failing.


dtarditi:
I am merging now because the fix is trivial and this is blocking my testing of LNT.   Please feel free to review this when you have time and point out anything that I missed.

lenary:
LGTM, :+1:, sorry I missed this last time.

Issue #282, opened 2017-04-19
Return value of ParseRelativeBoundsForDecl not being used


dtarditi:
I noticed that the return value of ParseRelativeBoundsForDecl seems to never be used.   At the call in ParseBoundsExpressionOrInteropType, I think we should check the return value and make sure that the function returns an invalid bounds expression if there was an error.   This will prevent spurious downstream errors later on.

I don't think we need to do anything at the call in ParseStructDeclaration.  It would be helpful to add a comment there explaining what is going on.  We expect this call to always generate an error message.  Because of that, we don't need to check the return value.

PR #285, opened 2017-04-25
Fix Return value of ParseRelativeBoundsForDecl not being used (#282)

jijoongmoon:
.According to the return value of ParseRelativeBoundsForDecl, it generates "invalid rel_align or rel_align_value" error message.

### DIFF
--- a/lib/Parse/ParseExpr.cpp
+++ b/lib/Parse/ParseExpr.cpp
@@ -2841,7 +2841,10 @@ ExprResult Parser::ParseBoundsExpressionOrInteropType(const Declarator &D,
 
   TempTok = Tok;
   if (StartsRelativeBoundsClause(Tok))
-    ParseRelativeBoundsClauseForDecl(Result);
+    if (ParseRelativeBoundsClauseForDecl(Result)) {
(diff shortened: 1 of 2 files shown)


jijoongmoon:
I think it is ready for review. Thanks.

lib/Parse/ParseExpr.cpp
```diff
@@ -2841,7 +2841,10 @@ ExprResult Parser::ParseBoundsExpressionOrInteropType(const Declarator &D,
 
   TempTok = Tok;
   if (StartsRelativeBoundsClause(Tok))
-    ParseRelativeBoundsClauseForDecl(Result);
+    if (ParseRelativeBoundsClauseForDecl(Result)) {
+      Diag(Tok, diag::err_invalid_relative_bounds_clause);
```
dtarditi: You don't need to issue an error message here.  I believe ParseRelativeBoundsClauseForDecl should have issued an error message if necessary.  This will just confuse the programmer who sees the message.  It is sufficient to just return an ExprError().

jijoongmoon:
I think it is ready for further review. Thanks always!

Issue #277, opened 2017-04-14
Do not allow some bounds_cast operations in checked scope


dtarditi:
Once the work on _Assume_bounds_cast is finished, we need to make sure it cannot be used in a checked scope.

wonsubkim:
Prevent _Assume_bounds_cast ?
Prevent _Assume_bounds_cast / _Dynamic_bounds_cast to unchecked type?

dtarditi:
We want to prevent all _Assume_bounds_cast in a checked scope.   They could be wrong even for checked pointer to checked pointer casts.

You are correct that we should also prevent _Dynamic_bounds_cast to unchecked types.

wonsubkim:
As soon as bounds_cast issue finished, I will apply this issue

dtarditi:
This work is complete.  Thank you!

PR #288, opened 2017-04-28
Do not allow some bounds_cast in checked scope (#277)

wonsubkim:
+ Do not allow some bounds_cast in checked scope (#277)
  + prevent all _Assume_bounds_cast operations in checked scope
  + prevent _Dynamic_bounds_cast to unchecked or variadic type in a checked scope
  it is same as constraints of C-style explicit casting in a checked scope
+ bounds cast operation
  + if source bounds is bounds(any), no runtime checks are neeeded

### DIFF
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -8946,6 +8946,9 @@ def err_bounds_type_annotation_lost_checking : Error<
   def err_checked_scope_no_variadic_func_for_expression : Error<
     "cannot use a variable arguments function in a checked scope or function">;
 
+  def err_checked_scope_no_assume_bounds_casting : Error<
+    "invalid _Assume_bounds_cast in a checked scope or function">;
+
   def err_bounds_cast_error_with_single_syntax
       : Error<"invalid bounds cast: expected _Ptr or * type">;
 
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -8947,7 +8947,7 @@ def err_bounds_type_annotation_lost_checking : Error<
     "cannot use a variable arguments function in a checked scope or function">;
 
   def err_checked_scope_no_assume_bounds_casting : Error<
-    "invalid _Assume_bounds_cast in a checked scope or function">;
+    "_Assume_bounds_cast not allowed in a checked scope or function">;
 
   def err_bounds_cast_error_with_single_syntax
       : Error<"invalid bounds cast: expected _Ptr or * type">;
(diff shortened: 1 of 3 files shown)


include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8946,6 +8946,9 @@ def err_bounds_type_annotation_lost_checking : Error<
   def err_checked_scope_no_variadic_func_for_expression : Error<
     "cannot use a variable arguments function in a checked scope or function">;
 
+  def err_checked_scope_no_assume_bounds_casting : Error<
+    "invalid _Assume_bounds_cast in a checked scope or function">;
```
dtarditi: I think it would be better to say "_Assume_bounds_cast not allowed in a checked scope or function".    Otherwise programmers may think there is something wrong with the _Assume_bounds_cast expression itself.

lib/Sema/SemaBounds.cpp
```diff
@@ -934,6 +934,10 @@ namespace {
         return true;
       }
 
+      // If inferred bounds of e1 are bounds(none), compile-time error.
+      // If inferred bounds of e1 are bounds(any), no runtime checks.
+      // Otherwise, the inferred bounds is bounds(lb, ub).
```
dtarditi: Your reasoning about bounds(any) is correct, but it belongs in the phase that generates dynamic checks for _Dynamic_boundsd_cast, not here. I think you should remove this change from this

lib/Sema/SemaCast.cpp
```diff
@@ -2621,6 +2621,30 @@ void CastOperation::CheckBoundsCast(tok::TokenKind kind) {
     Kind = CK_AssumePtrBounds;
   else if (kind == tok::kw__Dynamic_bounds_cast)
     Kind = CK_DynamicPtrBounds;
+
+  // Checked C - No C-style casts to unchecked pointer/array type or variadic
+  // type in a checked block.
```
dtarditi: This looks good!  Thanks.

lib/Sema/SemaBounds.cpp
wonsubkim: Okay, I will move this change into insertion of runtime checking
Thanks.

wonsubkim:
Can you let me know typo to fix?

lib/Sema/SemaBounds.cpp
```diff
@@ -934,6 +934,11 @@ namespace {
...
+      // bounds of cast operation is bounds(e2, e3).
+      // In code geneartion, it inserts dynamic_check(lb <= e2 && e3 <= ub).
```
dtarditi: There is misspelling in this comment: the word generation is misspelled.

Issue #287, opened 2017-04-27
Follow-up on issues from pull request #269.


dtarditi:
There are some issues for @jijoongmoon  to follow up on from pull request #269 and the corresponding testing changes in the Checked C repo (https://github.com/Microsoft/checkedc/pull/147).     Opening a Github issue to make sure this work is tracked.

dtarditi:
This work is complete.

PR #289, opened 2017-04-28
Follow-up on issues from pull request #269 (#287)

jijoongmoon:
Delete inferring the rvalue bounds for the entire cast expression in SemaBounds.


jijoongmoon:
I think it is ready for review. Thanks.

Issue #286, opened 2017-04-25
Check types of operands of *_bounds_cast operators


dtarditi:
This should be done separately from PR #269.  The first operand must be suitable for casting to a pointer type: it must be a pointer type or an integral type.    If there is only a second operand, it must have a type suitable for a count expression.  If there are second and third operands, they must have types suitable for a bounds expression.

We might want to change the syntax of dynamic_bounds_cast and assume_bounds_cast to take an actual bounds expression as the second argument:
```
array_ptr<int> x : count(10) = dynamic_bounds_cast<array_ptr<int>>(x, count(10));
array_ptr<int> x : bounds(a, b) = dynamic_bounds_cast<array_ptr<int>>(x, bounds(a, b));
```

jijoongmoon:
I think it is only needed to check first operand if it is pointer type or integral type. The check about second operand and third operand, the type checking are done in ActOnCountBoundsExpr and ActOnRangeBoundsExpr. In ActOnCountBoundsExpr, isIntegerType() is done for the E2 and ValidateBoundsExprArgument(E2) and ValidateBoundsExprArgument(E3) are done in ActOnRangeBoundsExpr. If we are doing these check during the parsing bounds cast expression, I think it is duplicated. Am I right?

dtarditi:
I agree that we only need to check the first operand.

Yes, it is OK to use the relative clause for the bounds as below.

dtarditi:
This work is complete - thanks Jijoong!

PR #290, opened 2017-05-02
Check types of operands of *_bounds_cast operators (#286)

jijoongmoon:
. Add check whether E1 is a pointer type or and integral type. If E1 is
   functional type, it checks the return type of E1.

 . Add parsing bounds expression for bounds cast expression.
   - Add check whether the second argument of bounds cast expression is bounds
     expression such as count(10) or bounds(a, b) and it parses them except
     byte_count expression. If it is byte_count then it return ExprError().

### DIFF
--- a/lib/Parse/ParseExpr.cpp
+++ b/lib/Parse/ParseExpr.cpp
@@ -3087,7 +3087,7 @@ ExprResult Parser::ParseBoundsCastExpression() {
   LParenLoc = T.getOpenLocation();
 
   // Parsing e1, e2, e3
-  ExprResult E1(true), E2(true), E3(true);
+  ExprResult E1(true), E2(true), E3(true), ParsedBounds(true);
 
   E1 = Actions.CorrectDelayedTyposInExpr(ParseAssignmentExpression());
   if (E1.isInvalid()) {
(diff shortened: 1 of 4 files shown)


jijoongmoon:
I think it is ready to review. Thanks!

lib/Sema/SemaExpr.cpp
```diff
@@ -12466,6 +12466,63 @@ ExprResult Sema::CreatePositionalParameterExpr(unsigned Index, QualType QT) {
   return new (Context) PositionalParameterExpr(Index, QT);
 }
 
+bool Sema::CheckBoundsCastBaseType(Expr *E1) {
+  bool Result = false;
+  QualType SrcTy = E1->getType();
+  if (SrcTy->isFunctionType())
```
dtarditi: I don't understand why we are checking the return type of E1 if E1 is a function type.    I think we  do not want to allow assume_bounds_cast or  dynamic_bounds cast to or from function types.  I think this line and the following line can be deleted.

lib/Sema/SemaExpr.cpp
```diff
@@ -12526,6 +12586,9 @@ ExprResult Sema::ActOnBoundsCastExprCount(
   ExprResult ResE2 = CorrectDelayedTyposInExpr(E2);
   SourceLocation TypeLoc = (castTInfo->getTypeLoc()).getBeginLoc();
 
+  if(CheckBoundsCastBaseType(E1))
```
dtarditi: There needs to be a space after the `if '

lib/Parse/ParseExpr.cpp
```diff
@@ -3097,11 +3097,23 @@ ExprResult Parser::ParseBoundsCastExpression() {
...
     ConsumeToken();
-    syntax = BoundsCastExpr::SyntaxType::Count;
-    E2 = Actions.CorrectDelayedTyposInExpr(ParseAssignmentExpression());
-    if (E2.isInvalid()) {
-      SkipUntil(tok::r_paren, StopAtSemi | StopBeforeMatch);
-      return ExprError();
+    if (StartsBoundsExpression(Tok)) {
+      IdentifierInfo *Ident = Tok.getIdentifierInfo();
+
+      if (Ident == Ident_byte_count)
```
jijoongmoon: I see. I will delete it.

lib/Sema/SemaExpr.cpp
jijoongmoon: Yes. I will delete it.

lib/Sema/SemaExpr.cpp
```diff
@@ -12479,6 +12536,9 @@ ExprResult Sema::ActOnBoundsCastExprSingle(
   QualType DestTy = GetTypeFromParser(D, &castTInfo);
   SourceLocation TypeLoc = (castTInfo->getTypeLoc()).getBeginLoc();
 
+  if(CheckBoundsCastBaseType(E1))
```
jijoongmoon: I will put space after the 'if'

lib/Sema/SemaExpr.cpp
jijoongmoon: I will put space after the 'if'

jijoongmoon:
I think it is ready for further review. Thanks!

dtarditi:
Looks good!

PR #292, opened 2017-05-09
Allow re-writer to work with allocator patterns

awruef:
This implements a very simple heuristic for allowing the converter to constrain allocations to `_Ptr<T>`. It also allows for a limited form of cast placement in the event of a conflict during unification.

### DIFF
--- a/test/CheckedCRewriter/allocator.c
+++ b/test/CheckedCRewriter/allocator.c
@@ -1,0 +1,22 @@
+// Tests for Checked C rewriter tool.
+//
+// Tests for malloc and friends. 
+//
+// RUN: checked-c-convert %s -- | FileCheck -match-full-lines %s
+// RUN: checked-c-convert %s -- | %clang_cc1 -verify -fcheckedc-extension -x c -
+// expected-no-diagnostics
+//
+#include <stdlib.h>
+
+void dosomething(void) {
+  int a = 0;
+  int *b = &a;
+  *b = 1;
+  return;
+}
+
+void foo(void) {
+  int *a = (int *) malloc(sizeof(int));
+  *a = 0;
+  return;
+}
--- a/test/CheckedCRewriter/allocator.c
+++ b/test/CheckedCRewriter/allocator.c
@@ -20,11 +20,13 @@ void dosomething(void) {
 void foo(void) {
   int *a = (int *) malloc(sizeof(int));
   *a = 0;
-  free((void*)a);
+  free(a);
   return;
 }
 //CHECK: void foo(void) {
 //CHECK-NEXT: _Ptr<int>  a = (int *) malloc(sizeof(int));
+//CHECK-NEXT: *a = 0;
+//CHECK-NEXT: free((void *)a);
 
 typedef struct _listelt {
   struct _listelt *next;
--- a/test/CheckedCRewriter/allocator.c
+++ b/test/CheckedCRewriter/allocator.c
@@ -6,7 +6,7 @@
 // RUN: checked-c-convert %s -- | %clang_cc1 -verify -fcheckedc-extension -x c -
 // expected-no-diagnostics
 //
-#include <stdlib.h>
+extern void *malloc(unsigned long);
 
 void dosomething(void) {
   int a = 0;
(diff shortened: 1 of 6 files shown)


test/CheckedCRewriter/allocator.c
```diff
@@ -0,0 +1,50 @@
...
+typedef struct _listelt {
+  struct _listelt *next;
+  int val;
+} listelt;
+
+typedef struct _listhead {
+  listelt *hd;
+} listhead;
+
+void add_some_stuff(listhead *hd) {
```
dtarditi: It would be useful to add a comment to this function explaining why we don't expect the malloc below to be converted.

tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -220,13 +221,54 @@ class FunctionVisitor : public RecursiveASTVisitor<FunctionVisitor> {
...
+            // the type passed to malloc is equal to both lhsType and rhsTy. 
+            // If it is, we can do something less conservative. 
+            if (CallExpr *CA = dyn_cast<CallExpr>(C->getSubExpr())) {
+              // Is this a call to malloc? Can we coerce the callee 
+              // to a NamedDecl?
+              FunctionDecl *calleeDecl = 
+                dyn_cast<FunctionDecl>(CA->getCalleeDecl());
+              if (calleeDecl && calleeDecl->getName() == "malloc") {
+                // It's a call to malloc. What about the parameter to the call?
+                if (CA->getNumArgs() > 0) {
```
dtarditi: I think you need to check the operator is a sizeof operator.

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -414,13 +414,96 @@ void emit(Rewriter &R, ASTContext &C, std::set<FileID> &Files,
...
+class CastPlacementVisitor : public RecursiveASTVisitor<CastPlacementVisitor> {
+public:
+  explicit CastPlacementVisitor(ASTContext *C, ProgramInfo &I, 
+      Rewriter &R, std::set<FileID> &Files)
+    : Context(C), Info(I), R(R), Files(Files) {} 
+
+  bool VisitCallExpr(CallExpr *);
+private:
+  std::set<unsigned int> getParamsForExtern(std::string);
+  bool anyTop(std::set<ConstraintVariable*>);
```
dtarditi: `anyTop` could use a comment explaining what it does.

test/CheckedCRewriter/allocator.c
awruef: Actually, I was just missing the CHECK line telling it to look for a cast. I'll add that.

tools/checked-c-convert/ConstraintBuilder.cpp
awruef: Yes, I do. I'll figure out how to do that and do that.

tools/checked-c-convert/CheckedCConvert.cpp
awruef: I'll add one.

test/CheckedCRewriter/allocator.c
lenary: Is there any aim here to also rewrite the explicit cast type?

awruef: We could do that, too. It wouldn't be that much extra work, I think. I'm not sure which is better, and not sure how to develop an opinion on which is better. What do you think?

lenary: Well If my understanding of `checked` blocks is correct, `ptr<T> var = (T*)malloc(sizeof(T));` would be disallowed in a checked block, but `ptr<T> var = (ptr<T>)malloc(sizeof(T));` would be allowed, suggesting if we want programmers to do as few changes as possible, then we should do the latter.

(Eventually) It doesn't seem beyond the bounds of possibility to have the checked c convert tool add `checked` keywords to functions that only use checked pointers (ie, functions that only use `T*` like `ptr<T>`, as that's the only analysis you're currently doing), and so preempting this doesn't seem like a bad idea, if it's a quick change.

awruef: Yeah, that makes sense to me. If it takes me more than half an hour to figure out tomorrow, I'll make a separate work item for it and do it later.

awruef: I'm going to punt on this, I made #294 to track it.

awruef:
I think this is ready for another look.

PR #296, opened 2017-05-12
Type-checking error leads to compiler crash

### LINKED ISSUE #291: Type-checking error leads to compiler crash.
The following snippet of incorrect code causes the debug build of the compiler to assert:
```
struct S {
  int arr _Checked[10];
  int data;
};

void passing_test_2(int i) {
  struct S s = { { 9, 8, 7, 6, 5, 4, 3, 2, 1, 0 }, 10 };
  s->arr[i] = 1;
}
```

Here is the output and the line of the assert.   The compiler is crashing during the checking of bounds declarations.
```
D:\test>%clang% -fcheckedc-extension tmp.c
tmp.c:8:4: error: member reference type 'struct S' is not a pointer; did you mean to use '.'?
  s->arr[i] = 1;
  ~^~
   .
Assertion failed: E->isLValue(), file D:\llvm\tools\clang\lib\Sema\SemaBounds.cpp, line 402
```
The assert is double-checking that an expression is an lvalue as expected.

jp1729:
Type-checking error removes lvalue attribute in MemberExpr. This causes assert to throw error. Changed assert to conditional. Although there are numerous lvalue assert checks, it looks like the assert in Line402 is the only assert that it comes across. Fixes #291

### DIFF
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -399,7 +399,11 @@ namespace {
     // should be the range of an object in memory or a subrange of
     // an object.
     BoundsExpr *LValueBounds(Expr *E) {
-      assert(E->isLValue());
+			// When type-checking errors, the resulting AST may not have 
+			// lvalue attribute to a node when it should. Then, if we use 
+			// assert, the compiler crashes. This should be okay, because 
+			// this will only occur when compiler has an error.
+			if (!E->isLValue()) return CreateBoundsNone();
       // TODO: handle side effects within E
       E = E->IgnoreParens();
       switch (E->getStmtClass()) {
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -924,8 +924,12 @@ namespace {
         DumpAssignmentBounds(llvm::outs(), E, LHSTargetBounds, RHSBounds);
       return true;
     }
-
-    // This includes both ImplicitCastExprs and CStyleCastExprs
+	    
+    // If inferred bounds of e1 are bounds(none), compile-time error. 
+    // If inferred bounds of e1 are bounds(any), no runtime checks. 
+    // Otherwise, the inferred bounds is bounds(lb, ub). 
+    // bounds of cast operation is bounds(e2, e3). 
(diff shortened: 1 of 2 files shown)


jp1729:
They are ready for review again. I have added comments and a regression test file in clang/test/CheckedC directory.

lib/Sema/SemaBounds.cpp
```diff
@@ -399,7 +399,11 @@ namespace {
     // should be the range of an object in memory or a subrange of
     // an object.
     BoundsExpr *LValueBounds(Expr *E) {
-      assert(E->isLValue());
+			// When type-checking errors, the resulting AST may not have
```
dtarditi: The indenting looks wrong.  If you are using Visual Studio, please be sure to set Visual Studio to insert spaces instead of tabs: Tools Options->Text Editor->All Languages->Tabs.

I think you can shorten your comment.   You don't really need to talk about assert.  It is more interesting to explain why this code is there.

test/CheckedC/type-checking-crash.c
```diff
@@ -0,0 +1,35 @@
+// Checks for crashes during type-checking error.
+// This was caused by an lvalue assert in SemaBounds.cpp
+// The goal of this test is to make sure we're seeing 
+// type-checking errors, but not get any crashes.
+// The llvm-lit test will fail the test if the compiler
+// crashes because of assert failure (This has been tested)
+// I wanted to use FileCheck to specifically look for
```
dtarditi: The test looks good.    This comment needs to be cleaned up. I generally like tests to have comments with the following form:

- 1-2 sentence summary of what the test does.
- Blank comment line.
- More detailed explanation.

Comments should be written with the idea that someone you don't know will look at it 2 years from now.  The first sentence needs to be more precise.   I'd suggest removing the comment about trying to use FileCheck.  As you mention, llvm-lit will fail a test if a compiler crashes because of an assert in a compiler, so you can't use FileCheck to check if an assertion was hit.

It seems a little strange to me that we get a type checking error after we get a bounds checking error.   I think it would be clearer to users if the type checking message happened first (generally, you'd want to fix typechecking errors first).

jp1729:
It is ready for review again. Indenting and comments are fixed.

PR #300, opened 2017-05-15
Handle arrays of _Ptr and function pointers

awruef:
The re-writer can now deal with some cases where there are arrays of pointers, including arrays of function pointers. This required a few small changes:

 * Type traversal needed to become array-aware. 
 * Type traversal also needed to become aware of clang "decayed" type nodes.
 * `PVConstraint`s now contain another constraint variable for the referent type.


tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -26,53 +26,94 @@ tyToStr(const Type *T) {
...
-		// Iterate.
-		QTy = QTy.getSingleStepDesugaredType(C);
-		QTy = QTy.getTypePtr()->getPointeeType();
-		Ty = QTy.getTypePtr();
+
+      K++;
+
+      // Iterate.
+      // TODO: Why is it the case that we can only get the unqualified type
+      //       of an array element? That's okay, right?
```
dtarditi: No, this would cause us to drop const on array elements.  You should call getElementType() instead on the base class of ArrayType.  You need to cast the qualified type to be an ArrayType first.

tools/checked-c-convert/ConstraintBuilder.cpp
```diff
@@ -575,7 +584,7 @@ class GlobalVisitor : public RecursiveASTVisitor<GlobalVisitor> {
           // that don't have any pointers.
 
           for (const auto &D : Definition->fields())
-            if (D->getType()->isPointerType()) {
+            if (D->getType()->isPointerType() || D->getType()->isArrayType()) {
```
awruef: Will fix.

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -108,20 +111,38 @@ class PointerVariableConstraint : public ConstraintVariable {
...
+    O_SizedArray,
+    O_UnSizedArray
+  };
+  // Map from constraint variable to original type and size. 
+  // If the original variable U was:
+  //  * A pointer, then U -> (a,b) , a = O_Pointer, b has no meaning.
+  //  * A sized array, then U -> (a,b) , a = O_SizedArray, b is static size.
+  //  * An unsized array, then U -(a,b) , a = O_UnSizedArray, b has no meaning.
+  std::map<uint32_t,std::pair<OriginalArrType,uint64_t>> arrSizes;
+  // If for all U in arrSized, any U -> (a,b) where a = O_SizedArray or
```
awruef: Will fix.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -26,53 +26,94 @@ tyToStr(const Type *T) {
...
+  PointerVariableConstraint(D->getType(), K, D, D->getName(), CS, C) { }
 
 PointerVariableConstraint::PointerVariableConstraint(const QualType &QT, uint32_t &K,
-	std::string N, Constraints &CS, const ASTContext &C) : 
+	DeclaratorDecl *D, std::string N, Constraints &CS, const ASTContext &C) : 
 	ConstraintVariable(ConstraintVariable::PointerVariable, 
-					   tyToStr(QT.getTypePtr())),FV(nullptr)
-{ 
+					   tyToStr(QT.getTypePtr()),N),FV(nullptr)
+{
```
awruef: Will fix.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -26,53 +26,94 @@ tyToStr(const Type *T) {
...
+					   tyToStr(QT.getTypePtr()),N),FV(nullptr)
+{   
 	QualType QTy = QT;
 	const Type *Ty = QTy.getTypePtr();
+  // If the type is a decayed type, then maybe this is the result of 
+  // decaying an array to a pointer. If the original type is some 
+  // kind of array type, we want to use that instead. 
+  if (const DecayedType *DC = dyn_cast<DecayedType>(Ty)) {
+    QualType QTytmp = DC->getOriginalType();
+    const Type * Tytmp = QTytmp.getTypePtr();
```
awruef: Probably, I'll try to make it cleaner. `isArrayType` won't work on `Ty` though, in the case of an array like `int arr[]` because it becomes a `DecayedType` that contains both an original type (which *is* an `ArrayType`) and a "pointer type."

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -26,53 +26,94 @@ tyToStr(const Type *T) {
...
+    if (Tytmp->isArrayType() || Tytmp->isIncompleteArrayType()) {
```
awruef: I thought that was the case as well, but in my testing it was not. An incomplete array type would return false for `isArrayType` which really confused me.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -26,53 +26,94 @@ tyToStr(const Type *T) {
...
-   
-    if (Ty->isCheckedPointerType()) {
-      if (Ty->isCheckedPointerPtrType()) {
-        // Constrain V so that it can't be either wild or an array.
-        CS.addConstraint(CS.createNot(CS.createEq(V, CS.getArr()))); 
-        CS.addConstraint(CS.createNot(CS.createEq(V, CS.getWild())));
-        ConstrainedVars.insert(K);
-      } else if (Ty->isCheckedPointerArrayType()) {
-        llvm_unreachable("unsupported!");
+  arrPresent = false;
```
awruef: Will fix.

tools/checked-c-convert/ProgramInfo.cpp
awruef: Okay, I'll look at that. That sounds like what I want here anyway.

tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -26,53 +26,94 @@ tyToStr(const Type *T) {
...
+	while (Ty->isPointerType() || Ty->isArrayType()) {
+    if (Ty->isArrayType() || Ty->isIncompleteArrayType()) {
```
awruef: See above.

Issue #256, opened 2017-04-07
Implement runtime checking for dynamic_bounds_cast


dtarditi:
The dynamic_bounds_cast operator requires runtime checks of bounds.    I think this runtime checking should be inserted during the lowering to clang IR.

dtarditi:
Correct the name of the operator: it is dynamic_bounds_cast.

wonsubkim:
I think that it works as follows
- DynamicBoundsCast/AssumeBoundsCast kinds exist
- For DynamicBoundsCast casting kind, CodeGen emits dynamic bounds check code
  - BoundsCastExpr has bounds for bounds casting operation - bounds(castlb, castub)
  - BoundsCastExpr has inferred bounds for source expression - bounds(lb, ub)
  - It inserts Dynamic_check(lb <= castlb && castub <= ub)
- How does the type casting works?
  - In C++ dynamic_cast, it generates code to call dynamic cast function call
  - In checked-c, how does it work except for generating dynamic check?

lenary:
@wonsubkim I can't say how code generation works for the casting itself, but I can help you with finding the code you need to call.

You want to call [`CGF.EmitDynamicBoundsCheck(const Address PtrAddr, const BoundsExpr *Bounds)`](https://github.com/Microsoft/checkedc-clang/blob/master/lib/CodeGen/CGDynamicCheck.cpp#L75), using the pointer you're casting and the BoundsExpr you have. Sometimes you may not have the pointer as an `Address` object (it may be an LValue, Clang has weird conventions around this). There are methods on the LValue struct to turn it into an address, but you have to use the right one for the kind of LValue you have (there's an enum in the LValue which tells you).

This method will do all the calculation for you. You can see it being used a lot in `CGExpr.cpp`.

wonsubkim:
I am now preparing runtime check insertion for bounds cast
After current bounds casting parsing implementation is completed, I will do PR for this issue

I think that dynamic_bounds_cast code generation consists of two parts
1. if required, dynamic bounds check insertion
2. explicit type casting code generation

For code generation of dynamic check, I refer to EmitDynamicBoundsCheck but it will be little bit different since it requires inferred source bounds not PtrAddr
I plan to implement as follows
`EmitDyanmicBoundsCheck(const BoundsExpr *DynamicBounds, const BoundsExpr *SrcBounds)`

DynamicBounds is bounds generated by bounds cast operator, `bounds(castlb, castub)`
[...]

PR #301, opened 2017-05-17
Implement runtime check insertion for bounds_cast operator (#256)

wonsubkim:
+ Implement runtime checking for bounds_cast (#256)
      + implement emission of dynamic checking for bounds_cast
      it does two things as follows:
      it generates codes for explicit type casting
      since subexpression can be non-pointer type, it SHOULD take care about it
      it also generates dynamic_check for bounds casting by using both bounds of cast and bounds of subexpression
      : bounds (castlb, castub), bounds of cast operation
      : bounds(lb, ub), bounds of subexpression of cast operation
      -> emits non-null check for subexpression (dynamic_check(base != NULL))
      -> emits dynamic check (dynamic_check(lb <= castlb && castub <= ub)
      + add inferred bounds of subexpression since code generation requires source bounds
      when inferring bounds of subexpression, it also binds inferred source bounds to cast operation

### DIFF
--- a/include/clang/AST/Expr.h
+++ b/include/clang/AST/Expr.h
@@ -2798,8 +2798,9 @@ private:
 /// classes).
 class CastExpr : public Expr {
 private:
-  // CASTBOUNDS - for expanded bounds of cast expression
-  // SUBEXPRBOUNDS - for bounds of subexpression
+  // BOUNDS - declared bounds of the result of the cast expression
+  // CASTBOUNDS - inferred bounds of cast expression
+  // SUBEXPRBOUNDS - inferred bounds of subexpression
   enum { OP, BOUNDS, CASTBOUNDS, SUBEXPRBOUNDS, END_EXPR = 4 };
   Stmt* SubExprs[END_EXPR];
 
--- a/include/clang/AST/Expr.h
+++ b/include/clang/AST/Expr.h
@@ -2798,9 +2798,9 @@ private:
 /// classes).
 class CastExpr : public Expr {
 private:
-  // BOUNDS - enum value for bounds of cast expression
-  // SUBBOUNDS - enum value for bounds of subexpression
-  enum { OP, BOUNDS, SUBBOUNDS, END_EXPR = 3 };
+  // CASTBOUNDS - for expanded bounds of cast expression
+  // SUBEXPRBOUNDS - for bounds of subexpression
+  enum { OP, BOUNDS, CASTBOUNDS, SUBEXPRBOUNDS, END_EXPR = 4 };
   Stmt* SubExprs[END_EXPR];
 
   bool CastConsistency() const;
@@ -2836,7 +2836,8 @@ protected:
     assert(kind != CK_Invalid && "creating cast with invalid cast kind");
     SubExprs[OP] = op;
     SubExprs[BOUNDS] = nullptr;
-    SubExprs[SUBBOUNDS] = nullptr;
+    SubExprs[CASTBOUNDS] = nullptr;
+    SubExprs[SUBEXPRBOUNDS] = nullptr;
     CastExprBits.Kind = kind;
     setBasePathSize(BasePathSize);
     assert(CastConsistency());
@@ -2847,7 +2848,8 @@ protected:
     : Expr(SC, Empty) {
     SubExprs[OP] = nullptr;
     SubExprs[BOUNDS] = nullptr;
-    SubExprs[SUBBOUNDS] = nullptr;
+    SubExprs[CASTBOUNDS] = nullptr;
+    SubExprs[SUBEXPRBOUNDS] = nullptr;
     setBasePathSize(BasePathSize);
   }
 
@@ -2920,15 +2922,26 @@ public:
     SubExprs[BOUNDS] = E;
   }
 
-  bool hasSubBoundsExpr() const { return SubExprs[SUBBOUNDS] != nullptr; }
-  BoundsExpr *getSubBoundsExpr() {
-    return cast_or_null<BoundsExpr>(SubExprs[SUBBOUNDS]);
+  bool hasCastBoundsExpr() const { return SubExprs[CASTBOUNDS] != nullptr; }
+  BoundsExpr *getCastBoundsExpr() {
+    return cast_or_null<BoundsExpr>(SubExprs[CASTBOUNDS]);
   }
-  const BoundsExpr *getSubBoundsExpr() const {
-    return const_cast<BoundsExpr*>(cast_or_null<BoundsExpr>(SubExprs[SUBBOUNDS]));
+  const BoundsExpr *getCastBoundsExpr() const {
+    return const_cast<BoundsExpr*>(cast_or_null<BoundsExpr>(SubExprs[CASTBOUNDS]));
   }
-  void setSubBoundsExpr(BoundsExpr *E) {
-    SubExprs[SUBBOUNDS] = E;
+  void setCastBoundsExpr(BoundsExpr *E) {
+    SubExprs[CASTBOUNDS] = E;
+  }
+
+  bool hasSubExprBoundsExpr() const { return SubExprs[SUBEXPRBOUNDS] != nullptr; }
+  BoundsExpr *getSubExprBoundsExpr() {
+    return cast_or_null<BoundsExpr>(SubExprs[SUBEXPRBOUNDS]);
+  }
+  const BoundsExpr *getSubExprBoundsExpr() const {
+    return const_cast<BoundsExpr*>(cast_or_null<BoundsExpr>(SubExprs[SUBEXPRBOUNDS]));
+  }
+  void setSubExprBoundsExpr(BoundsExpr *E) {
+    SubExprs[SUBEXPRBOUNDS] = E;
   }
 };
 
--- a/include/clang/AST/Expr.h
+++ b/include/clang/AST/Expr.h
@@ -2799,7 +2799,7 @@ private:
 class CastExpr : public Expr {
 private:
   // BOUNDS - declared bounds of the result of the cast expression
-  // CASTBOUNDS - inferred bounds of cast expression
+  // CASTBOUNDS - normalized version of BOUNDS
   // SUBEXPRBOUNDS - inferred bounds of subexpression
   enum { OP, BOUNDS, CASTBOUNDS, SUBEXPRBOUNDS, END_EXPR = 4 };
   Stmt* SubExprs[END_EXPR];
--- a/include/clang/AST/Expr.h
+++ b/include/clang/AST/Expr.h
@@ -2798,7 +2798,9 @@ private:
 /// classes).
 class CastExpr : public Expr {
 private:
-  enum { OP, BOUNDS, END_EXPR = 2 };
+  // BOUNDS - enum value for bounds of cast expression
+  // SUBBOUNDS - enum value for bounds of subexpression
+  enum { OP, BOUNDS, SUBBOUNDS, END_EXPR = 3 };
   Stmt* SubExprs[END_EXPR];
 
   bool CastConsistency() const;
@@ -2834,6 +2836,7 @@ protected:
     assert(kind != CK_Invalid && "creating cast with invalid cast kind");
     SubExprs[OP] = op;
     SubExprs[BOUNDS] = nullptr;
+    SubExprs[SUBBOUNDS] = nullptr;
     CastExprBits.Kind = kind;
     setBasePathSize(BasePathSize);
     assert(CastConsistency());
@@ -2844,6 +2847,7 @@ protected:
     : Expr(SC, Empty) {
     SubExprs[OP] = nullptr;
     SubExprs[BOUNDS] = nullptr;
+    SubExprs[SUBBOUNDS] = nullptr;
     setBasePathSize(BasePathSize);
   }
 
@@ -2914,6 +2918,17 @@ public:
 
   void setBoundsExpr(BoundsExpr *E) {
     SubExprs[BOUNDS] = E;
+  }
+
+  bool hasSubBoundsExpr() const { return SubExprs[SUBBOUNDS] != nullptr; }
+  BoundsExpr *getSubBoundsExpr() {
+    return cast_or_null<BoundsExpr>(SubExprs[SUBBOUNDS]);
+  }
+  const BoundsExpr *getSubBoundsExpr() const {
+    return const_cast<BoundsExpr*>(cast_or_null<BoundsExpr>(SubExprs[SUBBOUNDS]));
+  }
+  void setSubBoundsExpr(BoundsExpr *E) {
+    SubExprs[SUBBOUNDS] = E;
   }
 };
 
--- a/lib/CodeGen/CGDynamicCheck.cpp
+++ b/lib/CodeGen/CGDynamicCheck.cpp
@@ -141,6 +141,36 @@ void CodeGenFunction::EmitDynamicBoundsCheck(const Address BaseAddr,
 
   ++NumDynamicChecksRange;
 
+  // Dynamic_check(Base == NULL || (Lower <= CastLower && CastUpper <= Upper))
+  // Emit code blocks as follows:
+  // if (Base == NULL) {...}
+  // else {
+  // if (Lower <= CastLower && CastUpper <= Upper) {...}
+  // else { trap(); llvm_unreachable(); }
+  // }
+
+  Value *Cond1 =
+            Builder.CreateIsNull(BaseAddr.getPointer(), "_Dynamic_check.null");
+
+  // Constant Folding:
+  // If we have generated a constant condition, and the condition is true,
+  // then the check will always pass and we can elide it.
+  if (const ConstantInt *ConditionConstant = dyn_cast<ConstantInt>(Cond1)) {
+    if (ConditionConstant->isOne()) {
+      ++NumDynamicChecksElided;
+      return;
+    }
+  }
+
+  ++NumDynamicChecksInserted;
+
+  BasicBlock *Begin, *DyCkSuccess, *DyCkFail, *DyCkFallThrough;
+  Begin = Builder.GetInsertBlock();
+  DyCkSuccess = createBasicBlock("_Dynamic_check.succeeded");
+  DyCkFallThrough = createBasicBlock("_Dynamic_check.fallthrough", this->CurFn);
+  DyCkFail = createBasicBlock("_Dynamic_check.failed", this->CurFn);
+
+  Builder.SetInsertPoint(DyCkFallThrough);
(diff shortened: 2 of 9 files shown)


include/clang/AST/Expr.h
```diff
@@ -2798,7 +2798,9 @@ class BoundsExpr : public Expr {
 /// classes).
 class CastExpr : public Expr {
 private:
-  enum { OP, BOUNDS, END_EXPR = 2 };
+  // BOUNDS - enum value for bounds of cast expression
+  // SUBBOUNDS - enum value for bounds of subexpression
+  enum { OP, BOUNDS, SUBBOUNDS, END_EXPR = 3 };
```
dtarditi: I noticed that the code is overwriting the bounds expression at BOUNDS during bounds inference.  This is to put it into the form of a range.   I don't think we should be overwriting the bounds expression at BOUNDS.   This means that the AST dump would not reflect what the user wrote, for example.   

I think there should be a new entry for the converted bounds too.

include/clang/AST/Expr.h
```diff
@@ -2798,7 +2798,9 @@ class BoundsExpr : public Expr {
 /// classes).
 class CastExpr : public Expr {
 private:
-  enum { OP, BOUNDS, END_EXPR = 2 };
+  // BOUNDS - enum value for bounds of cast expression
+  // SUBBOUNDS - enum value for bounds of subexpression
```
dtarditi: Could use the name SUBEXPR_BOUNDS instead of SUBBOUNDS?    The term SUBBOUNDS is ambiguous.  It could also mean a subrange of the bounds?

lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -111,6 +111,66 @@ void CodeGenFunction::EmitDynamicBoundsCheck(const Address PtrAddr, const Bounds
   EmitDynamicCheckBlocks(Builder.CreateAnd(LowerChk, UpperChk, "_Dynamic_check.range"));
 }
 
+void CodeGenFunction::EmitDynamicBoundsCheck(const BoundsExpr *DynamicBounds,
```
dtarditi: Could you rename `DynamicBounds` to `CastBounds` and `SrcBounds` to `SubExprBounds`?   That will make it easier to remember which bounds are which?

lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -111,6 +111,66 @@ void CodeGenFunction::EmitDynamicBoundsCheck(const Address PtrAddr, const Bounds
...
+  if (!getLangOpts().CheckedC)
+    return;
+
+  if (!SrcBounds || !DynamicBounds)
+    return;
+
+  if (SrcBounds->isAny() || SrcBounds->isInvalid())
+    return;
+
+  if (DynamicBounds->isAny() || DynamicBounds->isInvalid())
```
dtarditi: You can't write a bounds of `Any` in the language, so I think we should assert if we see a DynamicBounds of any.

lib/CodeGen/CGExpr.cpp
```diff
@@ -3664,13 +3665,16 @@ LValue CodeGenFunction::EmitCastLValue(const CastExpr *E) {
     return MakeNaturalAlignAddrLValue(EmitDynamicCast(V, DCE), E->getType());
   }
 
+  case CK_DynamicPtrBounds:
+  case CK_AssumePtrBounds:
```
dtarditi: There should not be a dynamic check for AssumePtrBounds.

lib/CodeGen/CGExpr.cpp
```diff
@@ -4306,3 +4310,41 @@ RValue CodeGenFunction::EmitPseudoObjectRValue(const PseudoObjectExpr *E,
...
+    bool InputSigned = E->getType()->isSignedIntegerOrEnumerationType();
+    llvm::Value* IntResult =
+      Builder.CreateIntCast(Src, MiddleTy, InputSigned, "conv");
+    llvm::Value *Result = Builder.CreateIntToPtr(IntResult, DestLLVMTy);
+    CharUnits Align = getContext().getTypeAlignInChars(DestTy);
+    Addr = Address(Result, Align);
+  }
+  if (Kind == CK_DynamicPtrBounds) {
+    BoundsCastExpr *BCE = cast<BoundsCastExpr>(CE);
+    EmitDynamicNonNullCheck(Addr, E->getType());
```
dtarditi: Actually, dynamic_cast should always succeed when the runtime value of E is null.  We don't want to fail if e is null: we want to skip the dynamic bounds check instead.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -1444,6 +1444,10 @@ Value *ScalarExprEmitter::VisitCastExpr(CastExpr *CE) {
     return CGF.EmitDynamicCast(V, DCE);
   }
 
+  case CK_DynamicPtrBounds:
+  case CK_AssumePtrBounds:
+    return CGF.EmitDynamicBoundsCast(CE);
```
dtarditi: We don't want a dynamic check for AssumePtrbounds.

lib/CodeGen/CodeGenFunction.h
```diff
@@ -2288,10 +2288,13 @@ class CodeGenFunction : public CodeGenTypeCache {
   void EmitDynamicNonNullCheck(const Address BaseAddr, const QualType BaseTy);
   void EmitDynamicOverflowCheck(const Address BaseAddr, const QualType BaseTy, const Address PtrAddr);
   void EmitDynamicBoundsCheck(const Address PtrAddr, const BoundsExpr *Bounds);
+  void EmitDynamicBoundsCheck(const BoundsExpr* DynamicBounds, const BoundsExpr *SrcBounds);
```
dtarditi: Please rename these parameters to match the names that I suggested for the definition.

lib/Sema/SemaBounds.cpp
```diff
@@ -942,10 +942,15 @@ namespace {
...
         Expr *subExpr = E->getSubExpr();
         BoundsExpr *subExprBounds = S.InferRValueBounds(subExpr);
+        BoundsExpr *SrcBounds = S.InferRValueBounds(E);
 
         if (subExprBounds->isNone()) {
           S.Diag(subExpr->getLocStart(), diag::err_expected_bounds);
         }
+
+        assert(SrcBounds);
+        E->setBoundsExpr(SrcBounds);
```
dtarditi: We should not overwrite the bounds expression that was declared by the user.   I have suggested another slot in CastExpr to hold these modified bounds.

lib/CodeGen/CGExprScalar.cpp
wonsubkim: EmitDynamicBoundsCast has two functionalities
First, it is code generation for explicit type casting because bounds_cast operation is also explicit type casting operatoin
Second, it emits dynamic_checks for dynamic_bounds_cast only
In EmitDynamicBoundsCast method, it generats casting operation for both bounds cast type
and it only generates bounds operation for dynamic_bounds_cast operation

lib/CodeGen/CGExpr.cpp
wonsubkim: Thanks, I will apply this

lib/CodeGen/CGExpr.cpp
wonsubkim: This issue is same as below for EmitDynamicBoundsCast method
For AssumePtrBounds, it only generaets explicit type casting operation not dynamic bounds check

include/clang/AST/Expr.h
wonsubkim: I desinged as follows:
In bounds inferference, count/byte_count bounds information is expanded into range bounds information. Also, In code generation, it requires range bounds expression.
So I have overwritten bounds expression in bounds inference
In code generation, it requires also source bounds expression to compare bounds expression of cast operation. Since I do not have any idea to pass source bounds expression in code generation, I have set source bounds expression when bounds inference
I think that it is correct to expand range bounds expression of bounds cast expression in semantic analysis phase because code generation does not handle this
How about your idea?

lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -111,6 +111,66 @@ void CodeGenFunction::EmitDynamicBoundsCheck(const Address PtrAddr, const Bounds
...
+  Address CastUpper = EmitPointerWithAlignment(CastRange->getUpperExpr());
+
+  Value *CastLowerInt = Builder.CreatePtrToInt(CastLower.getPointer(), IntPtrTy,
+                                               "_Dynamic_check.castlower");
+  Value *CastUpperInt = Builder.CreatePtrToInt(CastUpper.getPointer(), IntPtrTy,
+                                               "_Dynamic_check.castupper");
+
+  // Make the lower check (Lower <= CastLower)
+  Value *LowerChk =
+      Builder.CreateICmpSLE(LowerInt, CastLowerInt, "_Dynamic_check.lower_cmp");
```
lenary: This should be `Builder.CreateICmpULE` (david has changed this in my code on master). Same goes for UpperChk.

wonsubkim: Thanks. I will apply this

lib/CodeGen/CGExprScalar.cpp
dtarditi: Oh, OK.  Could you rename this to EmitBoundsCast then?   That will be clearer.

include/clang/AST/Expr.h
dtarditi: Instead of overwriting the pointer to the existing bounds expression, I would like to have a new pointer that is set during the semantic analysis phase.    I suggest adding another entry to the SubExprs array for this pointer.

lib/CodeGen/CGExprScalar.cpp
wonsubkim: Okay, Thanks

lib/CodeGen/CGExpr.cpp
wonsubkim: In below code, expression E is subexpression of bounds cast operation (bounds_cast<T>(e))
expression E can be NULL?
I think that all cast expression has non-null subexpression.
Therefore, I think that it is not necessary to check if subexpression of bounds cast operation is NULL

lib/CodeGen/CGDynamicCheck.cpp
wonsubkim: You are right. I wll fix up this

include/clang/AST/Expr.h
wonsubkim: Okay, I will implement as follows:
SubExpr[SUBEXPRBOUNDS] - represent expanded/inferred bounds expression of subexpression
SubExpr[CASTBOUNDS] - represent expanded/inferred bounds expression of bounds cast operation
In code generation, it use those two kinds of bounds expression in bounds cast operation
Those are all inferred (expanded) from original bounds expression

lib/CodeGen/CGExpr.cpp
dtarditi: The subexpression E in the compiler  IR cannot be a null pointer.  However, the value of E at program runtime can be a null pointer.   I am concerned about the value of E at program runtime.   It is necessary to check at program runtime if the value of the subexpression is equal to NULL (0).   If it is, the check should always succeed.

include/clang/AST/Expr.h
dtarditi: Yes, that sounds good.

lib/CodeGen/CGExpr.cpp
wonsubkim: Let CastBounds be bounds (castlb, castub)
Let SubExprBounds be bounds (lb, ub)
I have designed to emit as follows:
```
dynamic_check(E != NULL);
dynamic_check(lb <= castlb && castub <= ub);
```
However, you mean that it is wrong and it SHOULD emit as follows:
`dynamic_check (E == NULL || (lb <= castlb && castub <= ub));`

Is it correct?
Can you explain why dynamic_check is skipped if E is NULL?

dtarditi: Yes, that is the correct `dynamic_check` expression.  I will explain tomorrow why we skip the dynamic check if E is NULL.

wonsubkim:
I have applied your feedback
+ dynamic_check(E == NULL || (lb <= castlb && castub <= ub))
+ EmitDynamicCheckBlocks(cond1, cond2) interface to handle two conditions
+ add enum value for CASTBOUNDS, SUBEXPRBOUNDS
+ rename function & parameters

include/clang/AST/Expr.h
```diff
@@ -2798,7 +2798,9 @@ class BoundsExpr : public Expr {
 /// classes).
 class CastExpr : public Expr {
 private:
-  enum { OP, BOUNDS, END_EXPR = 2 };
+  // CASTBOUNDS - for expanded bounds of cast expression
+  // SUBEXPRBOUNDS - for bounds of subexpression
```
dtarditi: Could a comment describing BOUNDS too?  It is the declared bounds of the result of the cast expression.

lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -111,6 +111,76 @@ void CodeGenFunction::EmitDynamicBoundsCheck(const Address PtrAddr, const Bounds
...
+
+  // if expression E is NULL, skip dynamic_check
+  // Dynamic_check(E == NULL || (lb <= castlb && castub <= ub))
+  // if (cond1) {...}
+  // else {
+  // if (cond2) {...}
+  // else { trap(); llvm_unreachable(); }
+  // }
+  EmitDynamicCheckBlocks(
+      Builder.CreateIsNull(BaseAddr.getPointer(), "_Dynamic_check.null"),
```
dtarditi: I think you should check for the non-null case before computing the upper and lower bound comparisons.   This doesn't quite match the semantics of the check.  In C, `(e1 || e2)` means that we first evaluate `e1`.    If that fails, then we evaluate `e2'.     In your implementation, we are always evaluating `e1` and `e2`.   Your implementation corresponds to `tmp1 = e1; tmp2 = e2`; if (tmp1 || tmp2)`.

The problem with always evaluating `e2' is that the upper/lower bound comparisons could end up doing pointer arithmetic involving null.   We are going to add runtime checks that disallow pointer arithmetic involving null.

wonsubkim: A newly implemented `EmitDynamicCheckBlocks(cond1, cond2)` generates code as follows:
```
if (cond1) { success_block }
else {
fallthrough_block
  if (cond2) {  success_block  }
  else {  fail_block  }
}
```
In current implementation, it emits dynamic check blocks for `(cond1|| cond2)`
cond1 is (E == NULL) and cond2 is (lb <= castlb && castub <= ub)
Therefore, if subexpression itself is NULL, it skips lower/upper bounds check
if subexpression is not NULL, then it also checks if lower/upper bounds condition is met or not
I think that it is equivalent for `dynamic_check (base == NULL || (lb <= castlb && castub <= ub))`

dtarditi:
There was a question about why a `dynamic_bounds_cast` operation should always succeed if the runtime value of subexpression of the cast is `NULL` (0).    The main reason is the definition of the meaning of a bounds expression, which is given in the 2nd-to-last paragraph of Section 3.1 of version 0.6 of the Checked C specification.   For an expression with bounds, either the expression value is 0 or the value must be in the range of the bounds..  It follows that the runtime value 0 is always valid for the result of the `dynamic_bounds_cast` operation.  It satisfies the meaning of bounds for the resulting bounds.

Put another way, `NULL` has bounds `Any`.  This means that `NULL` is valid given for any specific bounds for the `dynamic_bounds_cast` operator.

There is a more subtle problem that we must avoid.  If the value of the subexpression is `NULL`, it means that the inferred bounds for that subexpression may be invalid.   It is even worse than that: computing the inferred bounds for the subexpression or desired the bounds for `dynamic_bounds_cast` operation could fail with a runtime error.  The specification does not allow pointer arithmetic involving `NULL`, although this runtime check has has not been implemented yet.

lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -111,6 +111,76 @@ void CodeGenFunction::EmitDynamicBoundsCheck(const Address PtrAddr, const Bounds
...
+  const RangeBoundsExpr *SubRange = dyn_cast<RangeBoundsExpr>(SubExprBounds);
+  const RangeBoundsExpr *CastRange = dyn_cast<RangeBoundsExpr>(CastBounds);
+
+  ++NumDynamicChecksRange;
+
+  // SubRange - bounds(lb, ub) vs CastRange - bounds(castlb, castub)
+  // Dynamic_check(lb <= castlb && castub <= ub)
+
+  // Emit the code to generate the pointer values
+  Address Lower = EmitPointerWithAlignment(SubRange->getLowerExpr());
```
dtarditi: My understanding is that LLVM IR is a linearized sequence of instructions.   It is not an expression tree.   I believe that means that a sequence of instructions will be emitted here to compute the lower bound.   It is incorrect to compute the lower bound (or the upper bound) if BaseAddr is null.     The problem is the lower bound computation could fail at runtime if it involves a null pointer.   We must check that BaseAddr is not null before doing any of these computations.

wonsubkim: Thanks for comments
I got it what you mean. I have a mistake to do that
I will fix it and check the generated output code.

include/clang/AST/Expr.h
```diff
@@ -2798,7 +2798,10 @@ class BoundsExpr : public Expr {
 /// classes).
 class CastExpr : public Expr {
 private:
-  enum { OP, BOUNDS, END_EXPR = 2 };
+  // BOUNDS - declared bounds of the result of the cast expression
+  // CASTBOUNDS - inferred bounds of cast expression
```
dtarditi: Could you change the comment for CASTBOUNDS to say 'normalized version of BOUNDS'?

lib/CodeGen/CGExpr.cpp
```diff
@@ -4306,3 +4310,41 @@ RValue CodeGenFunction::EmitPseudoObjectRValue(const PseudoObjectExpr *E,
...
   return emitPseudoObjectExpr(*this, E, true, AggValueSlot::ignored()).LV;
 }
+
+llvm::Value *CodeGenFunction::EmitBoundsCast(CastExpr *CE) {
+  Expr *E = CE->getSubExpr();
+  QualType DestTy = CE->getType();
+  CastKind Kind = CE->getCastKind();
+  // Bounds casting has two functionalities
+  // - code generation for explicit type casting
+  // - code generation for dynamic check, dynamic_bounds_cast only
```
dtarditi: Could you update this comment?  I believe the dynamic_check expressions below need to be updated.

Issue #302, opened 2017-05-17
Function return type with bounds-safe interface is not properly checked in checked scope


wonsubkim:
While testing calloc/malloc header file, unexpected errors happen

In checked scope, function having return type, that is unchecked pointer type with bounds-safe interface, is not properly checked in checked scope.
Since it has proper bounds-safe interface, it SHOULD be considered as checked pointer type but compiler generates an error for this

```
void * calloc(size_t nmemb, size_t size) : byte_count(nmemb * size);
array_ptr<ptr<node_t>> retval : count(size) = calloc(size, sizeof(ptr<node_t>));
```

wonsubkim:
While testing checkedc-llvm-test-suite, I have found some missing errors
1. When checking declaration in checked scope, I have missed unexpected type case. Fixed it

2. When generating function call in checked scope, I have generated implicit cast expression for checked pointer type for FunctionToPointerDecay
In function declaration having KNR parameters, it has empty parameter types
For function call for KNR function , SemaBounds generates implicit casting expression from incompatible type.
For call expression, DO NOT generate checked pointer type of function

3. In code generation of call instruction (EmitCallArgs), bounds-safe interface has similar problem to interoperation type
we SHOULD handle same conversion for Sema.GetCheckedCInteropType
I have also handled unchecked pointer type with bounds-safe interface not also interop type

dtarditi:
These issues have been addressed- thanks Wonsub!

For point 2 above, we decided to keep generating an implicit cast to checked pointer type for FunctoinToPointerDecay.   We decided to not allow calls to functions declared in the old K&R prototype style in checked scopes.  Calls to such functions are treated as no-prototype calls, so there could be mismatches between actual arguments and declared argument.

PR #303, opened 2017-05-17
Fix up bugs related to #302

wonsubkim:
+ fix up bugs related to #302
  + if function return type is unchecked pointer type with bounds-safe interface, it is not checked properly
  For function declaration, it SHOULD consider its return type not function type itself


lenary:
This is no longer causing issues in the test suite with `calloc`, and the change seems reasonable, but I want david to approve it before you merge it.

PR #309, opened 2017-05-18
Implementation-specific definitions of NULL in checked scopes.

dtarditi:
The C standard allows NULL to be an implementation-defined null pointer
constant.  The clang implementation of C allows NULL to be defined as
((void *) 0).   This is rejected by the static checking of checked scopes
currently because it is a cast to an unchecked pointer.   However, we would
like to allow NULL to be used in checked scopes.

The solution is to allow null pointer constants to appear in checked scopes,
however an implementation defines them.  This changes modifies the checking
of casts in checked scopes to allow ((void *) 0).  We may need to do some
extra work to preserve the safety property of checked scopes. You can't
do pointer arithmetic because void is an incomplete type or convert it some
other unchecked pointer type in a checked scope.  You may be able to try
to dereference it.  This could lead to undefined behavior, which could
lead to safety problems given how C compilers optimize in the presence
of undefined behavior.

Testing:
- Add a new clang-specific test for its implementation defined null pointer
  constants.

### DIFF
--- a/lib/Sema/SemaCast.cpp
+++ b/lib/Sema/SemaCast.cpp
@@ -2565,7 +2565,15 @@ void CastOperation::CheckCStyleCast() {
     unsigned TypeKind = 0;
     bool HasUncheckedType = DestType->hasUncheckedType(TypeKind);
     bool HasVariadicType = DestType->hasVariadicType();
-    if (HasUncheckedType || HasVariadicType) {
+    bool ConstructsNullPointer = false;
+    if (HasUncheckedType)
+      if (DestType->isVoidPointerType() &&
+          !SrcExpr.isInvalid()) {
+        const IntegerLiteral *Lit = dyn_cast<IntegerLiteral>(SrcExpr.get());
+        if (Lit && !Lit->getValue())
+         ConstructsNullPointer = true;
+      }
+    if ((HasUncheckedType && !ConstructsNullPointer) || HasVariadicType) {
       if (HasUncheckedType) {
         Self.Diag(OpRange.getBegin(), diag::err_checked_scope_type_for_casting)
             << TypeKind;
--- a/test/CheckedC/checked-scope-null.c
+++ b/test/CheckedC/checked-scope-null.c
@@ -1,0 +1,11 @@
+// Tests that implementation-specific definitions of NULL can be used
+// in checked scopes.
+
+// RUN: %clang_cc1 -verify -fcheckedc-extension -Wno-unused-value %s
+// expected-no-diagnostics
+
+_Checked _Ptr<int> f(void) {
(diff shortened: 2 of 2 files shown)


lib/Sema/SemaCast.cpp
```diff
@@ -2565,7 +2565,15 @@ void CastOperation::CheckCStyleCast() {
...
     bool HasVariadicType = DestType->hasVariadicType();
-    if (HasUncheckedType || HasVariadicType) {
+    bool ConstructsNullPointer = false;
+    if (HasUncheckedType)
+      if (DestType->isVoidPointerType() &&
+          !SrcExpr.isInvalid()) {
+        const IntegerLiteral *Lit = dyn_cast<IntegerLiteral>(SrcExpr.get());
+        if (Lit && !Lit->getValue())
+         ConstructsNullPointer = true;
+      }
```
wonsubkim: Looks good :)

I have a question about this strategy.
Is only void type casting of zero `(void*)0` allowed?
non-void type casting of zero such as` (int*)0` or `(ptr<int>)0` is not allowed?

lenary: I think it might be clearer if you use the following method on SrcExpr, which will do the check better than we ever can: https://clang.llvm.org/doxygen/classclang_1_1Expr.html#aa1d79a39abcf576a82b97cff17f893f8

This will also do the right thing for other kinds of NULL (yes, there are other null expressions that clang supports).

As for `(int*)0`, it seems reasonable not to support it yet. If someone has written their own `NULL_INT_PTR` macro in a header, then they have to change it if they want to use it in a checked pointer.

`(ptr<int>)0` is allowed as this is entirely checked.

dtarditi:
Wonsub, if you are OK with this pull request, could you approve it?

Yes, only (void *)0 is allowed for unchecked pointer types.  As Sam points out, (ptr<int>)(0) is already allowed.   The intent is to support the implementation-specific definition of NULL found in header files on Mac and Linux.  They define NULL to be ((void *)0).

It would be hard to change the definition of NULL because of the way the preprocessor works.  By the time the compiler sees it, it is already ((void *)0).

Sam,  thanks for the suggestion - I looked at that function too.  The function assumes that the cast expression has already been built, which is why I didn't use it.  We are doing the checking in the midst of building a cast expression, with the goal of putting out an error message as soon as possible.  We could use that function if we moved the checking elsewhere.  For now, I made the expedient choice to put a simple test at the current check.

wonsubkim:
Okay. I will approve this pull request
Maybe I don't have authority to approve this pull request
After it is merged, I will also change llvm-test-suite to remove unnecessary NULL definition

PR #317, opened 2017-06-05
Copy Checked C headers to clang system header directory.

dtarditi:
The Checked C header files are stored in the Checked C repo.  If the Checked C repo is in the tree, add a dependency on the target that copies the Checked C headers.   The target is conditionally added by the LLVM cmake files and a variable is set for use by the clang build system.

### DIFF
--- a/tools/driver/CMakeLists.txt
+++ b/tools/driver/CMakeLists.txt
@@ -52,7 +52,7 @@ endif()
 add_dependencies(clang clang-headers)
 
 if (DEFINED CHECKEDC_IN_TREE)
-  add_dependencies(checkedc-headers)
+  add_dependencies(clang checkedc-headers)
 endif()
 
 if(NOT CLANG_LINKS_TO_CREATE)
--- a/tools/driver/CMakeLists.txt
+++ b/tools/driver/CMakeLists.txt
@@ -51,6 +51,10 @@ endif()
 
 add_dependencies(clang clang-headers)
 
+if (DEFINED CHECKEDC_IN_TREE)
+  add_dependencies(checkedc-headers)
+endif()
+
 if(NOT CLANG_LINKS_TO_CREATE)
   set(CLANG_LINKS_TO_CREATE clang++ clang-cl)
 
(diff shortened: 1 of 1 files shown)


tools/driver/CMakeLists.txt
```diff
@@ -51,6 +51,10 @@ endif()
 
 add_dependencies(clang clang-headers)
 
+if (DEFINED CHECKEDC_IN_TREE)
+  add_dependencies(checkedc-headers)
```
lenary: This should be `add_dependencies(clang checkedc-headers)` right? I think with just one argument isn't valid.

dtarditi: Good catch!  Fixed.

lenary:
Ok, even with the changes, `make clang` in my build directory doesn't invoke the `checkedc-headers` target. This is a little infuriating, and I don't know why. I'm looking further into this.

dtarditi:
I pushed a fix to the Checked C LLVM repo about 2 hours ago.     There was a change sitting on my local disk that set the variable used by this piece of code.   Make sure that you pick up that fix too.

lenary:
I have, i think it's a scoping issue, given the following output when I add print messages...

```
-- Constructing LLVMBuild project information
-- Targeting X86
CHECKEDC_IN_TREE set: TRUE
-- Found SDK version 10.12
[snip ...]
-- Clang version: 4.0.0
CHECKEDC_IN_TREE checked:
-- Configuring done
```

I think I've worked out a potential fix, let me see if it works. It involves pulling the "CHECKEDC_IN_TREE" all the way up to the top-level CMakeLists.txt as has been done for POLLY_IN_TREE

tools/driver/CMakeLists.txt
```diff
@@ -51,6 +51,10 @@ endif()
 
 add_dependencies(clang clang-headers)
 
+if (DEFINED CHECKEDC_IN_TREE)
```
lenary: Can we also change this to `if(CHECKEDC_IN_TREE)` as it should always be defined with the patch I've just sent you for checkedc-llvm.

PR #320, opened 2017-06-12
Adding _For_any() specifier to function definition specifier.

jp1729:
* Detects polymorphic type declaration
 * Adds type declaration to correct scope, and exits scope at the end of function defintion. 

Currently the polymorphic type declaration is treated as a typedef void *.


lib/Parse/ParseDecl.cpp
```diff
@@ -6822,6 +6843,85 @@ void Parser::ParseCheckedPointerSpecifiers(DeclSpec &DS) {
...
+      polyDS.SetStorageClassSpec(Actions, DeclSpec::SCS_typedef, StartLoc, 
+                                 PrevSpec, DiagID, Policy);
+      polyDS.SetTypeSpecType(DeclSpec::TST_void, StartLoc, PrevSpec, DiagID, 
+                             Policy);
+      DS.SetRangeStart(ForAnyStartLoc);
+      DS.SetRangeEnd(SourceLocation());
+      Declarator D(polyDS, Declarator::TheContext::BlockContext);
+      D.SetRangeBegin(ForAnyStartLoc);
+      D.SetRangeEnd(SourceLocation());
+      D.AddTypeInfo(DeclaratorChunk::getPointer(DS.getTypeQualifiers(), StartLoc,
```
dtarditi: I am not sure that I understood this line.   We want to create a typedef'ed name to a void * pointer.   Is this picking up information from the decl spec that is being declared and adding it to the void * type?    If it is, that doesn't seem right.

jp1729: You are right. Thank you for catching this. This is completely wrong.This line should pick up information from the newly created decl spec, instead of the old decl spec.

jp1729:
This pull request is ready for further review. I made changes to all the suggestions provided.

jp1729:
Thank you!
I ran the clang regression test suite, and it looks like there are 37 test cases that failed. Upon inspecting a few test cases, none of them seems to have any correlation with the changes I made. For example, one file was checking for "env," while another file was checking for get-cursor command, etc.

I'm a little confused as to why there are these failures. Right now, I'm building the master checked-c clang to see if same test cases fail or not. I will update when the compilation/testing finishes.

jp1729:
I was able to successfully run the regression test and encountered 0 unexpected failures.

PR #324, opened 2017-06-19
Parsing generic function calls.

jp1729:
This change allows clang to parse generic function calls. The generic function call syntax
foo<type, ...>(parameter) will only be accepted if the function "foo" is generic function, declared by specifying _For_any specifier.

The change has been tested with clang regression test and passed.
Further test cases will be added to checkedc and there will be a separate pull request.


include/clang/Basic/DiagnosticParseKinds.td
```diff
@@ -1187,5 +1187,14 @@ def err_forany_unexpected_token : Error<
 def err_forany_comma_or_parenthesis_expected : Error<
   "expected , or )">;
 
+def err_generic_function_unexpected_token : Error<
+  "unexpected token in generic function call">;
+
+def erro_type_function_comma_or_greater_expected : Error<
```
dtarditi: There's a typo in the name of the error message.

lib/Parse/ParseExpr.cpp
```diff
@@ -1386,6 +1386,55 @@ ExprResult Parser::ParseCastExpression(bool isUnaryExpression,
   // are compiling for OpenCL, we need to return an error as this implies
   // that the address of the function is being taken, which is illegal in CL.
 
+  // With primary-expression, before parsing postfix-expression, check to make
+  // sure that Expr is declRefExpr of generic function with _For_any specifier
+  Expr* resExpr = Res.get();
```
dtarditi: Please add spaces put the '*'* next to the variable name: `Expr *resExpr`, not `Expr* resExpr`.

lib/Parse/ParseExpr.cpp
```diff
@@ -1386,6 +1386,55 @@ ExprResult Parser::ParseCastExpression(bool isUnaryExpression,
...
+  if (resExpr != NULL && isa<DeclRefExpr>(resExpr)) {
+    DeclRefExpr* declRef = dyn_cast<DeclRefExpr>(resExpr);
+    ValueDecl* resDecl = declRef->getDecl();
+    if (resDecl != NULL && isa<FunctionDecl>(resDecl)) {
```
dtarditi: Please use nullptr, not NULL.   You can also just use `if (resDecl && isa<FunctionDecl>...`

lib/Parse/ParseExpr.cpp
```diff
@@ -1386,6 +1386,55 @@ ExprResult Parser::ParseCastExpression(bool isUnaryExpression,
   // are compiling for OpenCL, we need to return an error as this implies
   // that the address of the function is being taken, which is illegal in CL.
 
+  // With primary-expression, before parsing postfix-expression, check to make
```
dtarditi: Could you move your new parsing code into a separate function?   I  think it will make the code easier to maintain and understand in the long run.  I  think it would be good for your comment for the function to provide a more high-level description of what this code did.

lib/Parse/ParseExpr.cpp
```diff
@@ -1386,6 +1386,55 @@ ExprResult Parser::ParseCastExpression(bool isUnaryExpression,
...
+        // In this case, parse for <types, ...> in generic function call.
+        ExpectAndConsume(tok::less);
+        bool breakOutOfWhile = false;
+        bool lastEncounteredIsType = false;
+        while (!breakOutOfWhile) {
+          switch (Tok.getKind()) {
+          case tok::comma :
+          case tok::greater :
+            if (!lastEncounteredIsType) {
+              Diag(Tok, diag::err_generic_function_unexpected_token);
```
dtarditi: You need to add some error recovery code to this parsing code.   It might easier of the code had the structure:
1. Parse the less than.
2. See if the next token is greater than.  If it is, stop parsing.
3. If the next token is not greater than, parse the list of type names.  Repeat until greater than is reached or an error occurs:
- Parse type name.
- See if the next token is a ',' or '>'.   If it is not, report an error.
- If it is a comma, consume it.

If there is an error, we should at least try to consume all tokens until a 'greater than' or a ';" is hit.   If you look at ParseBoundsExpression, you'll see some error recovery code.

jp1729:
I am ready for further review.

include/clang/Basic/DiagnosticParseKinds.td
```diff
@@ -1187,5 +1187,11 @@ def err_forany_unexpected_token : Error<
 def err_forany_comma_or_parenthesis_expected : Error<
   "expected , or )">;
 
+def err_expected_list_of_types_expr_for_generic_function : Error<
```
dtarditi: I would suggest changing the error message to "expected a list of type arguments for a generic function".

jp1729:
It is ready for further review. Also, did clang regression test and it's passed.

jp1729:
adding functionality in the parser to read generic function call.

PR #333, opened 2017-06-26
Removes some Clang Build Warnings

lenary:
I had warnings for several things, which were annoying me.

- in `ASTDumper.cpp` clang has deduced the default branch is unneeded. We'll get warnings if we introduce another value to this enum, which is better than missing this case and printing "<<err>>".
- In `ParsePragma.cpp`, the `Actions` field of `PragmaCheckedScopeHandler` is currently totally unused, so I deleted it. If we want it for an `ActOnCheckedScopePragma` function we can reintroduce it.
- in `ParseStmt.cpp` `CheckedLoc` is unused. 
- in `SemaType.cpp`, assert much prefers a boolean to a string. Used llvm_unreachable which is semantically what we want here.
- in `ProgramInfo.h`, as these constraint classes all have virtual functions, clang wants them to have virtual destructors, so i just declare them as virtual. I haven't worked out if they need custom destruction behaviour.

### DIFF
--- a/lib/AST/ASTDumper.cpp
+++ b/lib/AST/ASTDumper.cpp
@@ -2559,7 +2559,6 @@ void ASTDumper::dumpBoundsKind(BoundsExpr::Kind K) {
     case BoundsExpr::Kind::ByteCount: OS << " Byte"; break;
     case BoundsExpr::Kind::Range: OS << " Range"; break;
     case BoundsExpr::Kind::InteropTypeAnnotation: OS << " InteropTypeAnnotation"; break;
-    default: OS << " <<err>>"; break;
   }
 }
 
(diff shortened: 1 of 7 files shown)


lenary:
According to the C++ book you gave me, Section 17.2.3, destructors first run the body, then invoke member destructors, then invoke base class destructors. This should mean that even with an empty body, the virtual destructors in `ProgramInfo.h` destroy the class' members implicitly.

lenary:
Ok, then we should save it somewhere. Currently it is unused.

lenary:
I'm punting on saving `CheckedLoc` somewhere, because I don't want to deal with AST (De)Serialization. 

I also found two more warnings which I corrected in this commit.
- `ParseExpr.cpp` removed a `default` from a case that won't be hit. This means later we'll get warnings here if we ever add to that Enum.
- `SemaInit.cpp` added a case for an unhandled entity kind.

There's one more warning that I found in the parser, but it's in something i think Jay is actively working on so I'm not going to touch that code. It's the following:
```
/path/to/checkedc-llvm/tools/clang/lib/Parse/ParseDecl.cpp:7128:13: warning: unused variable 'genericTypeDecl' [-Wunused-variable]
      Decl* genericTypeDecl = Actions.ActOnDeclarator(getCurScope(), D);
```

PR #335, opened 2017-06-29
Create lexicographic ordering of C expressions (part 1)

dtarditi:
This implements a lexicographic ordering of C expressions and switches comparison of bounds expressions to use it.  The lexicographic ordering provides a total order on C expressions.

We order by expression kind first  We then order by properties of expressions after that.  For expressions that have an identical kind and properties, we use the lexicographic order of their subexpressions.
-  The ordering is modulo parameter names and typedefs.  For parameters, we use the positional index of the parameter and its type in the ordering, and ignore the names.  For typedef type names, we use the canonical type name, where typedef'ed have been replaced with their underlying type.
- For non-parameter variables, we use their name, type, kind of scope, and order of declaration within a scope for the order.
- For constants, we use the value and type of the constant.

Some aspects of the lexicographic order have yet to be implemented.  The main one that is missing is a lexicographic order on types.   We determine equality/inequality properly for types by deferring to the clang type system APIs for that, but do not order types properly.   There are a few operators that have not been implemented - these are marked with TODOs.

This means that the lexicographic ordering can only be used reliably for equality/inequality of expressions at this point.

Testing:
- Adding a new test of lexical expression equality/inequality indirectly using redeclarations of bounds expressions to the Checked C repo.
- Passed existing Checked C and clang tests.
- Passed LNT test automation for Checked C, which includes benchmarks modified to use Checked C.

### DIFF
--- a/lib/AST/CanonBounds.cpp
+++ b/lib/AST/CanonBounds.cpp
@@ -1,0 +1,250 @@
+//===- CanonBounds.cpp: comparison and canonicalization for bounds exprs -===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+//
+//  This file implements methods to canoncialize expressions used in bounds
+//  expressions and determine if they should be considered semantically
+//  equivalent.
+//
+//===----------------------------------------------------------------------===//
+
+#include "clang/AST/CanonBounds.h"
+#include "clang/AST/Expr.h"
+#include "clang/AST/Stmt.h"
+
+using namespace clang;
+
+LexicographicCompare::Result LexicographicCompare::CompareInteger(signed I1, signed I2) {
+  if (I1 < I2)
+    return Result::LessThan;
+  else if (I1 > I2)
+    return Result::GreaterThan;
+  else
+    return Result::Equal;
+}
+
+LexicographicCompare::Result LexicographicCompare::CompareInteger(unsigned I1, unsigned I2) {
+  if (I1 < I2)
+    return Result::LessThan;
+  else if (I1 > I2)
+    return Result::GreaterThan;
+  else
+    return Result::Equal;
+}
+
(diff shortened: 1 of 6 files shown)


lib/AST/CanonBounds.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+   if (I1 != E1->child_end() && I2 == E2->child_end())
+     return Result::GreaterThan;
+
+   return Lexicographic::Result::Equal;
+}
+
+// Lexicographic comparion of properties specific to each expression type.
+// Does not check children of expressions.
+#define CHECK_WRAPPER(N) \
```
lenary: This is definitely going to catch someone out, that this opens but does not close a scope. I don't know how to do it better though. Maybe take 2 arguments, the second one being a block, like so:

```C
#define CHECK_WRAPPER(N, Block) \
Lexicographic::Result \
 Lexicographic::Compare##N(const Expr *Raw1, const Expr *Raw2) { \
   const N *E1 = dyn_cast<N>(Raw1); \
   const N *E2 = dyn_cast<N>(Raw2); \
   if (!E1 || !E2)  { \
     llvm_unreachable("dyn_cast failed"); \
     return Result::LessThan; \
   }
   Block
}

// Then used like so:
CHECK_WRAPPER(DeclRefExpr, return CompareDecl(E1->getDecl(), E2->getDecl());)

// Or:
CHECK_WRAPPER(IntegerLiteral, {
  BuiltinType::Kind Kind1 = E1->getType()->castAs<BuiltinType>()->getKind();
  BuiltinType::Kind Kind2 = E2->getType()->castAs<BuiltinType>()->getKind();
  Result Cmp = CompareInteger(Kind1, Kind2);
  if (Cmp != Result::Equal)
    return Cmp;
  return CompareAPInt(E1->getValue(), E2->getValue());
})
```

dtarditi: Good point - I've replaced this with a template that does this work instead.   I changed the comparison functions from having the type as part of the name to being an overloaded method.  This allows template resolution to rely on overloading instead of having to do something complicated to figure out which function to call.

PR #336, opened 2017-06-29
Refine and Improve Bounds Inference Code

lenary:
There were a few places where the bounds inference algorithm could have been improved.

- I wanted to make our intent clearer with respect to where we intend to return `bounds(any)` and always will, and where we have to return `bounds(any)` because right now we can't fully model the exact bounds, but we know we will be able to infer bounds eventually (`CreateBoundsAllowedButUncomputable()`)
- I also wanted to make the intent clearer around `bounds(none)` which means one of 3 things:
  - the bounds cannot be described statically (`CreateBoundsUnknown()`)
  - the bounds is an explicitly empty range (`CreateBoundsEmpty()`)
  - We hit some state we're not expecting in the inference algorithm: (`CreateBoundsError()`)

- Now using `Expr::isNullPointerConstant` to detect null pointers, which is more fail-safe than hand-rolled implementations of the very same.
- Can now substitute within the bounds on a CallExpr. The substitution happens as long as the argument expressions are non-modifying-expressions. If the resulting bounds expression is a count or byte count bounds expression, we cannot compute the correct value, but we can at least perform substitution.
- Explicitly allowing/disallowing more mathematics expressions:
  - Float expressions now definitely have bounds(none)
  - Expressions which return ints that represent booleans have bounds(none)  (i.e. `e && e`, `e || e`, `e <= e` `e == e`, `!e` etc.).
  - Explicitly making sure `e - p` where p is a ptr with bounds returns bounds(none). Pointer difference gives you an offset, not a ptr-typed thing. You'll add this offset back to a pointer-like thing to get a pointer again, and in that case we'll require the pointer being added to to have bounds.

### DIFF
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -871,7 +871,7 @@ namespace {
             return CreateBoundsEmpty();
 
           if (FunBounds->isInteropTypeAnnotation())
-            return CreateBoundsInferenceError();
+            return CreateBoundsAllowedButUncomputable();
 
           ArrayRef<Expr *> ArgExprs = llvm::makeArrayRef(const_cast<Expr**>(CE->getArgs()),
                                                          CE->getNumArgs());
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -894,32 +894,42 @@ namespace {
             return CreateBoundsInferenceError();
           }
 
-          // Get the function prototype, where the abstract function return bounds are kept.
-          // The callee is always a function pointer.
-          const FunctionProtoType *CalleeTy = dyn_cast<FunctionProtoType>(CE->getCallee()->getType()->getPointeeType());
-          if (!CalleeTy)
-            // K&R functions have no prototype, and we cannot perform inference on them,
-            // so we return bounds(none) for their results.
-            return CreateBoundsUnknown();
+          BoundsExpr *ReturnBounds = nullptr;
+          if (E->getType()->isCheckedPointerPtrType()) {
+            IntegerLiteral *One =
+              CreateIntegerLiteral(llvm::APInt(1, 1, /*isSigned=*/false));
+            ReturnBounds =  new (Context) CountBoundsExpr(BoundsExpr::Kind::ElementCount,
+                                                          One, SourceLocation(),
+                                                          SourceLocation());
+          }
+          else {
+            // Get the function prototype, where the abstract function return bounds are kept.
+            // The callee is always a function pointer.
+            const FunctionProtoType *CalleeTy = dyn_cast<FunctionProtoType>(CE->getCallee()->getType()->getPointeeType());
+            if (!CalleeTy)
+              // K&R functions have no prototype, and we cannot perform inference on them,
+              // so we return bounds(none) for their results.
+              return CreateBoundsUnknown();
 
-          BoundsExpr *FunBounds = const_cast<BoundsExpr *>(CalleeTy->getReturnBounds());
-          if (!FunBounds)
-            // This function has no return bounds
-            return CreateBoundsUnknown();
+            BoundsExpr *FunBounds = const_cast<BoundsExpr *>(CalleeTy->getReturnBounds());
+            if (!FunBounds)
+              // This function has no return bounds
+              return CreateBoundsUnknown();
 
-          if (FunBounds->isInteropTypeAnnotation())
-            return CreateBoundsAllowedButNotComputed();
+            if (FunBounds->isInteropTypeAnnotation())
+              return CreateBoundsAllowedButNotComputed();
 
-          ArrayRef<Expr *> ArgExprs = llvm::makeArrayRef(const_cast<Expr**>(CE->getArgs()),
-                                                         CE->getNumArgs());
+            ArrayRef<Expr *> ArgExprs = llvm::makeArrayRef(const_cast<Expr**>(CE->getArgs()),
+                                                          CE->getNumArgs());
 
-          // Concretize Call Bounds with argument expressions.
-          // We can only do this if the argument expressions are non-modifying
-          BoundsExpr *ReturnBounds = SemaRef.ConcretizeFromFunctionTypeWithArgs(FunBounds, ArgExprs);
-          // If concretization failed, this means we tried to substitute with a non-modifying
-          // expression, which is not allowed by the specification.
-          if (!ReturnBounds)
-            return CreateBoundsInferenceError();
+            // Concretize Call Bounds with argument expressions.
+            // We can only do this if the argument expressions are non-modifying
+            ReturnBounds = SemaRef.ConcretizeFromFunctionTypeWithArgs(FunBounds, ArgExprs);
+            // If concretization failed, this means we tried to substitute with a non-modifying
+            // expression, which is not allowed by the specification.
+            if (!ReturnBounds)
+              return CreateBoundsInferenceError();
+          }
 
           // Currently we cannot yet concretize function bounds of the forms
           // count(e) or byte_count(e) becuase we need a way of referring
@@ -1119,10 +1129,15 @@ namespace {
       // Bounds of the right-hand side of the assignment
       BoundsExpr *RHSBounds = nullptr;
 
-      // Check that the value being assigned has bounds if the
-      // target of the LHS lvalue has bounds.
-      if (LHSType->isCheckedPointerType() ||
+      if (!E->isCompoundAssignmentOp() &&
+          LHSType->isCheckedPointerPtrType() &&
+          RHS->getType()->isCheckedPointerPtrType()) {
+        // ptr<T> to ptr<T> assignment, no obligation to infer any bounds for either side
+      }
+      else if (LHSType->isCheckedPointerType() ||
           LHSType->isIntegerType()) {
+        // Check that the value being assigned has bounds if the
+        // target of the LHS lvalue has bounds.
         LHSTargetBounds = S.InferLValueTargetBounds(LHS);
         if (!LHSTargetBounds->isNone()) {
           if (E->isCompoundAssignmentOp())
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -31,6 +31,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "clang/AST/RecursiveASTVisitor.h"
+#include "llvm/ADT/SmallBitVector.h"
 #include "TreeTransform.h"
 
 using namespace clang;
@@ -155,12 +156,16 @@ namespace {
 
   private:
     const ArrayRef<Expr *> Arguments;
+    // This stores whether we've emitted an error for a particular substitution
+    // so that we don't duplicate error messages.
+    llvm::SmallBitVector ErroredForArgument;
     bool SubstututedModifyingExpression;
 
   public:
     ConcretizeBoundsExprWithArgs(Sema &SemaRef, ArrayRef<Expr *> Args) :
       BaseTransform(SemaRef),
       Arguments(Args),
+      ErroredForArgument(Args.size()),
       SubstututedModifyingExpression(false) { }
 
     bool substututedModifyingExpression() {
@@ -171,13 +176,16 @@ namespace {
       unsigned index = E->getIndex();
       if (index < Arguments.size()) {
         Expr *AE = Arguments[index];
+        bool ShouldReportError = !ErroredForArgument[index];
 
         // We may only substitute if this argument expression is
         // a non-modifying expression.
         if (!SemaRef.CheckIsNonModifyingExpr(AE,
                                              Sema::NonModifiyingExprRequirement::NMER_Bounds_Function_Args,
-                                             /*ReportError=*/true))
+                                             ShouldReportError)) {
           SubstututedModifyingExpression = true;
+          ErroredForArgument.set(index);
+        }
 
         return AE;
       } else {
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -871,7 +871,7 @@ namespace {
             return CreateBoundsEmpty();
 
           if (FunBounds->isInteropTypeAnnotation())
-            return CreateBoundsAllowedButUncomputable();
+            return FunBounds;
 
           ArrayRef<Expr *> ArgExprs = llvm::makeArrayRef(const_cast<Expr**>(CE->getArgs()),
                                                          CE->getNumArgs());
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -169,8 +169,8 @@ namespace {
         // We may only substitute if this argument expression is
         // a non-modifying expression.
         if (!SemaRef.CheckIsNonModifyingExpr(AE,
-                                             Sema::NonModifiyingExprRequirement::NMER_Unknown,
-                                             /*ReportError=*/false))
+                                             Sema::NonModifiyingExprRequirement::NMER_Bounds_Function_Args,
+                                             /*ReportError=*/true))
           return ExprResult();
 
         return AE;
@@ -338,10 +338,19 @@ namespace {
     }
 
     // Currently our inference algorithm has some limitations,
-    // where we cannot compute bounds for things that we want to be able
-    // to compute bounds for.
+    // where we cannot express bounds for things that will have bounds
+    //
+    // This is for the case where we want to allow these today,
+    // but we need to re-visit these places and disallow some instances
+    // when we can accurately calculate these bounds.
     BoundsExpr *CreateBoundsAllowedButUncomputable() {
       return CreateBoundsAny();
+    }
+    // This is for the opposite case, where we want to return bounds(none)
+    // at the moment, but we want to re-visit these parts of inference
+    // and in some cases compute bounds.
+    BoundsExpr *CreateBoundsNotAllowedYet() {
+      return CreateBoundsNone();
     }
 
     BoundsExpr *CreateSingleElementBounds(Expr *LowerBounds) {
@@ -445,7 +454,7 @@ namespace {
                                                SourceLocation());
         }
         case BoundsExpr::Kind::InteropTypeAnnotation:
-          return CreateBoundsAllowedButUncomputable();
+          return CreateBoundsNotAllowedYet();
         default:
           return B;
       }
@@ -533,7 +542,7 @@ namespace {
         if (!SemaRef.CheckIsNonModifyingExpr(ME->getBase(),
                              Sema::NonModifiyingExprRequirement::NMER_Unknown,
                                            /*ReportError=*/false))
-          return CreateBoundsAllowedButUncomputable();
+          return CreateBoundsNotAllowedYet();
 
         if (ME->getType()->isArrayType())
           return ArrayExprBounds(ME);
@@ -623,7 +632,7 @@ namespace {
           if (!SemaRef.CheckIsNonModifyingExpr(MemberBaseExpr,
               Sema::NonModifiyingExprRequirement::NMER_Unknown,
               /*ReportError=*/false))
-            return CreateBoundsAllowedButUncomputable();
+            return CreateBoundsNotAllowedYet();
           B = SemaRef.MakeMemberBoundsConcrete(MemberBaseExpr, M->isArrow(), B);
           if (!B)
             return CreateBoundsInferenceError();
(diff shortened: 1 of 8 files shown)


lib/Sema/SemaBounds.cpp
```diff
@@ -140,7 +140,56 @@ BoundsExpr *Sema::ConcretizeFromFunctionType(BoundsExpr *Expr,
...
+}
+
+BoundsExpr *Sema::ConcretizeFromFunctionTypeWithArgs(BoundsExpr *Bounds, ArrayRef<Expr *> Args) {
+  if (!Bounds)
+    return Bounds;
+
+  BoundsExpr *Result;
+  ExprResult ConcreteBounds = ConcretizeBoundsExprWithArgs(*this, Args).TransformExpr(Bounds);
+  if (ConcreteBounds.isInvalid()) {
+    return nullptr;
```
lenary: No `llvm_unreachable` here (unlike the other similar functions) because this will return nullptr if one of the arguments we need to substitute is a modifying expression.

lib/Sema/SemaBounds.cpp
```diff
@@ -369,7 +445,7 @@ namespace {
                                                SourceLocation());
         }
         case BoundsExpr::Kind::InteropTypeAnnotation:
-          return CreateBoundsNone();
+          return CreateBoundsAllowedButUncomputable();
```
lenary: This is somewhere where we used to compute `bounds(none)` and now compute `bounds(any)`, because we want to allow this, but cannot express these bounds at the moment.

lib/Sema/SemaBounds.cpp
```diff
@@ -453,49 +529,57 @@ namespace {
         MemberExpr *ME = dyn_cast<MemberExpr>(E);
         FieldDecl *FD = dyn_cast<FieldDecl>(ME->getMemberDecl());
         if (!FD)
-          return CreateBoundsNone();
+          return CreateBoundsInferenceError();
         if (!SemaRef.CheckIsNonModifyingExpr(ME->getBase(),
                              Sema::NonModifiyingExprRequirement::NMER_Unknown,
                                            /*ReportError=*/false))
-          return CreateBoundsNone();
+          return CreateBoundsAllowedButUncomputable();
```
lenary: Here is another place we now return `bounds(any)` instead of `bounds(none)`, because we want to allow these, but can't express the bounds correctly because if we did codegen again, we'd emit the modifying expression twice.

lib/Sema/SemaBounds.cpp
```diff
@@ -524,25 +608,25 @@ namespace {
...
           if (!B || B->isNone() || B->isInteropTypeAnnotation())
-            return CreateBoundsNone();
+            return CreateBoundsInferenceError();
 
           Expr *MemberBaseExpr = M->getBase();
           if (!SemaRef.CheckIsNonModifyingExpr(MemberBaseExpr,
               Sema::NonModifiyingExprRequirement::NMER_Unknown,
               /*ReportError=*/false))
-            return CreateBoundsNone();
+            return CreateBoundsAllowedButUncomputable();
```
lenary: Again, another place where we now return `bounds(any)` instead of `bounds(none)` because we want to allow these expressions eventually, but can't express them in checked c today.

lib/Sema/SemaBounds.cpp
```diff
@@ -578,43 +662,29 @@ namespace {
...
-          DeclRefExpr *DR = dyn_cast<DeclRefExpr>(E);
-          if (!DR) {
-            llvm_unreachable("unexpected cast failure");
-            return CreateBoundsNone();
-          }
-          EnumConstantDecl *ECD = dyn_cast<EnumConstantDecl>(DR->getDecl());
-          if (ECD->getInitVal() == 0)
-            return CreateBoundsAny();
 
-          return CreateBoundsNone();
```
lenary: The two cases that return `bounds(none)` when not a null pointer expression are now caught by the `default` at the end of the block.

lib/Sema/SemaBounds.cpp
```diff
@@ -628,93 +698,210 @@ namespace {
...
+          // `e2` is an RValue
+          //
+          // @ can stand for: +, -, *, /, %, &, |, ^, >>, <<
           if (LHS->getType()->isIntegerType() &&
               RHS->getType()->isIntegerType() &&
               (BinaryOperator::isAdditiveOp(Op) ||
                BinaryOperator::isMultiplicativeOp(Op) ||
-               BinaryOperator::isBitwiseOp(Op))) {
+               BinaryOperator::isBitwiseOp(Op) ||
+               BinaryOperator::isShiftOp(Op))) {
```
lenary: Added shift operators here, because they're kinda bitwise, kinda multiplication, and it seems right to treat them like the other mathematical operators, rather than like a logical operator.

lib/Sema/SemaBounds.cpp
```diff
@@ -628,93 +698,210 @@ namespace {
...
             if (LHSBounds->isNone() && !RHSBounds->isNone())
               return RHSBounds;
             if (!LHSBounds->isNone() && RHSBounds->isNone())
               return LHSBounds;
-            // TODO: check if both LHS and RHS have bounds and
-            // indicate cause of failure to have bounds.
+            if (!LHSBounds->isNone() && !RHSBounds->isNone()) {
+              // TODO: Check if LHSBounds and RHSBounds are equal.
+              // if so, return one of them. If not, return bounds(none)
+              return CreateBoundsUnknown();
```
lenary: This is `bounds(none)` for the moment, but we will eventually want it to be the bounds if both the LHS and RHS have the same bounds. I almost wanted it to be `bounds(any)` but that seems too wide.

lib/Sema/SemaBounds.cpp
```diff
@@ -628,93 +698,210 @@ namespace {
...
+        }
+        case Expr::CallExprClass: {
+          const CallExpr *CE = dyn_cast<CallExpr>(E);
+          if (!CE) {
+            llvm_unreachable("unexpected cast failure");
+            return CreateBoundsInferenceError();
+          }
+
+          // Get function prototype.
+          // The Callee is a function *pointer*, almost always.
```
lenary: I should revisit this comment, the callee is always a function pointer, hence needing to `getPointeeType()`

lib/Sema/SemaBounds.cpp
```diff
@@ -628,93 +698,210 @@ namespace {
...
+                                                         CE->getNumArgs());
+
+          // Concretize Call Bounds with argument expressions.
+          // We can only do this if the argument expressions are non-modifying
+          BoundsExpr *ReturnBounds = SemaRef.ConcretizeFromFunctionTypeWithArgs(FunBounds, ArgExprs);
+
+          // If concretization failed, we can't yet compute static bounds
+          // for the return type of this function.
+          if (!ReturnBounds)
+            return CreateBoundsAllowedButUncomputable();
```
lenary: Concretization will fail if any of the argument expressions that we need to substitute with are modifying expressions. We will eventually be able to describe these bounds, just not yet.

lib/Sema/SemaBounds.cpp
```diff
@@ -628,93 +698,210 @@ namespace {
...
+
+          // Currently we cannot yet concretize function bounds of the forms
+          // count(e) or byte_count(e) becuase we need a way of referring
+          // to the function's return value which we currently lack in the
+          // general case.
+          if (ReturnBounds->isElementCount() ||
+              ReturnBounds->isByteCount())
+            return CreateBoundsAllowedButUncomputable();
```
lenary: We also can't compute the right bounds at the moment if they're `count(n)` or `byte_count(n)` without being able to expand these to use `return_result` or similar.

lenary:
Oh I still need to write far more tests, especially around bounds on function calls.

lib/Sema/SemaBounds.cpp
```diff
@@ -140,7 +140,56 @@ BoundsExpr *Sema::ConcretizeFromFunctionType(BoundsExpr *Expr,
...
+    ExprResult TransformPositionalParameterExpr(PositionalParameterExpr *E) {
+      unsigned index = E->getIndex();
+      if (index < Arguments.size()) {
+        Expr *AE = Arguments[index];
+
+        // We may only substitute if this argument expression is
+        // a non-modifying expression.
+        if (!SemaRef.CheckIsNonModifyingExpr(AE,
+                                             Sema::NonModifiyingExprRequirement::NMER_Bounds_Function_Args,
+                                             /*ReportError=*/true))
```
lenary: This will add an error if we try to substitute an argument that's a modifying expression

lib/Sema/SemaBounds.cpp
```diff
@@ -628,93 +726,209 @@ namespace {
...
+            // so we return bounds(none) for their results.
+            return CreateBoundsEmpty();
+
+          BoundsExpr *FunBounds = const_cast<BoundsExpr *>(CalleeTy->getReturnBounds());
+          if (!FunBounds)
+            // This function has no return bounds
+            return CreateBoundsEmpty();
+
+          if (FunBounds->isInteropTypeAnnotation())
+            return FunBounds;
```
dtarditi: I think this falls into the category of "not yet computed y et".

lenary:
> I am not sure about your uses of CreateBoundsEmpty. To me, I distinguish between an empty range and an unknown range. The empty range does not cause checking of bounds declarations to fail, but would provoke a runtime failure.

I view some things as having a semantically "empty" bounds, which says "this value can never point to any range of memory", for instance float expressions, or `ptr1 - ptr2`. This, I feel, is semantically different to "we don't know (can't statically express) the bounds on this value" or "with particular dynamic values, this dynamic check should fail, but sometimes it can pass". I want "CreateBoundsEmpty()" to mean "we know this range is empty, there is no way it cannot be empty, and statically knowing this is useful".

> I think CreateBoundsInferenceError is conflating internal implementation errors and programmer (source-level) errors. I am confused about what it is supposed to do.

I agree. I'm going to try to only use it for internal implementation errors, such as after `llvm_unreachable` and 

> This change is extensive enough that I'd like you to do automation runs before committing it. We want to make sure that we haven't broken any LNT tests, in particular. This means creating a branch in the GitHub repo for testing.

Sure

> You need to write additional tests for the new inference cases. Given the extent of this change, I'm fine with those being committed separately. Please be sure to open a new issue on the Checked C repo to track that.

Still in progress. I have more commits to push containing these. I want to address your above comments too.

lib/Sema/SemaBounds.cpp
```diff
@@ -524,25 +636,25 @@ namespace {
...
           Expr *MemberBaseExpr = M->getBase();
           if (!SemaRef.CheckIsNonModifyingExpr(MemberBaseExpr,
               Sema::NonModifiyingExprRequirement::NMER_Unknown,
               /*ReportError=*/false))
-            return CreateBoundsNone();
+            return CreateBoundsNotAllowedYet();
           B = SemaRef.MakeMemberBoundsConcrete(MemberBaseExpr, M->isArrow(), B);
           if (!B)
-            return CreateBoundsNone();
+            return CreateBoundsInferenceError();
```
lenary: Here it feels like if we failed to make the bounds concrete, then it's our fault, rather than the programmer's. I also just took a look at `MakeMemberBoundsConcrete` and I think it has a logic bug.

lib/Sema/SemaBounds.cpp
```diff
@@ -524,25 +636,25 @@ namespace {
...
 
           FieldDecl *F = dyn_cast<FieldDecl>(M->getMemberDecl());
           if (!F)
-            return CreateBoundsNone();
+            return CreateBoundsInferenceError();
 
           BoundsExpr *B = F->getBoundsExpr();
           if (!B || B->isNone() || B->isInteropTypeAnnotation())
-            return CreateBoundsNone();
+            return CreateBoundsInferenceError();
```
lenary: Then the first two are `CreateBoundsUnknown()` and the third is `CreateBoundsAllowedButNotComputed()` right?

lib/Sema/SemaBounds.cpp
```diff
@@ -262,12 +330,48 @@ namespace {
...
                                              SourceLocation());
     }
 
+    // Currently our inference algorithm has some limitations,
+    // where we cannot express bounds for things that will have bounds
+    //
+    // This is for the case where we want to allow these today,
+    // but we need to re-visit these places and disallow some instances
+    // when we can accurately calculate these bounds.
+    BoundsExpr *CreateBoundsAllowedButUncomputable() {
```
lenary: I'm changing this to `CreateBoundsAllowedButNotComputed()` - the original intent was to go for something along the lines of "this is not expressible/computable" but I'm happy with the above

lib/Sema/SemaBounds.cpp
```diff
@@ -628,93 +726,209 @@ namespace {
...
           if (BinaryOperator::isCompoundAssignmentOp(Op)) {
             Op = BinaryOperator::getOpForCompoundAssignment(Op);
             IsCompoundAssignment = true;
           }
 
           // Pointer arithmetic.
+          //
+          // `p + i` has the bounds of `p`. `p` is an RValue.
+          // `p += i` has the lvalue target bounds of `p`. `p` is an LValue.
+          // same applies for `-` and `-=` respectively
```
lenary: :+1: fixed

lib/Sema/SemaBounds.cpp
```diff
@@ -628,93 +726,209 @@ namespace {
...
+        case Expr::CallExprClass: {
+          const CallExpr *CE = dyn_cast<CallExpr>(E);
+          if (!CE) {
+            llvm_unreachable("unexpected cast failure");
+            return CreateBoundsInferenceError();
+          }
+
+          // Get function prototype.
+          // The Callee is a function *pointer*, almost always.
+          const FunctionProtoType* CalleeTy = dyn_cast<FunctionProtoType>(CE->getCallee()->getType()->getPointeeType());
```
lenary: :+1: fixed.

lib/Sema/SemaBounds.cpp
```diff
@@ -628,93 +726,209 @@ namespace {
...
+          if (!CalleeTy)
+            // K&R functions have no prototype, and we cannot perform inference on them,
+            // so we return bounds(none) for their results.
+            return CreateBoundsEmpty();
+
+          BoundsExpr *FunBounds = const_cast<BoundsExpr *>(CalleeTy->getReturnBounds());
+          if (!FunBounds)
+            // This function has no return bounds
+            return CreateBoundsEmpty();
```
lenary: :+1: fixed

lib/Sema/SemaBounds.cpp
```diff
@@ -628,93 +726,209 @@ namespace {
...
+            return CreateBoundsInferenceError();
+          }
+
+          // Get function prototype.
+          // The Callee is a function *pointer*, almost always.
+          const FunctionProtoType* CalleeTy = dyn_cast<FunctionProtoType>(CE->getCallee()->getType()->getPointeeType());
+          if (!CalleeTy)
+            // K&R functions have no prototype, and we cannot perform inference on them,
+            // so we return bounds(none) for their results.
+            return CreateBoundsEmpty();
```
lenary: :+1: fixed.

lib/Sema/SemaBounds.cpp
lenary: From the other `->isInteropTypeAnnotation()` if-statement, I attempted to make this some equivalent to `bounds(none)` for conservativeness, but it was failing existing tests, so I went back to returning the interop type. 

I've now changed everywhere we come across an interop type annotation to return `CreateBoundsAllowedButNotComputed()`, including in `BoundsExpr *ExpandToRange(Expr *Base, BoundsExpr *B)`

dtarditi:
It would be helpful to add the explanatory comment to CreateBoundsEmpty.   That's a different notion of empty range than what I was thinking about.

lib/Sema/SemaBounds.cpp
dtarditi: Yes.

test/CheckedC/inferred-bounds/member-base.c
```diff
@@ -145,35 +145,6 @@ void f3(_Array_ptr<struct S> a : bounds(a, a + 7)) {
 // CHECK: `-ImplicitCastExpr {{0x[0-9a-f]+}} '_Array_ptr<struct S>' <LValueToRValue>
 // CHECK: `-DeclRefExpr {{0x[0-9a-f]+}} '_Array_ptr<struct S>' lvalue ParmVar {{0x[0-9a-f]+}} 'a' '_Array_ptr<struct S>'
 
-// CHECK: BinaryOperator {{0x[0-9a-f]+}} '_Ptr<int>' '='
```
lenary: I should probably replace this test, as there's still an Implicit Cast going on in this code which needs testing, even though it is not tested in the version of the test I've removed these lines from.

lib/Sema/SemaBounds.cpp
```diff
@@ -896,10 +1129,15 @@ namespace {
       // Bounds of the right-hand side of the assignment
       BoundsExpr *RHSBounds = nullptr;
 
-      // Check that the value being assigned has bounds if the
-      // target of the LHS lvalue has bounds.
-      if (LHSType->isCheckedPointerType() ||
+      if (!E->isCompoundAssignmentOp() &&
+          LHSType->isCheckedPointerPtrType() &&
+          RHS->getType()->isCheckedPointerPtrType()) {
+        // ptr<T> to ptr<T> assignment, no obligation to infer any bounds for either side
```
lenary: This code is new - it stops requiring bounds for both sides if it is a `ptr<T>` to `ptr<T>` assignment.

dtarditi:
The changes look good - thanks!

You explained to me that you removed a bunch of inferred bounds for member expressions because the check for Ptr to Ptr assignments means that no bounds inference happens there.  Could you make sure that we still have good coverage of various right-hand sides of expressions being assigned to member expressions?

test/CheckedC/inferred-bounds/calls.c
```diff
@@ -0,0 +1,445 @@
...
+// RUN: %clang_cc1 -fcheckedc-extension -verify -fdump-inferred-bounds %s | FileCheck %s
+
+// In all these functions, though they take two arguments, the bounds
+// are only expressed in terms of the first of the two.
+// This means we can test that substitution is successful even if
+// a non-modifying-expression is used for the second argument.
+
+extern _Array_ptr<int> f_bounds(_Array_ptr<int> start, _Array_ptr<int> end) : bounds(start, start + 1);
+extern _Array_ptr<int> f_count(int i, int j) : count(i);
+extern _Array_ptr<int> f_byte(int i, int j) : count(i);
```
lenary: Yes

test/CheckedC/inferred-bounds/calls.c
```diff
@@ -0,0 +1,445 @@
...
+
+extern int* f_boundsi(int* start, int* end) : bounds(start, start + 1);
+extern int* f_counti(int i, int j) : count(i);
+extern int* f_bytei(int i, int j) : count(i);
```
lenary: Yes

test/CheckedC/inferred-bounds/member-base.c
```diff
@@ -145,22 +145,25 @@ void f3(_Array_ptr<struct S> a : bounds(a, a + 7)) {
...
 // CHECK: `-ImplicitCastExpr {{0x[0-9a-f]+}} '_Array_ptr<struct S>' <LValueToRValue>
 // CHECK:           `-DeclRefExpr {{0x[0-9a-f]+}} '_Array_ptr<struct S>' lvalue ParmVar {{0x[0-9a-f]+}} 'a' '_Array_ptr<struct S>'
 // CHECK: Target Bounds:
 // CHECK: RangeBoundsExpr {{0x[0-9a-f]+}} 'NULL TYPE'
-// CHECK: |-ImplicitCastExpr {{0x[0-9a-f]+}} '_Ptr<int>' <LValueToRValue>
-// CHECK: | `-DeclRefExpr {{0x[0-9a-f]+}} '_Ptr<int>' lvalue Var {{0x[0-9a-f]+}} 'p' '_Ptr<int>'
-// CHECK: `-BinaryOperator {{0x[0-9a-f]+}} '_Ptr<int>' '+'
-// CHECK: |-ImplicitCastExpr {{0x[0-9a-f]+}} '_Ptr<int>' <LValueToRValue>
-// CHECK: | `-DeclRefExpr {{0x[0-9a-f]+}} '_Ptr<int>' lvalue Var {{0x[0-9a-f]+}} 'p' '_Ptr<int>'
-// CHECK: `-IntegerLiteral {{0x[0-9a-f]+}} 'unsigned long long' 1
```
lenary: This is the most infuriating diff.

`_Array_ptr<T> p : count(i) = expr`, where `i` is a literal, `i` has type `int`
`_Ptr<T> p = expr`, the `1` literal in the range-expanded bounds has type `size_t` i think. We should aim for these to be the same, if we can.

lenary:
I think this is ready for final re-review and merging. 

I'm going to set off a testing run on the DevTest machines.

lenary:
DevTest run finished last night with 4/4 succeeded.

PR #340, opened 2017-07-06
Type variable type.

jp1729:
This pull request has TypeVariableType implementation.
When a generic function is declared, each type variable is assigned a TypeVariableType, and is stored in Typedef. At the same time, each Typedef is stored inside of FunctionDecl to keep track of the name (identifier) and the type (deBruijn indexes).
Function Type is also updated to keep track of the number of type variables.
When invoking -ast-dump, the type variable information is shown as well.
* The Typedef that stores TypeVariableType is shown first.
* Then, the Function declaration is shown with the new function type. The new function type shows the number of type variables declared in the function.
* Next, each TypeVariable will be shown.
* Then the list continues like it was before, starting with the list of parameters.


include/clang/AST/Type.h
```diff
@@ -3689,6 +3693,33 @@ class UnresolvedUsingType : public Type {
 };
 
 
+class TypeVariableType : public Type, public llvm::FoldingSetNode {
+  unsigned int deBruijnDepth;
```
dtarditi: Please look at how parameter variables are being symbolically represented elsewhere in clang.  They have scope depth and positional offset information.  We probably want to treat type variables in a similar fashion.

include/clang/Sema/DeclSpec.h
```diff
@@ -601,6 +606,17 @@ class DeclSpec {
   bool isForanySpecified() const { return FS_forany_specified; }
   SourceLocation getForanySpecLoc() const { return FS_foranyLoc; }
 
+  void setTypeVars(ASTContext &C, ArrayRef<TypedefDecl *> NewTypeVarInfo, unsigned NewNumTypeVars);
+  void setNumTypeVars(unsigned NewNumTypeVars) { NumTypeVars = NewNumTypeVars; }
+  unsigned getNumTypeVars(void) const { return NumTypeVars; }
+
+  ArrayRef<TypedefDecl *> typevariables() const {
+    return{ TypeVarInfo, getNumTypeVars() };
```
dtarditi: Please add a space after the return keyword.

lib/AST/ASTContext.cpp
```diff
@@ -8500,6 +8527,8 @@ QualType ASTContext::mergeTypes(QualType LHS, QualType RHS,
            "Equivalent pipe types should have already been handled!");
     return QualType();
   }
+  case Type::TypeVariable:
```
dtarditi: Could you add a comment to explain what is going on here?    Type variables are compatible only if they are identical.  That's covered by the earlier check that the canonical types are identical.  We've reached this point only if the type variables are not compatible, so we return QualType().

include/clang/AST/Decl.h
```diff
@@ -1642,6 +1643,7 @@ class FunctionDecl : public DeclaratorDecl, public DeclContext,
   /// parameters of this function.  This is null if a prototype or if there are
   /// no formals.
   ParmVarDecl **ParamInfo;
+  TypedefDecl **TypeVarInfo;
```
dtarditi: Please add a comment explaining this.

lib/AST/ASTContext.cpp
```diff
@@ -8059,6 +8080,11 @@ QualType ASTContext::mergeFunctionTypes(QualType lhs, QualType rhs,
       }
     }
 
+    // Compatible functions must have the same number of type variables
+    // FIXME : Make sure to pass correct QualType.
```
dtarditi: I don't understand the FIXME comment.  It doesn't look like there's anything to check.

I suggest moving this check earlier so that is with other similar checks such as checks on the number of parameters.

lib/Parse/ParseDecl.cpp
```diff
@@ -7096,36 +7109,21 @@ void Parser::ParseForanySpecifier(DeclSpec &DS) {
...
-                                                genericType.getVolatileSpecLoc(),
-                                                genericType.getRestrictSpecLoc(),
-                                                genericType.getAtomicSpecLoc(),
-                                                genericType.getUnalignedSpecLoc()),
-                                                genericType.getAttributes(),
-                                                SourceLocation());
-      D.SetIdentifier(Tok.getIdentifierInfo(), Tok.getLocation());
-      D.SetRangeEnd(Tok.getLocation());
-      Decl* genericTypeDecl = Actions.ActOnDeclarator(getCurScope(), D);
+      // the binding to type variable.
```
dtarditi: I think the comment needs to be updated.

lib/Sema/SemaInit.cpp
```diff
@@ -7093,6 +7093,15 @@ InitializationSequence::Perform(Sema &S,
       CurInit = CurInitExprRes;
 
       bool Complained;
+
+      // FIXME : This code is inserted only as a hotfix for type variable
```
dtarditi: I think the comment should be different.    This is a workaround because generic type instantiation has not been implemented.  That leaves type variables in unexpected places.    You've actually implemented compatibility for type variables in the current change.  Also, could you use the word TODO instead of FIXME.   FIXME implies that this is a bug, whereas TODO means unimplemented functionality.

lib/Sema/TreeTransform.h
```diff
@@ -5434,6 +5434,12 @@ QualType TreeTransform<Derived>::TransformTypeOfExprType(TypeLocBuilder &TLB,
 }
 
 template<typename Derived>
+QualType TreeTransform<Derived>::TransformTypeVariableType(TypeLocBuilder &TLB,
+                                                           TypeVariableTypeLoc TL) {
+  return QualType();
```
dtarditi: Could yo fill this in, following the pattern for other types?    As far as I can tell, this should just return the underlying type variable.

lib/Serialization/ASTWriter.cpp
```diff
@@ -334,6 +334,13 @@ void ASTTypeWriter::VisitTypedefType(const TypedefType *T) {
   Code = TYPE_TYPEDEF;
 }
 
+void ASTTypeWriter::VisitTypeVariableType(const TypeVariableType *T) {
```
dtarditi: There should be matching code in ASTReader.cpp for reading the type variable in.    I think the right function for implementing the functionality is readTypeRecord.

lib/AST/ASTContext.cpp
jp1729: The FIXME is wrong. I must have mistaken it for being the code that is doing compatibility check. The FIXME should be removed.

jp1729:
It is ready for further review.

include/clang/AST/ASTContext.h
```diff
@@ -1182,6 +1183,11 @@ class ASTContext : public RefCountedBase<ASTContext> {
   /// pointer to blocks.
   QualType getBlockDescriptorExtendedType() const;
 
+  /// Returns a type variable type based on the depth of the for any scope
```
dtarditi: Please put quotes around 'for any' or use _For_any in comments.

include/clang/AST/Type.h
```diff
@@ -3689,6 +3693,35 @@ class UnresolvedUsingType : public Type {
 };
 
 
+class TypeVariableType : public Type, public llvm::FoldingSetNode {
```
dtarditi: Please add a classof method to this class.   That is necessary for Clang's/LLVM hand-implemented dynamic type tests to work properly.

lib/AST/ASTContext.cpp
dtarditi: The assert is even better.  Thanks!

lib/Parse/ParseDecl.cpp
```diff
@@ -7082,8 +7082,21 @@ void Parser::ParseForanySpecifier(DeclSpec &DS) {
     return;
   }
   tok::TokenKind prevToken = tok::l_paren;
-  // Obtain list of type variables bound to for any.
+
+  SmallVector<TypedefDecl*, 16> typevars;
   bool breakOutOfWhileLoop = false;
+  unsigned int typeVariableIndex = 0;
+
+  // Calculate the depth of for any scope for our generic types
```
dtarditi: Please put quotes around for any to make this easier to parse: 'for any'

include/clang/AST/Type.h
jp1729: There is classof method in line 3709! 
```
static bool classof(const Type *T) { return T->getTypeClass() == TypeVariable; }
```

dtarditi: Sorry I missed that.  Could you move it to be after the Profile method, so that it matches the order in most of the other definitions of classes for types?

jp1729: That sounds good!

PR #345, opened 2017-07-17
Initialize Types when calling a generic function

jp1729:
When calling a generic function, initialize the type of the function, as well as save information about the type and its source location in DeclRefExpr. The type of DeclRefExpr will be the completely initialized type. Any parent node will keep the initialized type.


include/clang/AST/Expr.h
```diff
@@ -955,9 +955,43 @@ class DeclRefExpr final
       private llvm::TrailingObjects<DeclRefExpr, NestedNameSpecifierLoc,
                                     NamedDecl *, ASTTemplateKWAndArgsInfo,
                                     TemplateArgumentLoc> {
+public :
+  /// \brief extra information for DeclRefExprInfo.
+  class GenericFunctionCallInfo {
```
dtarditi: I suggest renaming this to GenericInstInfo.  Please add a comment saying that this is information for an instantiation of a generic function.   I suggested renaming TypeNameInfo to TypeArgument.

include/clang/AST/Expr.h
```diff
@@ -955,9 +955,43 @@ class DeclRefExpr final
...
+  public:
+    struct TypeNameInfo {
+      QualType typeName;
+      TypeSourceInfo *sourceInfo;
+    };
+
+  private :
+    /// \brief If DeclRefExpr references generic function, it must store
```
dtarditi: I suggest rewording the comment to be more precise: references to generic functions in code must always be instantiated (applied to type arguments)  We store the type arguments in DeclRefExpr.

Instead of calling the member TypeNameInfos, I suggest TypeArguments.

include/clang/AST/Expr.h
```diff
@@ -955,9 +955,43 @@ class DeclRefExpr final
...
+    MutableArrayRef<TypeNameInfo> typeNameInfos() {
+      return{ TypeNameInfos, NumTypeNameInfo };
+    }
+  };
+
+private :
   /// \brief The declaration that we are referencing.
   ValueDecl *D;
 
+  GenericFunctionCallInfo *GenFuncInfo;
```
dtarditi: I suggest renaming this to something like TypeArguments.   Please add a comment describing this field.

lib/Parse/ParseExpr.cpp
```diff
@@ -3023,6 +3022,74 @@ ExprResult Parser::ParseBoundsExpression() {
...
 }
 
+// Recurse through types and substitute any TypedefType with TypeVariableType
+// as the underlying type.
+QualType Parser::SubstituteTypeVariable(QualType QT,
+  SmallVector<DeclRefExpr::GenericFunctionCallInfo::TypeNameInfo, 16> &typeNames) {
+  const Type *T = QT.getTypePtr();
+  switch (T->getTypeClass()) {
+  case Type::Pointer: {
+    const PointerType *pType = dyn_cast<PointerType>(T);
```
dtarditi: We try to practice "defensive programming" in our changes in clang, more so than usually seen. After a dyn_cast, we usually check that the result is non-null as expected.   If it is null, we use llvm_unreachable and return some sort of sensible type.   The other uses of dyn_cast in this function should be updated too.

lib/Parse/ParseExpr.cpp
```diff
@@ -3023,6 +3022,74 @@ ExprResult Parser::ParseBoundsExpression() {
...
+    // If underlying type is TypeVariableType, must replace the entire
+    // TypedefType. If underlying type is not TypeVariableType, there cannot be
+    // a type variable type that this context can replace.
+    const TypedefType *tdType = dyn_cast<TypedefType>(T);
+    const Type *underlying = tdType->getDecl()->getUnderlyingType().getTypePtr();
+    if (const TypeVariableType *tvType = dyn_cast<TypeVariableType>(underlying))
+      return typeNames[tvType->GetIndex()].typeName;
+    else
+      return QT;
+  }
```
dtarditi: It is possible for various transformations in clang to eliminate the typedef.  Please add a case for type variable by itself.    Also, please add a case for function types.  Function types could use a type variable or a type that contains a type variable.

test/CheckedC/generic-function/generic-func-call-info.c
```diff
@@ -0,0 +1,29 @@
+// Test to make sure that the generic function calls are correctly instantiated
+//
+// In CallExpr, the type of the DeclRefExpr should be completely instantiated,
+// by replacing all type variable types with specified type names in the list
+// of type names provided when calling generic function.
+//
+//
+// RUN: %clang_cc1 -ast-dump -fcheckedc-extension %s | FileCheck %s
```
dtarditi: Please add test cases for the different forms of constructed types:
- Pointers to type variables.
- Function types that have type variables within them.
- Array types that use type variables.

Also please add a test case for the new error message that you have introduced.

jp1729: I added the test case for the new error message in Checkedc repository.

jp1729:
It's ready for further review.

lib/Parse/ParseExpr.cpp
```diff
@@ -3023,6 +3022,142 @@ ExprResult Parser::ParseBoundsExpression() {
...
+    // TypedefType. If underlying type is not TypeVariableType, there cannot be
+    // a type variable type that this context can replace.
+    const TypedefType *tdType = dyn_cast<TypedefType>(T);
+    if (!tdType) {
+      llvm_unreachable("Type cannot be empty when substituting type variable.");
+      return QT;
+    }
+    const Type *underlying = tdType->getDecl()->getUnderlyingType().getTypePtr();
+    if (const TypeVariableType *tvType = dyn_cast<TypeVariableType>(underlying)) {
+      if (!tvType) {
```
dtarditi: You can remove this if-statement: we know tvType will be non-null.

lib/Parse/ParseExpr.cpp
```diff
@@ -3053,6 +3189,7 @@ bool Parser::ParseGenericFunctionExpression(ExprResult &Res) {
     return false;
   }
   else {
+    SmallVector<DeclRefExpr::GenericInstInfo::TypeArgument, 16> typeArgumentInfos;
```
dtarditi: Instead of 16, could you just use 4?   It is unlikely someone will write a function with more than 1 or 2  type parameters.

lib/Serialization/ASTReaderStmt.cpp
```diff
@@ -599,8 +612,8 @@ void ASTStmtReader::VisitUnaryExprOrTypeTraitExpr(UnaryExprOrTypeTraitExpr *E) {
 
 void ASTStmtReader::VisitArraySubscriptExpr(ArraySubscriptExpr *E) {
   VisitExpr(E);
-  E->setLHS(Record.readSubExpr());
-  E->setRHS(Record.readSubExpr());
+  E->setLHS(Record.readSubExpr());
```
dtarditi: This looks like a spurious white-space change.  Could you revert it?

lib/Serialization/ASTReaderStmt.cpp
```diff
@@ -662,7 +675,7 @@ void ASTStmtReader::VisitCastExpr(CastExpr *E) {
   VisitExpr(E);
   unsigned NumBaseSpecs = Record.readInt();
   assert(NumBaseSpecs == E->path_size());
-  E->setSubExpr(Record.readSubExpr());
+  E->setSubExpr(Record.readSubExpr());
```
dtarditi: This looks like a spurious white space change.  Could you revert it?

test/CheckedC/generic-function/generic-func-call-info.c
dtarditi: Could you add tests for array types used as parameters?   There are 2 cases to cover:
- Incomplete array types as parameters:
`_For_any(T) int f(T[] arr)`
- Complete array types as parameters:
`_For_any(T) int f(T[10] arr)`

test/CheckedC/generic-function/generic-func-call-info.c
```diff
@@ -0,0 +1,67 @@
...
+//
+// In CallExpr, the type of the DeclRefExpr should be completely instantiated,
+// by replacing all type variable types with specified type names in the list
+// of type names provided when calling generic function.
+//
+//
+// RUN: %clang_cc1 -ast-dump -fcheckedc-extension %s | FileCheck %s
+
+
+_For_any(T, S, R) T genericTest(T test, S steve, R ray) {
```
dtarditi: This test won't work after you make type variables be incomplete types.  Variables cannot have an incomplete type.

lib/Parse/ParseExpr.cpp
```diff
@@ -3023,6 +3022,142 @@ ExprResult Parser::ParseBoundsExpression() {
...
+  case Type::TypeVariable: {
+    // Although Type Variable Type is wrapped with Typedef Type, there may be
+    // transformations in clang that eliminates Typedef.
+    const TypeVariableType *tvType = dyn_cast<TypeVariableType>(T);
+    if (!tvType) {
+      llvm_unreachable("Type cannot be empty when substituting type variable.");
+      return QT;
+    }
+    return typeNames[tvType->GetIndex()].typeName;
+  }
```
dtarditi: I think you also need to handle DecayedType and AdjustedType.   This will arise when a parameter is declared to have an array type or a function type.

lib/Parse/ParseExpr.cpp
```diff
@@ -3023,6 +3022,142 @@ ExprResult Parser::ParseBoundsExpression() {
...
+// Recurse through types and substitute any TypedefType with TypeVariableType
+// as the underlying type.
+QualType Parser::SubstituteTypeVariable(QualType QT,
+  SmallVector<DeclRefExpr::GenericInstInfo::TypeArgument, 16> &typeNames) {
+  const Type *T = QT.getTypePtr();
+  switch (T->getTypeClass()) {
+  case Type::Pointer: {
+    const PointerType *pType = dyn_cast<PointerType>(T);
+    if (!pType) {
+      llvm_unreachable("Type cannot be empty when substituting type variable.");
```
dtarditi: I think a better unreachable error message would be "dynamic cast failed unexpectedly."

lib/Serialization/ASTReaderStmt.cpp
jp1729: I made sure to save as Unix encoding, but it seems to have no effect.

lib/Serialization/ASTReaderStmt.cpp
jp1729: I made sure to save as Unix encoding, but it seems to have no effect.

jp1729:
It is ready for further review. I wrote the two test cases for array types in line 26 and line 30.

PR #348, opened 2017-07-21
Implement partition refinement data structure.

dtarditi:
Implement a data structure for representing a set integers partitioned into equivalence classes and operations for refining partitions..   Partition refinement will be used to compute variables that hold equivalent values, which in turn will be used during checking of equivalence of bounds expressions.

Testing:
- Implement unit tests for partition refinement, using the Google Test infrastructure supported by clang.
- Existing tests continue to pass, as expected.   The partition refinement data structure is unused by the compiler.


lib/Analysis/VarEquiv.cpp
```diff
@@ -0,0 +1,392 @@
...
+         "add operation makes this no longer a partition");
+  ListNode *Node = new ListNode(Elem, S);
+  NodeMap->set(Elem, Node);
+  linkNode(S, Node);
+  return Node;
+}
+
+// Add Elem to the set S for Member.  If Member does not have
+// a set, create a new set to contain Elem and Member.
+// It is an error of Elem is already a member of another set.
```
jp1729: >It is an error "of" Elem is already a member of another set.

I think "of" should be "if."

lib/Analysis/VarEquiv.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+
+  count = R->Sets->size();
+  for (unsigned i = 0; i < count; i++)
+    refine(R->Sets->get(i));
+}
+
+// Return a representative element from the equivalence set
+// for Elem.
+  Element Partition::getRepresentative(Element Elem) const {
```
jp1729: There's spaces in front of the function declaration. I think you accidentally hit tab when creating the function.

lib/Sema/SemaBounds.cpp
```diff
@@ -1115,7 +1115,7 @@ namespace {
 
   public:
     CheckBoundsDeclarations(Sema &S) : S(S),
-     DumpBounds(S.getLangOpts().DumpInferredBounds) {}
+      DumpBounds(S.getLangOpts().DumpInferredBounds) {}
```
jp1729: I think you accidentally added a space here.

unittests/CheckedC/PartitionRefinementTests.cpp
```diff
@@ -0,0 +1,433 @@
...
+  EXPECT_EQ(P1->getRepresentative(0), P1->getRepresentative(1));
+  EXPECT_NE(P1->getRepresentative(0), P1->getRepresentative(3));
+  EXPECT_FALSE(P1->isSingleton(0));
+  EXPECT_FALSE(P1->isSingleton(1));
+  EXPECT_TRUE(P1->isSingleton(2));
+  EXPECT_TRUE(P1->isSingleton(3));
+  EXPECT_TRUE(P1->isSingleton(4));
+
+  P1->makeSingleton(0);
+  EXPECT_TRUE(P1->isSingleton(0));
```
jp1729: I don't think this line is necessary since line 33 checks for the same thing?

unittests/CheckedC/PartitionRefinementTests.cpp
```diff
@@ -0,0 +1,433 @@
...
+  EXPECT_EQ(P1->getRepresentative(3), P1->getRepresentative(2));
+  EXPECT_EQ(P1->getRepresentative(3), P1->getRepresentative(3));
+  EXPECT_NE(P1->getRepresentative(3), P1->getRepresentative(4));
+
+  EXPECT_NE(P1->getRepresentative(4), P1->getRepresentative(0));
+  EXPECT_EQ(P1->getRepresentative(4), P1->getRepresentative(1));
+  EXPECT_NE(P1->getRepresentative(4), P1->getRepresentative(2));
+  EXPECT_NE(P1->getRepresentative(4), P1->getRepresentative(3));
+  EXPECT_EQ(P1->getRepresentative(4), P1->getRepresentative(4));
+  EXPECT_EQ(P1->getRepresentative(4), P1->getRepresentative(4));
```
jp1729: I think line 126 and line 127 are duplicate and they test for the same thing.

lib/Sema/SemaBounds.cpp
dtarditi: This change is fixing the spacing - it should have been two spaces to start with.

unittests/CheckedC/PartitionRefinementTests.cpp
dtarditi: Yes, you are correct.

unittests/CheckedC/PartitionRefinementTests.cpp
dtarditi: Yes, you are correct.

dtarditi:
Thanks for the review!  I believe I have addressed all the issues you pointed out.

PR #351, opened 2017-07-25
Remove warning signs

jp1729:
Some warnings are introduced while creating Type Variable Types. All of the cases are trivial, as Type Variable Types are not affected by any of the functions.

### DIFF
--- a/lib/CodeGen/CGDebugInfo.cpp
+++ b/lib/CodeGen/CGDebugInfo.cpp
@@ -539,6 +539,11 @@ void CGDebugInfo::CreateCompileUnit() {
       CGM.getCodeGenOpts().DebugInfoForProfiling);
 }
 
+llvm::DIType *CGDebugInfo::CreateType(const TypeVariableType *TvT) {
+  // Since type variable is incomplete type, model it similar to void
+  return nullptr;
+}
+
 llvm::DIType *CGDebugInfo::CreateType(const BuiltinType *BT) {
   llvm::dwarf::TypeKind Encoding;
   StringRef BTName;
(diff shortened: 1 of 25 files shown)


lib/CodeGen/CGDebugInfo.cpp
```diff
@@ -2703,6 +2703,7 @@ llvm::DIType *CGDebugInfo::CreateTypeNode(QualType Ty, llvm::DIFile *Unit) {
   case Type::Decltype:
   case Type::UnaryTransform:
   case Type::PackExpansion:
+  case Type::TypeVariable:
```
dtarditi: I don't think this is the right change for type variables.   Shouldn't it be mapped the same way as void * is mapped?

lib/CodeGen/ItaniumCXXABI.cpp
```diff
@@ -2828,6 +2828,9 @@ void ItaniumRTTIBuilder::BuildVTablePointer(const Type *Ty) {
   case Type::Pipe:
     llvm_unreachable("Pipe types shouldn't get here");
 
+  case Type::TypeVariable:
+    llvm_unreachable("Type Variable shouldn't get here");
```
dtarditi: I think it should be a "not yet implemented message", not "type variable shouldn't get here.".

lib/CodeGen/CGDebugInfo.cpp
jp1729: Should it be mapped the same way as "void" or "void *"? I think it should be mapped same as "void." I'll look more into it and will definitely have to fix it, because otherwise I cannot use clang-cl to build benchmarks. I am currently getting an error building zlib when I use type variables in the code, and this seems to be the cause of the problem.

dtarditi: I think it is fine to treat it the same way as void for now.

jp1729:
I have made changes and it is ready for further review.

Issue #355, opened 2017-08-03
Allow old-style K&R functions with prototypes in checked blocks


dtarditi:
I added a check that K&R-style function definitions are not allowed in checked block.  There's a special case that we need to allow, which is when the definition is preceded by a function prototype.   We're doing the check too early during processing of a function definition.

dtarditi:
This work is done.

PR #357, opened 2017-08-05
Allow K&R style function definitions preceded by prototypes in a checked scope.

dtarditi:
We disallow function declarators that have no prototypes in checked scopes because calls to them may not be typechecked.  We were being too strict in the checking and  just looking at whether a function declarator provided a prototype. 

We need to take into account that according to the C11 specification, type information from prior declarations of a function is merged with a function declarator to determine whether a function has a  prototype.    This change moves checking to later in ActOnFunctionDeclarator after type merging, which addresses issue #355.  If a no-prototype function declarator is preceded by a prototype, the declarator is considered to have a prototype.  

However, we need to add some special-case handling for the case of K&R style function definitions. 
This is where the function declarator for a function definition is followed by a list of declarations for parameters, for example:
```
int f(a, b)
int a;
int b;
{ }
```
clang assigns a prototype function type to this definition, even though the C11 standard does not say that such a definition creates a prototype for a function.  We could have issues with other C compilers in the future, which may not type check succeeding uses of the function. We specifically check that there is a prior declaration with a prototype.  There may be an easier way in clang to determine that a K&R function definition has a prototype in a standard-compliant way (because of a prior declaration),  but I couldn't find it.

Even though K&R-style function definition are obsolete, we care about this because we are trying to minimize code changes required to use Checked C.  We don't want to force someone to modify a K&R-style function definition when they have done their homework and provided a prototype.

This change also includes some minor clean up and improvements:
- We now record whether a function declarator has a checked/unchecked specifier.
- There were some identifiers that started with lowercase letters  that needed to start with uppercase letters (and vice versa) to match the coding convention.
- Make some comments more accurate.

Testing:
- I added a new test case to the Checked C tests for a K&R function definition preceded by a prototype.  This will be covered by a separate pull request for the Checked C repo.

### DIFF
--- a/lib/Sema/SemaDecl.cpp
+++ b/lib/Sema/SemaDecl.cpp
@@ -8324,7 +8324,13 @@ Sema::ActOnFunctionDeclarator(Scope *S, Declarator &D, DeclContext *DC,
       Diag(NewFD->getLocation(), diag::no_prototype_generic_function);
       NewFD->setInvalidDecl();
     }
+
   }
+
+  if (D.getDeclSpec().isCheckedSpecified())
+    NewFD->setCheckedSpecifier(CheckedFunctionSpecifiers::CFS_Checked);
+  else if (D.getDeclSpec().isUncheckedSpecified())
+    NewFD->setCheckedSpecifier(CheckedFunctionSpecifiers::CFS_UnChecked);
 
   if (OriginalLexicalContext && OriginalLexicalContext->isObjCContainer())
     NewFD->setTopLevelDeclInObjCContainer();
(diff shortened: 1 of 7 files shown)


lib/Sema/SemaDecl.cpp
```diff
@@ -12786,6 +12816,8 @@ Decl *Sema::ActOnFinishFunctionBody(Decl *dcl, Stmt *Body,
   if (FD) {
     FD->setBody(Body);
 
+
```
lenary: This whitespace change seems unneeded.

PR #358, opened 2017-08-09
Extend -Wcheck-bounds-decls to variable declarations with initializers

dtarditi:
Checking the correctness of bounds declarations is currently under a warning flag that is off by default (-Wcheck-bounds-decls).    This is because the checking that declared bounds are implied by inferred bounds is not very powerful.  It is limited to just syntactically identical bounds declarations and special cases for bounds(none) and bounds(any).    There'd be lots of spurious warnings if it was on by default.

The checking of bounds declarations is done only for assignments.  This extends it to variable declarations with initializers.   Most of the work went into producing good warning messages that have notes explaining what is going on.   Assignments were using expanded bounds declarations (where count/byte_count have been expanded into their underlying range expressions).   I implemented new logic to create expanded bounds declarations for variable declarations with bounds.   This is so that the contents of notes match up.  I had to make some methods public on BoundsInference.  It isn't clear that the methods are in the right place, but I've left that code refactoring for another day.

Other than that, I refactored the logic for checking bounds declarations for assignments so that we could re-use it for declarations with initializers.

Testing:
- Clang test run in progress.
- Added a new clang-specific tests that checks bounds inference for initializers.  Also check the notes generated with warnings.

### DIFF
--- a/test/CheckedC/bounds-decl-checking.c
+++ b/test/CheckedC/bounds-decl-checking.c
@@ -13,7 +13,7 @@ void f1(_Array_ptr<int> p : bounds(p, p + x), int x) {
 // Initialization by expression with syntactically identical
 // normalized bounds - no warning.
 void f2(_Array_ptr<int> p : bounds(p, p + x), int x) {
-  _Array_ptr<int> r : bounds(p, p + x) = 0;
+  _Array_ptr<int> r : bounds(p, p + x) = p;
 }
 
 // Initialization by expression without syntactically identical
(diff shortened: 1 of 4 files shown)


test/CheckedC/bounds-decl-checking.c
```diff
@@ -3,14 +3,46 @@
...
 void f1(_Array_ptr<int> p : bounds(p, p + x), int x) {
     _Array_ptr<int> r : bounds(p, p + x) = 0;
-    r = p;
 }
 
-void f2(_Array_ptr<int> p : count(x), int x) {
+// Initialization by expression with syntactically identical
+// normalized bounds - no warning.
+void f2(_Array_ptr<int> p : bounds(p, p + x), int x) {
+  _Array_ptr<int> r : bounds(p, p + x) = 0;
```
lenary: The comment says "identical normalized bounds" but the initializer is null. Typo, or my misunderstanding?

dtarditi:
I've fixed the test case for syntactically identical bounds so they actually check  that.

Issue #312, opened 2017-05-31
Test IR generated for runtime checking for dynamic_bounds_cast


dtarditi:
We have clang-specific tests that check the IR generated for runtime checks for bounds checks.   It is in clang/test/CheckedC/dynamic-checks.  We need corresponding tests for the IR generated for dynamic_bounds_cast.

PR #360, opened 2017-08-12
Remove Ptrtoint from Dynamic Check CG

lenary:
It turns out you can icmp pointer types in LLVM IR, without converting them to integers first. This pull request removes our generation of "ptrtoint" instructions.

However there is an issue in inference, ICMP needs both operandi to have the same type, and for complex reasons sometimes the type of the bounds doesn't match the types of the pointer we're comparing to. This means sometimes we have to insert a bitcast in the LLVM IR. Semantically this keeps the same value but changes the IR type to match what is needed. I need to add code comments to clarify how we use these casts. 

I also cleared up the method name used for generating dynamic_bounds_cast so it doesn't overload that of the dynamic bounds check, and split out statistics for dynamic checks from casts to be desperate to bounds checks. 

Tested:
- Windows 32bit
- Mac 64 bit. 

No tests failed when I cleaned up and clarified the dynamic bounds cast code, which worries me slightly. Do we not test the LLVM IR that's generated?

### DIFF
--- a/lib/CodeGen/CGDynamicCheck.cpp
+++ b/lib/CodeGen/CGDynamicCheck.cpp
@@ -28,6 +28,7 @@ namespace {
   STATISTIC(NumDynamicChecksNonNull, "The # of dynamic non-null checks found");
   STATISTIC(NumDynamicChecksOverflow, "The # of dynamic overflow checks found");
   STATISTIC(NumDynamicChecksRange, "The # of dynamic bounds checks found");
+  STATISTIC(NumDynamicChecksCast, "The # of dynamic cast checks found");
 }
 
 //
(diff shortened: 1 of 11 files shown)


lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -94,26 +95,28 @@ void CodeGenFunction::EmitDynamicBoundsCheck(const Address PtrAddr, const Bounds
 
   // Emit the code to generate the pointer values
   Address Lower = EmitPointerWithAlignment(BoundsRange->getLowerExpr());
-  Address Upper = EmitPointerWithAlignment(BoundsRange->getUpperExpr());
+  if (Lower.getType() != PtrAddr.getType())
```
dtarditi: It would be good to add comments that explain why this is necessary.

dtarditi:
There's already a work item that we're not checking the LLVM IR for dynamic_bounds_cast: #312.

lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -139,78 +142,90 @@ void CodeGenFunction::EmitDynamicBoundsCheck(const Address BaseAddr,
...
+  EmitBlock(DyCkSubsumption);
+
+  Builder.SetInsertPoint(DyCkSubsumption);
 
-  Builder.SetInsertPoint(DyCkFallThrough);
   // SubRange - bounds(lb, ub) vs CastRange - bounds(castlb, castub)
   // Dynamic_check(lb <= castlb && castub <= ub)
 
-  // Emit the code to generate the pointer values
+  // Emit the code to generate lb and ub
```
lenary: `lb` and `ub` reference the two lines above, where we have `SubRange - bounds(lb, ub)`.

lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -139,78 +142,90 @@ void CodeGenFunction::EmitDynamicBoundsCheck(const Address BaseAddr,
...
+  Value *CastCond =
+      Builder.CreateAnd(LowerChk, UpperChk, "_Dynamic_check.cast");
 
-  Builder.SetInsertPoint(DyCkFail);
-  CallInst *TrapCall = Builder.CreateCall(CGM.getIntrinsic(Intrinsic::trap));
-  TrapCall->setDoesNotReturn();
-  TrapCall->setDoesNotThrow();
-  Builder.CreateUnreachable();
+  // Constant Folding:
+  // If IsNull is true (one), then a) we don't need to insert the rest
```
lenary: Will Change.

lib/CodeGen/CGDynamicCheck.cpp
lenary: Yep, on my list.

lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -139,80 +150,138 @@ void CodeGenFunction::EmitDynamicBoundsCheck(const Address BaseAddr,
...
-  TrapCall->setDoesNotReturn();
-  TrapCall->setDoesNotThrow();
-  Builder.CreateUnreachable();
+  // Constant Folding:
+  // If CastCond is true (one), then we need to insert a direct branch
+  // to the success block, emit it, and set the insert point there for
+  // further code generation.
+  if (const ConstantInt *CastCondConstant = dyn_cast<ConstantInt>(CastCond)) {
+    if (CastCondConstant->isOne()) {
+      ++NumDynamicChecksElided;
```
lenary: You asked "how do we know constant folding is correct". Well, turns out in writing comments about how it should work made me find a pretty major bug in it, which I have fixed in this most recent commit. 

I'll file a ticket about writing tests for this, but maybe in the short term I should also add a flag to tell the codegen to constant fold that defaults to `false`. What's your feeling on this?

dtarditi: I think it is fine to write tests later, and save the time of adding a flag and then removing it.

lenary: Ok, let me know when you're happy for this to be merged.

lenary:
I thought this was only a semantics issue. Turns out it actually helps remove two dynamic checks from the hot loop of `anagram.c`.

Issue #323, opened 2017-06-15
Bounds-safe interfaces on function pointers not being used during typechecking.


dtarditi:
The following code snippet:
```
typedef int(*callback_fn1) (int *a : count(n) , int n); 
void test2(callback_fn1 fn, array_ptr<int> arr : count(k), int k) {
   (*fn)(arr, k);
}
```
results in a compiler error message:
```
callback2.c:6:9: error: passing '_Array_ptr<int>' to parameter of incompatible type 'int *'
  (*fn)(arr, k);
        ^~~
1 error generated.
```
This code should work fine and not result in an error message.   This may be a bug or missing functionality.

dtarditi:
The problem appears to be missing functionality.   Typechecking of individual arguments against their formal parameters is done by Sema::GatherArgumentsForCall in SemaExpr.cpp,     If it is a direct call, the function records the formal parameter in "InitializedEntity" Entity.   This carries the bounds-safe interface on it.  It is an indirect call, the function records only the parameter type.  So the bounds-safe interface for the parameter that is (or should be) part of the function type is not propagated.   

Type checking is done in SemaInit.cpp in InitializationSequence::Perform, under the case CK_CAssignment (around line 7056). This typechecking step at line 7072 will fail because there is no bounds-safe interface to pick up.

dtarditi:
This is fixed.

PR #363, opened 2017-08-29
Use bounds-safe interfaces on function pointers during type checking

dtarditi:
This addresses Github issue #323.  The bounds-safe interfaces on parameters for function pointers were not being used during type checking.    This modifies type checking for function calls to obtain bounds-safe interfaces for parameters from function types, instead of from the declaration of the function being called.   For indirect function calls, there is no declaration.  SemaInit does type checking of parameter passing.  It is extended to track bounds for entities.  For parameters, the bounds are set using the bounds from function types.

Most of the work for this change was revamping the error messages for uses of unchecked types in checked scopes.    It is unclear what the term “unchecked type” means when constructed types are involved.  Constructed types are types such as function types, array types, and pointer types.  The problem may be a type that a type is constructed from (for example, a function type involving a parameter with a unchecked pointer type) or it may be in a bounds-safe interface.  In addition, the problematic unchecked type may be unclear. 

We revamped the error messages to be clearer and added notes explain what is going on.  We:
- use a simple error message for simple errors such as trying to declare or use a variable or parameter with an unchecked pointer or array type in a checked scope
- use a longer error message when the problem is in a type that a type has been constructed from.
- use a different error message when the problem is the use of bounds-safe interface with an unchecked type in a checked scope.

We customize the error message based on whether the problem is a definition or a use.  For a use, we add a diagnostic note that points out the definition.   If it is not obvious what the problematic unchecked type is, we add a diagnostic note with the problematic type. We do this when the problem is a type that the type was constructed from, a type occurring in a bounds-safe interface, or the type is a type definition.

Testing:
- There will be a separate pull request for the Checked C repo the updates the checked scope tests for the new error messages.
- Add clang-specific tests of the notes and error messages for uses of unchecked types in checked scopes.
- Add tests of AST dumps with bounds-safe interface types, since those were missing.
- Passed existing Checked C and clang tests.


include/clang/Sema/Sema.h
```diff
@@ -9421,11 +9438,13 @@ class Sema {
...
 private:
-  QualType GetCheckedCInteropType(const ValueDecl *Decl);
+  QualType GetCheckedCInteropType(const DeclaratorDecl *Decl);
 public:
-  /// \brief Get the bounds-safe interface type for Entity.
-  /// Returns a null QualType if there isn't one.
-  QualType GetCheckedCInteropType(const InitializedEntity &Entity);
+  /// \brief Given a value with type Ty and bounds Bounds,
+  /// compute the bounds-safe interface type.
+  QualType Sema::GetCheckedCInteropType(QualType Ty,
```
lenary: This gives a compiler error on my machine, you don't need the `Sema::` prefix as you're already in that scope.

```
/path/to/checkedc-llvm/tools/clang/include/clang/Sema/Sema.h:9445:18: error: extra qualification on member 'GetCheckedCInteropType'
  QualType Sema::GetCheckedCInteropType(QualType Ty,
           ~~~~~~^
1 error generated.
```

lib/Sema/SemaChecking.cpp
```diff
@@ -11582,46 +11638,58 @@ bool Sema::DiagnoseCheckedDecl(const ValueDecl *Decl, SourceLocation UseLoc) {
   }
   else if (const VarDecl *Var = dyn_cast<VarDecl>(Decl)) {
     TargetDecl = Var;
-    DeclKind = 2; // decl var
+    DeclKind = Var->isLocalVarDecl() ? 2 : 3; // decl var
```
lenary: Where are these constants coming from? Are they for the error reporting? Do we need a local enum?

lib/Sema/SemaType.cpp
```diff
@@ -3717,6 +3717,22 @@ QualType Sema::MakeCheckedArrayType(QualType T, bool Diagnose,
...
 /// construct and return the checked type for the interface. Otherwise
 /// return a null QualType.
+QualType Sema::GetCheckedCInteropType(const DeclaratorDecl *Decl) {
+  const BoundsExpr *Bounds = Decl->getBoundsExpr();
+  if (!Bounds)
+    return QualType();
+
+  QualType Ty;
+  if (const FunctionDecl *FD = dyn_cast<FunctionDecl>(Decl))
+    Ty = FD->getReturnType();
```
lenary: Why do we return only the return type, don't we want the whole function type?

lib/Sema/SemaType.cpp
```diff
@@ -3725,68 +3741,54 @@ QualType Sema::MakeCheckedArrayType(QualType T, bool Diagnose,
...
-          Ty = FD->getReturnType();
-        else
-          Ty = Decl->getType();
+  // Parameter types that are array types are adusted to be
+  // pointer types.  We'll work with the original type instead.
+  // For multi-dimensional array types, the nested array types need
+  // to become checked array types too.
+  if (isParam) {
+    if (const DecayedType *Decayed = dyn_cast<DecayedType>(Ty)) {
+      Ty = Decayed->getOriginalType();
```
lenary: This seems like a good improvement.

lib/Sema/SemaType.cpp
```diff
@@ -3725,68 +3741,54 @@ QualType Sema::MakeCheckedArrayType(QualType T, bool Diagnose,
...
+        if (PtrType->isUnchecked()) {
+          ResultType = Context.getPointerType(PtrType->getPointeeType(),
+                                              CheckedPointerKind::Array);
+          ResultType.setLocalFastQualifiers(Ty.getCVRQualifiers());
         }
-        break;
       }
-      default:
-        break;
+      else if (Ty->isConstantArrayType() || Ty->isIncompleteArrayType())
```
lenary: Can we really do "interop-type inference" for incomplete arrays?

lenary:
I just ran clang from this patch against the converted benchmarks and there are no errors, which is good.

include/clang/Sema/Sema.h
dtarditi: Fixed - thanks for pointing this out.

lib/Sema/SemaChecking.cpp
dtarditi: Replaced magic constants with an enum.

lib/Sema/SemaType.cpp
dtarditi: Good catch - I agree this is confusing.  The prior code wanted the return type, and this was left over from that use.   I looked more closely and saw that there was only one other method calling this method.  It wasn't calling this on FunctionDecls anyway.   I cleaned up the code by deleting this method and folding its functionality its caller.

lib/Sema/SemaType.cpp
dtarditi: Yes, for parameters.   For an array with variable length, you can write:
```
void f(int arr[], int len)
```
In Checked C this might become:
```
void f(int arr[] : count(len), int len)
```
By the way, it is considered bad form in C to write something like:
```
void f(int arr[5]);  // deceptive, no checking that callers actually
                     // pass arrays with 5 elements.
```
There's no guarantee in C that `arr` has 5 elements.   There's no language-level checking required and some C compilers will allow things like:
```
void bad_call(void) {
  int arr[4];
  f(arr);
}
```
Clang provides some limited checking , but it'll let you pass a pointer to f without any complaints
```
sketchy_call(int *p) {
  f(p);
}
```

Issue #364, opened 2017-08-29
Allow checked pointers to be cast to any kind of checked void pointer


dtarditi:
Currently we require that a checked pointer be cast to the same kind of void pointer.  For example, `_Array_ptr<int>` can be cast to `_Array_ptr<void>`, but not `_Ptr<void>`.  We should allow casting to `_Ptr<void>`.  This would address issue #318.

dtarditi:
This work is complete.

PR #365, opened 2017-08-30
Conversions involving checked void pointers

dtarditi:
This addresses GitHub issue #364.   Allow any kind of checked pointer to be implicitly converted to any kind of checked void pointer.   Previously we required that the target and destination types be the same kind of checked pointer.  

Also disallow conversions between void pointers and checked function types.   In examining the code for conversions involving void pointers, I noticed that the code allowed conversions between void * and function pointers as an extension.   This would also allow conversions between checked void pointers and function pointers.  We do not want to allow this, even if the same static analysis for function pointers might disallow this.

Improve the comment about conversions between different checked pointers to the same referent type.

Testing:
- Update tests in the Checked C repo to allow the conversions to different kinds of checked void pointers.  This is covered by a separate pull request.
- Add  tests that conversions between void pointers and checked function pointers are not allowed.
- Passes existing Checked C and clang tests.

### DIFF
--- a/lib/Sema/SemaExpr.cpp
+++ b/lib/Sema/SemaExpr.cpp
@@ -7474,37 +7474,41 @@ checkPointerTypesForAssignment(Sema &S, QualType LHSType, QualType RHSType) {
   // version of void...
 
   // Allow conversion from any unchecked pointer to any kind of void pointer,
-  // including void *.  In Checked C, conversions from checked pointers to void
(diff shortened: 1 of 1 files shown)


dtarditi:
I've revamped the code to be easier to understand.  I split the code into separate cases: the "plain C" case where the pointer types are unchecked and the "Checked C" case where the source or target type is a checked pointer.

PR #396, opened 2017-10-17
String literals with checked types and improved support for array literals.

dtarditi:
This adds compiler support for using string literals in checked scopes.   In C, a string literal has the type "array of character", where there are several possible character types.  In Checked C, that corresponds to an "unchecked array of character".  To be able to use string in checked scopes, we need a way that they can be typed as a "checked null-terminated array of characters."

We support this in two ways.  First we alter the type of string literals that appears in checked scopes to have null-terminated array types.   Array-to-pointer conversion converts this to having an nt_array_ptr test, and at assignments nt_array_ptr can converted array_ptr..

Second,  we allow array literals (of which strings are a special case) to be converted implicitly to a checked pointer type when a checked pointer type is expected.  The checked pointer type should have the same pointer kind as the expected pointer type.    The implicit conversions can happen at assignments, function calls, and within initializers.    The language rules haven't been added to the Checked C specification yet.  The proposed rule is described[here](https://github.com/Microsoft/checkedc/issues/207).

Array literals specified using initializers or compound literal expressions already worked for the most part.   The C language rules use the type of the variable being initialized or the type of the compound literal expression as guides for filling in the type.   The clang machinery for filling in arrays worked for checked arrays too.   

This commit adds one important missing case: incomplete array types.   We were not propagating the checked enum when the complete array type was formed after the number of array elements had been determined.   It also checks in a checked scope that the compound literal type is a checked type.   This closes off one more route by which an unchecked type could appear in a checked scope.

This commit adds a new method for checking types that appear syntactically in expressions in checked scopes (for example, in cast operators and compound literal expressions).    This produces error message with similar detail to those produced for declarations, calling out the problematic unchecked type within a constructed type if necessary.    The existing error message for cast operations are switched to use this method, so we get better error message for cast expressions too.

Testing:
- Added new tests in the Checked C repo in initializers.c that cover strings, array initializers, and pointers initialized via  strings or compound literal expressions with array type.
- Passes existing automated testing for Linux and Windows.  I had to modify several LNT benchmarks because the improved typing discovered that we were passing strings to methods with bounds-safe interfaces.

### DIFF
--- a/lib/Sema/SemaExpr.cpp
+++ b/lib/Sema/SemaExpr.cpp
@@ -1731,9 +1731,13 @@ Sema::ActOnStringLiteral(ArrayRef<Token> StringToks, Scope *UDLScope) {
   // Get an array type for the string, according to C99 6.4.5.  This includes
   // the nul terminator character as well as the string length for pascal
   // strings.
+  CheckedArrayKind ArrayKind = getCurScope()->isCheckedScope() ?
+    CheckedArrayKind::Checked : CheckedArrayKind::Unchecked;
+
   QualType StrTy = Context.getConstantArrayType(CharTyConst,
-                                 llvm::APInt(32, Literal.GetNumStringChars()+1),
-                                 ArrayType::Normal, 0);
+                                                llvm::APInt(32, Literal.GetNumStringChars() + 1),
+                                                ArrayType::Normal, 0,
+                                                ArrayKind);
 
   // OpenCL v1.1 s6.5.3: a string literal is in the constant address space.
   if (getLangOpts().OpenCL) {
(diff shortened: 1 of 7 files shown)


lib/Sema/SemaExpr.cpp
```diff
@@ -1731,9 +1731,13 @@ Sema::ActOnStringLiteral(ArrayRef<Token> StringToks, Scope *UDLScope) {
   // Get an array type for the string, according to C99 6.4.5.  This includes
   // the nul terminator character as well as the string length for pascal
   // strings.
+  CheckedArrayKind ArrayKind = getCurScope()->isCheckedScope() ?
+    CheckedArrayKind::Checked : CheckedArrayKind::Unchecked;
+
   QualType StrTy = Context.getConstantArrayType(CharTyConst,
-                                 llvm::APInt(32, Literal.GetNumStringChars()+1),
-                                 ArrayType::Normal, 0);
+                                                llvm::APInt(32, Literal.GetNumStringChars() + 1),
```
lenary: Is this constant `32` correct? Surely it should be the same as the width of `size_t` which you can get from… somewhere?

PR #401, opened 2017-10-24
Rename bounds(none) to bounds(unknown).

dtarditi:
Change the compiler to match the specification and update the clang-specific Checked C tests.

Testing:
- There is a corresponding pull request for the Checked C repo.
- Automated testing running now.

### DIFF
--- a/lib/AST/StmtPrinter.cpp
+++ b/lib/AST/StmtPrinter.cpp
@@ -1905,8 +1905,8 @@ void StmtPrinter::VisitNullaryBoundsExpr(NullaryBoundsExpr *Node) {
     case BoundsExpr::Any:
       OS << "bounds(any)";
       break;
-    case BoundsExpr::None:
-      OS << "bounds(none)";
+    case BoundsExpr::Unknown:
+      OS << "bounds(Unknown)";
       break;
     default:
       llvm_unreachable("unexpected bounds kind for nullary bounds expr");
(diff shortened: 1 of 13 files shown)


lib/AST/StmtPrinter.cpp
```diff
@@ -1905,8 +1905,8 @@ void StmtPrinter::VisitNullaryBoundsExpr(NullaryBoundsExpr *Node) {
     case BoundsExpr::Any:
       OS << "bounds(any)";
       break;
-    case BoundsExpr::None:
-      OS << "bounds(none)";
+    case BoundsExpr::Unknown:
+      OS << "bounds(Unknown)";
```
lenary: You probably want this to be `bounds(unknown)` (see case) if it's to match the parsed expression

lib/Parse/ParseExpr.cpp
```diff
@@ -2961,17 +2961,17 @@ ExprResult Parser::ParseBoundsExpression() {
...
   } else if (Ident == Ident_bounds) {
-    // Parse bounds(none) or bounds(e1, e2)
+    // Parse bounds(unknown) or bounds(e1, e2)
     bool FoundNullaryOperator = false;
 
     // Start with "none"
     if (Tok.getKind() == tok::identifier) {
       IdentifierInfo *NextIdent = Tok.getIdentifierInfo();
-      if (NextIdent == Ident_none) {
+      if (NextIdent == Ident_unknown) {
```
lenary: I totally forgot that we parsed `bounds(none/unknown)` rather than just had it for inference. I should update my report.

lib/AST/StmtPrinter.cpp
dtarditi: Yes, I've fixed this.

lib/Parse/ParseExpr.cpp
dtarditi: I am fine either way.

Issue #405, opened 2017-11-02
Correctly determine type of integer literals in count expressions


dtarditi:
For arrays, we construct a count bounds expression with an integer literal determined by the size of the array.  We are currently typing the integer literal as a 64-bit integer.  This causes subtle disagreement in types for compiler-generated literals vs. ones written by users.  We should follow the C rules for typing integer literals, which is that an integer literal is given the smallest integer type that it can fit into, from the set of `int`, `long int`, and `long long int`.

 The C specification is somewhat ambiguous because for hex/octal integer literals, it includes unsigned versions of those types.   This gives those literals a slightly larger maximum range, but we don't need that range for arrays.

This retyping will cause numerous changes in existing Checked C tests that examine ASTs.

dtarditi:
This is complete.

PR #406, opened 2017-11-02
Add default bounds for null-terminated array pointers

dtarditi:
This change adds a default bounds of count(0) for null-terminated array pointers.    This bounds is used when no bounds is specified for an nt_array_ptr variable or in cases where we currently do not provide a way to specify bounds (such as the bounds of a pointer returned by a pointer dereference).

- Add prebuilt bounds expressions for count(0), count(1), and bounds(unknown).   We use count(0) and bounds(unknown) for now.  I deferred switching to count(1) because I want to make sure fix the types of integer literals first (issue #405).
- Extend bounds expressions with a bit that tells whether a bounds expression is compiler-generated or not.  The bit is not meaningful from a semantics-checking perspective.  It is used when printing error messages so that we know when we are dealing with bounds that did not appear in the program text.
- For variables, members, and returns, add the default bounds of count(0) early during semantics checking.  We now always call ActOnBoundsDecl, providing it with a null parameter for the bounds expression when none is specified.
- Extend bounds checking to infer count(0) for pointers returned by memory access operations (both lvalue targets and lvalues).    Also infer count(0) for bounds inferred for interface types (I eventually plan to move that inference earlier when I separate the representation of interface types from bounds expressions).
- In the IR, we currently represent no bounds specified by the user using null pointers.  I experimented with adding prebuilt bounds of bounds(unknown) during semantics checking to variables and methods, but decided not to.  This made handling of type compatibility for unchecked pointer types more complex,.  It would also trigger the allocation and unnecessary processing of arrays of bounds for function types.  It also made handling of .
- I did decide to make the code more robust about how it handles the equivalence between "no bounds" and bounds unknown.   The method `EquivalentBounds` now handles null pointers and checks for that case.  This led to some clean up in existing code for checking type compatibility and redeclarations.
- Bug fix: for non-scalar initializers, don't check the bounds of the right-hand side value.   The C semantics for initializers ensure that a properly-sized value is allocated.

Testing:
- Extending existing Checked C tests with tests for nt_array_ptr.  Those tests will be in a separate pull request for the Checked C repo.
- Passes existing automated testing for Windows and Linux.

### DIFF
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -796,11 +796,6 @@ namespace {
 
           BoundsExpr *B = D->getBoundsExpr();
 
-          // Null-terminated pointers with no bounds get a zero-element
-          // bounds.
-          if (!B && DR->getType()->isCheckedPointerNtArrayType())
-            B = Context.getPrebuiltCountZero();
-
           if (!B || B->isUnknown())
             return CreateBoundsAlwaysUnknown();
 
@@ -858,12 +853,6 @@ namespace {
             return CreateBoundsInferenceError();
 
           BoundsExpr *B = F->getBoundsExpr();
-
-          // Null-terminated pointers with no bounds get a zero-element
-          // bounds.
-          if (!B && F->getType()->isCheckedPointerNtArrayType())
-            B = Context.getPrebuiltCountZero();
-
           if (!B || B->isUnknown())
             return CreateBoundsAlwaysUnknown();
 
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -476,6 +476,13 @@ namespace {
     // Given an array type with constant dimension size, produce a count
     // expression with that size.
     BoundsExpr *CreateBoundsForArrayType(QualType QT) {
+      const IncompleteArrayType *IAT = Context.getAsIncompleteArrayType(QT);
+      if (IAT) {
+        if (IAT->getKind() == CheckedArrayKind::NtChecked)
+          return Context.getPrebuiltCountZero();
+        else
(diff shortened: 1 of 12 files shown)


lib/Sema/SemaBounds.cpp
```diff
@@ -797,6 +818,36 @@ namespace {
            Expr *Base = CreateImplicitCast(QT, CastKind::CK_LValueToRValue, E);
            return ExpandToRange(Base, B);
         }
+        case Expr::UnaryOperatorClass: {
+          UnaryOperator *UO = dyn_cast<UnaryOperator>(E);
```
awruef: Could you use `cast<>` instead of `dyn_cast<>` followed by a check, if what you do in the check is assert?

lib/Sema/SemaBounds.cpp
```diff
@@ -797,6 +818,36 @@ namespace {
...
+              UO->getType()->isCheckedPointerNtArrayType())
+              return CreateTypeBasedBounds(UO, UO->getType(), false, false);
+          return CreateBoundsAlwaysUnknown();
+        }
+        case Expr::ArraySubscriptExprClass: {
+          //  e1[e2] is a synonym for *(e1 + e2).  The bounds are
+          // the bounds of e1 + e2, which reduces to the bounds
+          // of whichever subexpression has pointer type.
+          ArraySubscriptExpr *AS = dyn_cast<ArraySubscriptExpr>(E);
+          if (!AS) {
```
awruef: Same as above.

lib/Sema/SemaBounds.cpp
lenary: There's a lot of code that uses `dyn_cast` with a check afterwards. We should move over to `cast`, but let's do that in a separate PR after the deadline?

PR #408, opened 2017-11-03
Fix internal assert caused by visiting AST nodes twice.

dtarditi:
This fixes a compiler crash reported in https://github.com/Microsoft/checkedc/issues/221.  The checking of bounds information was failling an assertion that bounds are only set once for cast expression nodes.  I tracked the problem to an issue in RecursiveASTVisitor.h.  It turns out that nodes in initializer lists are visited multiple times if they appear in both semantic and syntactic forms of an initializer list. The comment in RecursiveASTVisitor.h about AST nodes being visited exactly once isn't quite accurate.

The fix is to borrow an approach used in lib\index\IndexBody.cpp and visit only one form.  In our case, we want to visit the semantic form. We'll eventually have more complex checking that ensures that structs are initialized to satisfy their member bounds invariants, and we'll need to use the semantic form for that.

Testing:
- Added a regression test case to the test\CheckedC directory.
- Passed clang regression tests on Linux.  Additional automated testing progress.

### DIFF
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -1428,13 +1428,16 @@ namespace {
     // RecursiveASTVisitor visits both syntactic and semantic forms of
     // initializer lists, causing AST nodes used in both forms to be visited
     // twice by default. The statement in RecursiveASTVisitors that AST nodes
-    // are visit exactly once isn't quite correct.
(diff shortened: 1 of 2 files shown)


mwhicks1:
This works on the vsftpd code that prompted my bug report.

PR #411, opened 2017-11-08
Correctly determine type of integer literals

dtarditi:
For arrays, we construct a count bounds expression with an integer literal determined by the size of the array.  We have been typing the integer literal as a 64-bit unsigned integer.  This causes subtle disagreements in types for compiler-generated literals vs. ones written by users.   This change updates the integer literal construction method to follow the C rules for typing integer literals.  An an integer literal is given the smallest integer type that it can fit into, from the set of `int`, `long int`, `long long int`, and 'unsigned long long int'.  This addresses issue #405.

Testing:
- For the Checked C tests that verify IR, update the types assigned to literals.
- Passed locally on x86.
- Passes existing automation for x64 Linux and Windows.  I discovered that bit sizes for integer literals using APInt shouldn't really be trusted: they can be larger than the number of bits needed to represent the integer.  For x64, array sizes come back as APInts with 64 bits.

### DIFF
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -452,15 +452,18 @@ namespace {
     }
 
     bool Fits(QualType Ty, const llvm::APInt &I, llvm::APInt &Result) {
+      assert(Ty->isSignedIntegerType());
       unsigned bitSize = Context.getTypeSize(Ty);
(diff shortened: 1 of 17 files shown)


dtarditi:
I looked and couldn't find a utility.  There is a general-purpose code for dealing with constant literals, but it handles lots of other cases and had nothing easily extractable.

PR #414, opened 2017-11-10
Bounds declaration checking of constant-sized ranges

dtarditi:
This change extends bounds declaration checking to understand constant-sized ranges.  It also adds checking of static casts to bounds declarations checking (a static cast is a cast to checked pointer types using the C cast operator, in the Checked C spec the bounds have to be statically verifiable).

A constant-sized range has the from (e1 + const1, e1 + const2), where e1 is a non-modifying expression (all complete array types in C have constant-sized ranges, excluding variable-length arrays).  Bounds expressions that are constant-sized ranges are recognized and converted into an internal form.   The bounds expression has to be expressed as pointer arithmetic (e1 + const, const + e1, or e1 - const).   Existing clang code is used for recognizing constant expressions.  When checking whether the source bounds expression for an expression implies the destination bounds expression for the expression, it is checked whether two constant-sized ranges have the same base expression.  If they do, the constant offsets are used to check the truth or falsity of this.  Even if the constant-sized ranges do not have the same base expression, their width is checked.  If the source bounds expression is narrower than the destination bounds expression, an bounds declaration checking error is reported.

When checking bounds declarations, the checker now returns three possible values: true, false, or maybe.  It also returns a separate value explaining why a false result was obtained. These are used to produce diagnostics.  If there are multiple causes of failure, we try to report them.

Bounds declaration checking is extended to check static casts to pointer types.  It follows the same pattern as other operations, with one important difference.   The Checked C spec only requires that the source bounds be wider than the target bounds for casts between Ptr types.  Combined with constant-sized ranges, this lets us check all `Ptr<T>` to `Ptr<S>` casts.  Casts from array_ptrs with variable width bounds still won't be checked, though.

APSInts are used to represent the constant-sized offsets for constant ranges.  These are a little tricky to work with.  They carry a bit with them that indicates whether they are a signed or unsigned integer.  APSInt is a sublcass of APInt, which represents the bits associated with the integer.  The bits are interpreted differently for signed vs. unsigned integers.  APInts have bitwidths associated with them and only operations on integers with the same bitwidth are allowed.  This means that we have to convert integers to have the same bitwidth at times.  We use the bitwidth of pointers as the bitwidth of integers.  This leads some issues with large unsigned integers.  For now, we ignore constant ranges involving unsigned integer offsets >= the max signed integer that will fit into a pointer bitwidth (by ignore, we just treat it was not being a constant range).

Bounds declaration checking is extended to treat `_Ptr<void>` as having a 1 character width.  This is done by using _Bytecount instead of _Count for it.  The 1 character requirement is based on the C requirement that any void * be convertible to a char *. We have a similar rule for `_Ptr<void>`, which means `_Ptr<void>` has to be at least 1 character wide.

For casts to pointers to _Nt_checked arrays, we need to be able to determine the physical size of the right-hand side.  Normally in bounds inference, we exclude the null terminator from the bounds of an array. Add a flag that optionally lets it be included for this calculation.

Testing:
- In the Checked C repo, there were a number of tests that were flagged with TODOs where bounds declaration checking errors were expected.  Update a number of them with the expected error messages.  There will be a separate pull request covering this.
- Extend the bounds_decl_checking.c file in the clang\tests\CheckedC tests with additional tests of bounds declaration checking.  These tests include the expected diagnostics notes explaining what happened.
- One of the LNT tests (Olden\bh) now fails to compile because bounds declaration checking is detecting casts from pointers to smaller objects to pointers to larger objects (equivalent to unchecked upcasts).   The other LNT tests compile and run successfully.
- Passes automated testing.

### DIFF
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -459,7 +459,7 @@ namespace {
 
     // Determine if the mathemtical value of I (an unsigned integer) fits within
     // the range of Ty, a signed integer type.  APInt requires that bitsizes
-    // match exactly, so if it does, fit, return an integer via Result with
+    // match exactly, so if I does fit, return an APInt via Result with
     // exactly the bitsize of Ty.
     bool Fits(QualType Ty, const llvm::APInt &I, llvm::APInt &Result) {
       assert(Ty->isSignedIntegerType());
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -36,8 +36,6 @@
 #include "llvm/ADT/SmallString.h"
 #include "TreeTransform.h"
 
-
-
 using namespace clang;
 using namespace sema;
 
@@ -1436,17 +1434,22 @@ namespace {
       return false;
     }
 
+    // The result of trying to prove a statement about bounds declarations.
+    // The proof system is incomplete, so there are will be statements that
+    // cannot be proved true or false.  That's why "maybe" is a result.
     enum class ProofResult {
       True,  // Definitely provable.
       False, // Definitely false (an error)
       Maybe  // We're not sure yet.
     };
 
+    // ProofFailure: codes that explain why a statement is false.  This is a
+    // bitmask because there may be multiple reasons why a statement false.
     enum class ProofFailure : unsigned {
       None = 0x0,
-      LowerBound = 0x1,
-      UpperBound = 0x2,
-      Width = 0x4
+      LowerBound = 0x1,   // The lower bound is out-of-range.
+      UpperBound = 0x2,   // The upper bound is out-of-range.
+      Width = 0x4         // The source value is not wide enough.
     };
 
     enum class DiagnosticNameForTarget {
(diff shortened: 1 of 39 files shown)


dtarditi:
Update comment so that markdown doesn't misinterpret the commit text and cross half of it out.

lib/Sema/SemaBounds.cpp
```diff
@@ -1418,20 +1434,280 @@ namespace {
...
+            Offset = Offset.smul_ov(ElemInt, Overflow);
+            if (Overflow)
+              goto exit;
+            return;
+          }
+        }
+      }
+
+    exit:
+      Base = Base = E->IgnoreParens();
```
lenary: Not clear what the double assignment to `Base` is doing here.

lib/Sema/SemaBounds.cpp
```diff
@@ -1418,20 +1434,280 @@ namespace {
...
+        if (BO->isAdditiveOp()) {
+          Expr *Other = nullptr;
+          if (BO->getLHS()->getType()->isPointerType()) {
+            Base = BO->getLHS();
+            Other = BO->getRHS();
+          } else if (BO->getRHS()->getType()->isPointerType()) {
+            Base = BO->getRHS();
+            Other = BO->getLHS();
+          } else
+            goto exit;
```
lenary: It seems this goes to `exit` if neither the LHS or RHS have pointer type. Except you also goto `exit` in error conditions. It's not clear these should have the same logic.

dtarditi: I'll add a comment explain what's going on here.  We try to split an expression into a base + offset, where offset is in characters.  If we can't, for whatever reason, we return (E, 0).   

This includes pointer arithmetic overflow (where the size of the object times a constant integer overflows).  Eventually we ought to flag that as another error, but right now we're being imprecise for large unsigned integer offsets, so that can wait for another day.

lib/Sema/SemaBounds.cpp
dtarditi: Typo - I'll fix this.

PR #420, opened 2017-11-14
Add booleans about checked properties to compound statements

dtarditi:
This adds two Booleans about checked properties to compound statements:
- IsCheckedScope: is this semantically a checked scope?
- IsCheckedPropertyDeclared: was this information explicitly declared for the block (using `_Checked { ... }` or `_Unchecked { ... }`.

I extended AST dumps to include this information for compound statements.  I specifically omit dumping information for the existing behavior (where there is no declared property and the scope is unchecked).   I do this to avoid breaking any existing tests that depend upon ASTs not having this information.   I think there's only one test that breaks, but there could be more.

I didn't update AST reading/writing to record this information.  I opened issue #419 to track this remaining work.

This does reduce the number of statements allowed in a compound statement from 2^24 to 2^22.  I think we can live with that for now.

Testing:
- Add new test that checks that AST dumps include this information.
- Passed automated testing.


include/clang/AST/Stmt.h
```diff
@@ -90,8 +90,12 @@ class alignas(void *) Stmt {
   class CompoundStmtBitfields {
     friend class CompoundStmt;
     unsigned : NumStmtBits;
+    // Is this compound statement checked or unchecked?
+    unsigned IsCheckedScope : 1;
+    // Was the property declared explicitly in the code?
+    unsigned CheckedPropertyDeclared : 1;
 
-    unsigned NumStmts : 32 - NumStmtBits;
+    unsigned NumStmts : 32 - 2 - NumStmtBits;
```
lenary: Oh god. We really reducing the number of statements in a block to a quarter of its previous value? If this turns into an issue I'm going to be really sad.

PR #421, opened 2017-11-15
Warn about unprovable bounds declarations in checked scopes. 

dtarditi:
The Checked C compiler will now always warn in checked scopes about bounds declarations that cannot be proved true or false.  For unchecked scopes, the warning is off by default.

I replaced the RecursiveASTVisitor with a recursive function because I needed to pass information about checked scopes downward from compound statements to their children.

To avoid having to unnecessarily change tests in the Checked C repo, I generalized the code for constant-sized ranges to detect equivalence of base expressions beyond lexical equivalence.  It now canonicalizes base expressions by stripping off outer operations that do not change the runtime value of the expression.  This includes parentheses, value-preserving casts, and certain operations involving
function/array types.  For example, the address-of operator applied to an array does not actually change runtime values.  It gives you a pointer to the same location with a different type.

Testing:
- Updated Checked C repo tests with new warnings about that occur for bounds declarations in checked scopes.  There is a separate pull request for this: https://github.com/Microsoft/checkedc/pull/240
- For the clang Checked C repo, added a new test of bounds declarations that verifies that warnings are always produced for checked scopes.  It includes the existing test of bounds declarations, but uses a pragma to put it in a checked scope.
- Update tests with new warning messages.
- Passed automated testing on Linux; local testing on Windows.

### DIFF
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -9373,9 +9373,14 @@ def err_bounds_type_annotation_lost_checking : Error<
       : Error<"invalid bounds cast: expected _Array_ptr type">;
 
   def warn_bounds_declaration_invalid : Warning<
-    "declared bounds for %1 may be invalid after "
-    " %select{assignment|initialization}0">,
+    "cannot prove declared bounds for %1 are valid after "
+    "%select{assignment|initialization}0">,
     DefaultIgnore,
+    InGroup<CheckBoundsDecls>;
+
+  def warn_checked_scope_bounds_declaration_invalid : Warning<
+    "cannot prove declared bounds for %1 are valid after "
+    "%select{assignment|initialization}0">,
     InGroup<CheckBoundsDecls>;
 
   def error_bounds_declaration_invalid : Error<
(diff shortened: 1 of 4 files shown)


awruef:
To continue, it seems like what happens is that in some code, pointers to forward declared structures are used opaquely (i.e. no field elements of the structure are referred to) and despite the compilation unit not having access to information about the size of the forward declared structure, life was good. Now, even when a field of the record isn't used, the type checker wants to know the size of the pointed-to value, so it asks clang. 

For forward declared records, that might not exist. This is probably a contingency that checkedc-clang should account for. Perhaps the thing to do is to check if the current type for which you are requesting the size is forward declared. If it is, i.e. there is no layout/size information at all, then assume that it has a size of 1 (because all structs have to have at least a size of 1) and see if the proof goes through. In our example, and in many examples we can think of where pointers are used in this opaque manner, it would. 

Does that make sense?

dtarditi:
I think we should leave the size symbolic in that case.  If we have a forward declared struct S, we should use sizeof(struct S).   Another forward declared struct T will have sizeof(struct T).  sizeof(T) won't be equivalent to sizeof(S),, which is what we want.

awruef:
That sounds fine too, I was just trying to make it easier on you / make more programs typecheck :)

dtarditi:
I've opened issue #422 to track the work for handling symbolic sizes for Ptrs to incomplete types.

PR #441, opened 2018-01-08
Produce error messages for problems with returns.

dtarditi:
Clang checks for problems with return statements in C.  These problems include returning a value when none is expected, not returning a value when one is expected, and falling off the end of a function when a return value is expected.

These checks result in warnings.  Some of them are escalated to errors by default, but they can still be turned off by programmers. For Checked C, these problems must be errors when checked pointers and checked scopes are involved,  so that soundness guarantees for them  hold.  There should be errors when:
- A value is returned in a checked scope, when none is expected.
- The return value has a checked pointer type or bounds, and no value is returned when expected.

This modifies the existing code to issue error messages in these situations.  It cleans up SemaBounds.cpp to avoid returning unused boolean values when walking the AST while checking bounds and includes an empty function body for checking bounds on function return statements.

Testing:
- Added new tests cases to the Checked C repo for return statements.
- Passed automated testing.

### DIFF
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -565,6 +565,22 @@ def warn_suggest_noreturn_block : Warning<
   "block could be declared with attribute 'noreturn'">,
   InGroup<MissingNoreturn>, DefaultIgnore;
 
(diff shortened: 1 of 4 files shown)


dtarditi:
@awruef, I committed this change.  Please feel free to provide feedback after the fact.

PR #450, opened 2018-02-08
Fix bug with declaration context lookup.

awruef:
When I added logic to change variable lookup to be sensitive to whether or not you wanted the parameter or the global declaration, I didn't thread through this logic into the recursive calls of `getVariable`. So sometimes you ask for the parameter but instead you get the global. This fixes that by threading through the boolean flag indicating whether or not you want the parameter or the global.

### DIFF
--- a/tools/checked-c-convert/ProgramInfo.cpp
+++ b/tools/checked-c-convert/ProgramInfo.cpp
@@ -1051,11 +1051,12 @@ bool ProgramInfo::getDeclStmtForDecl(Decl *D, DeclStmt *&St) {
 // Returns true if E resolves to a constraint variable q_i and the
 // currentVariable field of V is that constraint variable. Returns false if
 // a constraint variable cannot be found.
+// ifc mirrors the inFunctionContext boolean parameter to getVariable. 
 std::set<ConstraintVariable *> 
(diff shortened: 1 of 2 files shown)


tools/checked-c-convert/ProgramInfo.cpp
```diff
@@ -1042,22 +1052,25 @@ bool ProgramInfo::getDeclStmtForDecl(Decl *D, DeclStmt *&St) {
 // currentVariable field of V is that constraint variable. Returns false if
 // a constraint variable cannot be found.
 std::set<ConstraintVariable *> 
-ProgramInfo::getVariableHelper(Expr *E, 
-  std::set<ConstraintVariable *> V, ASTContext *C) {
+ProgramInfo::getVariableHelper( Expr                            *E,
```
dtarditi: The formatting seems slight off - there seems to be an extra space.

dtarditi: Could you update the comment to describe what the ifc parameter means?

awruef: I added a longer comment at the declaration of `getVariable` that explains the purpose of the flag, and then cited that at `getVariableHelper`.

Issue #451, opened 2018-02-09
argument substitution not working properly in presence of implicit casts


dtarditi:
As part of checking bounds declarations, we substitute argument expressions for occurrences of parameter variables in bounds expressions.     We are using TreeTransform.h, which was designed for use with C++ templates and re-runs semantic analysis during substitution.   It skips implicit casts, assuming that they will be recomputed during semantic analysis.

There are differences in the implicit conversion rules used for arguments and those used at operators.   Argument expressions are subject to the same conversion rules as assignments, in particular pointers can be cast to/from void pointers,    In Checked C, bounds-safe interface conversions are also done at assignments, but not at operators.  Those implicit casts are not recomputed during semantic analysis when an argument expression is substituted for a parameter variable, causing the IR to be incorrectly typed.

dtarditi:
This is fixed.

PR #453, opened 2018-02-12
Fix subtle issues related to substitution of arguments for parameters

dtarditi:
This addresses issue #451.  During checking of bounds declarations, we need to check that the actual arguments meet the bounds constraints of the parameters.  The way we do this is by taking the bounds declaration for a parameter and substituting in the actual arguments to derive the desired bounds expression, in the context of the caller.  We then check that the actual argument implies the validity of this bounds expression.

It turns out that substituting an argument expression for a parameter variable is trickier than you might think in clang.  The issue is that the argument expression might be subject to implicit casts.   You can substitute the expression in for the parameter in your bounds expression, but if you call TreeTransform on the bounds expression, your implicit casts might get dropped.    The rules for implicit casts differ between call arguments/assignments and other operators.   This can lead to changes in the types of subexpressions of the bounds expressions, which could change their meaning or lead to illegally-typed expressions.  For example, an argument expression with void pointer type could be implicitly cast to some pointer to T.   The implicit cast could be lost, leading to a bounds expression with pointer arithmetic on void pointers, which is nonsensical.

For now, I'm taking the approach that we'll convert these implicit casts to explicit casts in the IR, so that these subtle problems don't happen.  Note that we can't keep all implicit casts during TreeTransform: it assumes that implicit casts are being resynthesized.

This change updates argument substitution to convert implicit casts that aren't the usual unary conversions to explicit casts. 

It also removes a version of GetTypeBasedBounds that can be dangerous to use.   The bounds inference code is generally expecting to work on expressions in an "expanded" form, where count/byte_count have been removed as sugar.   The removed version of GetTypeBasedBounds produces a count/byte_count, which could unexpectedly propagate if the caller isn't careful.

Finally, it puts the target bounds expressions for a parameter in the expanded standard form as soon as possible, when checking call arguments.  While I think this isn't strictly necessary because the checking of call arguments puts the desired target into the standard form, it is easier to reason about.

There's a hack in the code: we use CStyleCasts to represent the explicit casts.   However, when using TreeTransform to actually do substitution of arguments for parameters, this re-runs semantic checking on CStyleCasts.   This can cause an error when semantic checking is re-run in the context of a checked scope.  An explicit cast may be a bounds-safe interface that converts from a checked pointer to an unchecked pointer   We need to fix how we represent whether an operation is in a checked scope/unchecked scope, but that's a separate issue.  For now, we override the handling of CStyleCasts in the substitution class to not do semantic checking.

Testing:
- Passed local testing on Windows.
- Passed automated testing on Linux, including running the LNT tests and tests that have been converted to Checked C.
- Add new test that highlights the subtle issue around a dropped conversion.

### DIFF
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -2261,10 +2261,15 @@ namespace {
 
         // We have a bounds of the form  count(e) or byte_count(e).  Expand
         // the bounds to a standard form, using the current argument as the
-        // base expression. We are short circuiting a step here.  The full
-        // expansion would use the parameter represented as a symbolic index.
-        // In the step below, we would then substitute the actual argument for
-        // that parameter represented using an index.
+        // base expression.
+        //
+        // We are are short-ciructing a step here.  In expanding the parameter
+        // bounds, we could use the parameter represented as a symbolic index.
+        // This more properly represents the parameter bounds.  However, code
+        // below will eventually substitute Arg for the symbolic parameter.
+        // Instead of going to the trouble of building the symbolic parameter
+        // expression, which we will throw away soon anyway, we just use Arg
+        // here.
         if (ParamBounds->isElementCount() || ParamBounds->isByteCount())
           ParamBounds = S.ExpandToRange(Arg, const_cast<BoundsExpr *>(ParamBounds));
 
@@ -2286,7 +2291,7 @@ namespace {
         }
         else {
           // Concretize parameter bounds with argument expressions. This fails
-          // and return null if an argument expression is a modifying
+          // and returns null if an argument expression is a modifying
           // expression, but we've already issued an error about about that.
           BoundsExpr *SubstParamBounds =
             S.ConcretizeFromFunctionTypeWithArgs(
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -194,44 +194,6 @@ namespace {
       return SubstitutedModifyingExpression;
     }
 
-    // HACK - we don't want semantic analysis re-run on C style casts.  We may have synthesized
-    // them to make bounds-safe interfaces casts explicit in  the IR, and they could be illegal
-    // in a checked context.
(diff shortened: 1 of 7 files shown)


dtarditi:
I removed the hack in the code to prevent semantic analysis from running on C-style casts.    Instead, we only verify the requirements for C-style casts in checked scopes on the path from the parser.   For tree rewriting and other uses, we don't (re)-verify the requirements.   This is still kind of hacky, but better than before.
 
It turns out that first hack was more dangerous than I thought.  It dropped implicit casts, and didn't insert the needed unary casts.   This has passed automated testing on Linux.

lib/Sema/SemaBounds.cpp
```diff
@@ -230,6 +230,15 @@ BoundsExpr *Sema::ConcretizeFromFunctionTypeWithArgs(
...
+    llvm::outs() << "Failed concretizing\n";
+    llvm::outs() << "Bounds:\n";
+    Bounds->dump(llvm::outs());
+    int count = Args.size();
+    for (int i = 0; i < count; i++) {
+      llvm::outs() << "Dumping arg " << i << "\n";
+      Args[i]->dump(llvm::outs());
+    }
+    llvm::outs().flush();
     llvm_unreachable("unexpected failure in making function bounds concrete with arguments");
```
awruef: This debug printing will be noisy in release builds of the compiler (pretty sure that `llvm_unreachable` doesn't `abort` in release builds).

lib/Sema/SemaBounds.cpp
```diff
@@ -721,62 +740,40 @@ namespace {
...
-          ImplicitCastExpr *ICE = CreateImplicitCast(Ty, CastKind::CK_LValueToRValue, E);
-          ICE->setBoundsSafeInterface(IsBoundsSafeInterface);
-          Base = ICE;
-        } else
-          Base = E;
-        if (Ty->isVoidPointerType()) {
-          return ExpandToRange(Base, Context.getPrebuiltByteCountOne());
-        } else
-          return ExpandToRange(Base, Context.getPrebuiltCountOne());
+          ;
```
awruef: Is this `;` lost?

Issue #449, opened 2018-02-08
nt_array_ptr bounds-safe interface causes compiler crash


dtarditi:
The following program causes an internal compiler assert:
```
#pragma BOUNDS_CHECKED ON

void f(char *source : itype(nt_array_ptr<char>)) {
  *source;
}
```
There is an lvalue-to-rvalue cast in the inferred bounds for `*source`, where the source type `char *` and the destination type is `nt_array_ptr<char>`.   Code generation expects the types to match exactly (ignoring modifiers).

dtarditi:
This issue has been fixed.

PR #456, opened 2018-02-15
Fix compiler assert when using pointers with null-terminated interface types.

dtarditi:
This change fixes issue #449.  You can use the `itype` construct to specify that a pointer has an interface type that is a null-terminated pointer type: `char *p : itype(nt_array_ptr<char>)'.   The null-terminated pointer will have bounds of count(0), which means it can be read and written (only null values can be written).

Reading or writing via the pointer triggered an internal compiler assert.  The problem was that we were creating an lvalue-to-rvalue cast where the source and destination types were not equal (they differed in checkedness).  The fix is to use the same type for the lvalue-to-rvalue cast, and cast the resulting value.

This fix produces some chains of casts in error messages involving bounds declaration.  These were full of redundancy and hard to understand.   When creating an explicit cast, elide intermediate casts that don't change the value of an expression.  Use the same method we use during semantic analysis of bounds declarations.   This makes the (bounds) ranges described in error messages simpler and easier to use.

Also make sure that Ptr values are cast to Array_ptr when putting bounds in standard form.  We were expanded bounds to ranges that were doing pointer arithmetic on Ptr values.   When TreeTransform is run, it re-runs semantic analysis, which errors on those ranges.

Testing:
- Passed local testing on Windows x64
- Passed automating testing on Linux x64, including LNT.
- Add new regression test to the Checked C repo that tests that dynamic behavior is what is expected.  This is in a separate pull request  (https://github.com/Microsoft/checkedc/pull/259).

### DIFF
--- a/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/include/clang/Basic/DiagnosticSemaKinds.td
@@ -565,6 +565,22 @@ def warn_suggest_noreturn_block : Warning<
   "block could be declared with attribute 'noreturn'">,
   InGroup<MissingNoreturn>, DefaultIgnore;
 
+// Checked C-specific errors for falling off ends of functions
+def err_maybe_falloff_function_with_checked_return : Error<
+  "control may reach end of function with checked pointer return type">;
+def err_falloff_function_with_checked_return : Error<
+  "control reaches end of function with checked pointer return type">;
+
+def err_maybe_falloff_checked_function : Error<
+  "control may reach end of checked function body">;
+def err_falloff_checked_function : Error<
+  "control reaches end of checked function body">;
+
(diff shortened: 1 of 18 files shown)


dtarditi:
@awruef  - good question..  The bounds-safe interface flag is set to true for compiler-introduced casts from unchecked pointers to checked pointers.  These aren't normally allowed in checked scopes. The flag indicates that the compiler introduced the cast according to bounds-safe interface rules, not the programmer.   This will let us check that "unsafe" casts  in checked scopes haven't been accidentally let through by earlier compiler phases (we haven't written this checker yet).

The casts that you asked about are either to checked pointer types or between checked pointer types.  Those kinds of casts are always allowed, so the flag can be set to false.   I added comments to clarify this.

Issue #443, opened 2018-01-19
incorrect error message regarding bounds-safe interface declared in a checked scope


dtarditi:
A programmer has reported that this code:
```
        #pragma BOUNDS_CHECKED ON

        void
        myfunc(char *buf : bounds(buf, end),
               char *end : itype(_Array_ptr<char>))
        {
                /* … */
        }
````
results in an incorrect error message:
```
        test.c:4:27: error: parameter used in a checked scope must have a checked type or a bounds-safe interface
        myfunc(char *buf : bounds(buf, end),
                                  ^
        test.c:4:8: note: parameter declared here
        myfunc(char *buf : bounds(buf, end),
                    ^
        1 error generated.
```

dtarditi:
This should be fixed now.  Let me know if you see any issues.

Issue #445, opened 2018-01-31
Check for initializers on struct/array variables


dtarditi:
If a variable has a member or element with checked pointer type that may be used to access memory, it must be initialized when declared.   We currently aren't checking this.  We are checking that variables that have check pointer types that may be used to access memory have initializers.

dtarditi:
@sxl463 implemented this in PR #506.

PR #461, opened 2018-03-13
Allow both interop types and bounds expressions in bounds-safe interfaces.

dtarditi:
This change extends the Checked C version of clang to allow both  interop types and bounds expressions in a bounds-safe interfaces.  Before, we allowed only one or the other, which didn't provide proper support for pointer types with multiple levels of pointers.  A bounds expression and an interop type expression can now appear after the colon, in any order.   For example, a programmers can write `int f(int **p : itype(array_ptr<ptr<int>>) count(len, int len)`.    This change addresses issue #445 and also fixes issue #443.

Within the IR, we were storing both bounds expression and interop types using a single field in declarations and function t ypes.  We also stored interop types as bounds expressions, even though they were really a different AST node.  The change replaces these occurrences of `BoundsExpr *` with the value class `BoundsAnnotations`.   It took some gyrations before I arrived at this simple representation choice.   This abstraction let us add additional information such as where clauses later.   Also, InteropTypeBoundsAnnotation is no lo longer a subclass of BoundsExpr.  It is a distinct class that inherits directly from Expr.   The changes led to a large amount of boilerplate changes: replacing occurrences of BoundsExpr *, modifying the AST class structure, and so on.

The changes  also include:
1.  Modifying the code that adds bounds expressions/interop type to AST nodes to infer bounds expressions and interop types when one of them is missing.  We now always infer an interop type from the existence of a bounds expression (this change fixes the causes of issue #443).  We infer bounds for checked parameter arrays and nt_array_ptrs from types or interop types.  (there are still declarations where we don't infer bound by default when perhaps we could).   Before, we could not infer bounds for interop types because we had no place to put them.
2.  Revising the bounds inference code to handle the fact that both may appear now. We try use bounds expressions before trying to infer bounds based on the interop type.
3.  Improving the error messages generated for more conflicts between interop types on redeclarations.
4, Centralizing the code for parsing what I now call "bounds annotations" (a bounds expression and a bounds annotation).   This change simplifies the code, as we no longer have duplicated logic for deferred parsing of bounds expressions.
5. Checked that bounds expressions aren't declared for interop types that are ptr types.

I also updated some code in the Checked C rewriter that was accessing interop type information.

Testing:
- Modify existing clang Checked C tests for revised error messages and AST dumps from the compiler.
- Adding new tests of bounds expressions/interop types to the Checked C repo.  That will be covered by a separate PR.
- Removing a workaround in the Checked C LLVM test suite for KSDdist.  That will be covered by a separate PR
- Pass revised Checked C clang tests, Checked C regression tests, and LNT testing


lib/Sema/TreeTransform.h
```diff
@@ -5228,11 +5230,47 @@ bool TreeTransform<Derived>::TransformFunctionTypeParams(
...
+  BoundsAnnotations &Annot, bool &Changed) {
+  BoundsExpr *ExistingBounds = Annot.getBoundsExpr();
+  BoundsExpr *NewBounds = ExistingBounds;
+  if (ExistingBounds) {
+    ExprResult Result = getDerived().TransformExpr(ExistingBounds);
+    if (Result.isInvalid())
+      return true;
+    NewBounds = dyn_cast<BoundsExpr>(Result.get());
+    if (!NewBounds) {
+      assert("unexpected dynamic cast failure");
```
awruef: Should this be something like `llvm_unreachable`? Is this a recoverable error, or something you never expect to see, since asserts will be compiled out in release mode.

lib/Sema/TreeTransform.h
```diff
@@ -5228,11 +5230,47 @@ bool TreeTransform<Derived>::TransformFunctionTypeParams(
...
+
+  InteropTypeExpr *ExistingIType = Annot.getInteropTypeExpr();
+  InteropTypeExpr *NewIType = ExistingIType;
+  if (ExistingIType) {
+    ExprResult Result = getDerived().TransformExpr(ExistingIType);
+    if (Result.isInvalid())
+      return true;
+    NewIType = dyn_cast<InteropTypeExpr>(Result.get());
+    if (!NewIType) {
+      assert("unexpected dynamic cast failure");
```
awruef: Same here.

lib/Parse/ParseDecl.cpp
```diff
@@ -2298,19 +2298,19 @@ Decl *Parser::ParseDeclarationAfterDeclaratorAndAttributes(
   // declarators already.
   if (getLangOpts().CheckedC && isa<VarDecl>(ThisDecl)) {
     VarDecl *ThisVarDecl = dyn_cast<VarDecl>(ThisDecl);
```
awruef: Check `ThisVarDecl` for `null`?

lib/Sema/TreeTransform.h
dtarditi: Good catch.  No, this isn't a recoverable error.  I'll change this and the other assert to be unreachable.

lib/Sema/TreeTransform.h
dtarditi: There's an isa<VarDecl> test in the enclosing test, so we don't need to check for null here.

PR #468, opened 2018-03-21
Remove parsing support for old bounds cast expression syntax.

dtarditi:
We changed bounds cast expressions to take a bounds expression as the
second argument, instead of implicitly inferring the kind of bounds
expression from the number of arguments.  This is more verbose but
clearer.  Bounds cast expressions now have two forms:

Op<T>(e1)
Op<T>(e1, bounds-expression)

where Op is one of _Assume_bounds_cast or _Dynamic_bounds_cast
and T must be a pointer type.

This change also improves the error messages and renames the
error message names to be clearer.

Testing:
- There will be a corresponding Checked C repo change to tests
  for bounds cast expressions.
- Add a test for error recovery from bounds cast expression syntax
  errors.
- Passed local testing on x64 Windows and automated testing on Linux.

### DIFF
--- a/lib/Sema/SemaExpr.cpp
+++ b/lib/Sema/SemaExpr.cpp
@@ -13556,9 +13556,9 @@ ExprResult Sema::ActOnBoundsCastExprCount(
   if (!DestTy->isCheckedPointerArrayType()) {
     Diag(TypeLoc, diag::err_bounds_cast_error_with_array_syntax);
     return ExprError();
-  } else if ((E1->getType())->isVoidPointerType()) {
-    Diag(E1->getLocStart(),
-         diag::err_typecheck_void_pointer_count_return_bounds);
+  } else if (DestTy->isVoidPointerType()) {
+    Diag(OpLoc,
+         diag::err_typecheck_void_pointer_count_bounds_cast);
     return ExprError();
   } else {
     if (!ResE2.isInvalid())
--- a/lib/Sema/SemaExpr.cpp
+++ b/lib/Sema/SemaExpr.cpp
@@ -13465,170 +13465,59 @@ bool Sema::CheckBoundsCastBaseType(Expr *E1) {
 ExprResult Sema::ActOnBoundsCastExprBounds(
     Scope *S, SourceLocation OpLoc, tok::TokenKind Kind,
     SourceLocation LAngleBracketLoc, ParsedType D,
-    SourceLocation RAngleBracketLoc, RelativeBoundsClause *RelativeClause,
+    SourceLocation RAngleBracketLoc,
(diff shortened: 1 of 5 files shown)


lib/Sema/SemaExpr.cpp
```diff
@@ -13465,170 +13465,59 @@ bool Sema::CheckBoundsCastBaseType(Expr *E1) {
...
+    SourceLocation RAngleBracketLoc,
     SourceLocation LParenLoc, SourceLocation RParenLoc, Expr *E1) {
-  TypeSourceInfo *castTInfo;
-  BoundsExpr *bounds;
+  TypeSourceInfo *CastTInfo;
+  BoundsExpr *Bounds;
 
-  QualType DestTy = GetTypeFromParser(D, &castTInfo);
-  SourceLocation TypeLoc = (castTInfo->getTypeLoc()).getBeginLoc();
+  QualType DestTy = GetTypeFromParser(D, &CastTInfo);
```
awruef: I looked at the code for `GetTypeFromParser` and I think that either way, it will set `CastTInfo` to `nullptr`, so it's okay that it's declared un-initialized. Maybe there should be an `assert` before it's de-referenced, though?

lib/Sema/SemaExpr.cpp
```diff
@@ -13465,170 +13465,59 @@ bool Sema::CheckBoundsCastBaseType(Expr *E1) {
...
+    BoundsExpr *Bounds) {
+  TypeSourceInfo *CastTInfo;
 
-  RangeBoundsExpr *Range = nullptr;
-  TypeSourceInfo *castTInfo;
-
-  QualType DestTy = GetTypeFromParser(D, &castTInfo);
-  ExprResult ParsedBounds = CorrectDelayedTyposInExpr(bounds);
-  SourceLocation TypeLoc = (castTInfo->getTypeLoc()).getBeginLoc();
+  QualType DestTy = GetTypeFromParser(D, &CastTInfo);
```
awruef: I looked at the code for `GetTypeFromParser` and I think that either way, it will set `CastTInfo` to `nullptr`, so it's okay that it's declared un-initialized. Maybe there should be an `assert` before it's de-referenced, though?

lib/Sema/SemaExpr.cpp
dtarditi: Good catch.  We should check that CastTInfo is non-null and use a different source location if it is.  I'll use the location of the '<' symbol given an expression of the form `Op<T>(...)`, where `Op` is a bounds cast operator.  It'll point the programmer in the right direction, although we could be off by any white space between the '<' and T.

Issue #475, opened 2018-04-10
Decide on typing rules for variables with bounds-safe interfaces in bounds expressions


dtarditi:
When inferring bounds for a variable with a bounds-safe interface, the inferred bounds always use the original unchecked type of the variable.   This is problematic for two reasons.  First, in a checked scope, the computations of the bounds themselves should use checked types and operations.   Second, it causes problems for checking of bounds declarations.  Comparisons of canonicalized bounds expressions to fail sometimes because the types of operations differs.   This happens when a bounds expression is declared in a scope whose checkedness differs from the scope where the variable is used.  For example:
```
extern void check_nullterm_call_bsi(int *arg1 : itype(Nt_array_ptr<int>),
                                    _Nt_array_ptr<int> arg3) {
  _Checked{
    arg1 = arg3;
  }
}
```
As a work around when comparing bounds expressions, when we compare expressions, we ignore checkedness when comparing types. 

This issue is to decided on he typing rules hen a variable with a bounds-safe interface is used in an inferred bounds expression.  At the very least, bounds inferred for a variable in a checked scope should use any variables at their bounds-safe interface types.   This makes everything consistent in checked scopes.

This leaves the question of what to do in unchecked scopes.   Here, things become trickier because unchecked pointer operations have potentially undefined behavior, while checked pointer operations do not. 

The approach we've taken for checked scopes vs. unchecked scopes is that an unchecked scope that uses only checked types should act the same as a checked scope.     This suggests the following rule at bounds-safe interface assignments (where a checked pointer is assigned to a variable with an unchecked pointer type and a bounds-safe interface): 
- any variables in the inferred bounds for the LHS should be used at their bounds-safe interface types.

The bounds of the checked pointer for the RHS would have to imply the validity of the bounds for the LHS, which is what we want.  

I need to think about what we should for the RHS some more.  My initial thought is that when we infer  bounds for the RHS, we should likely treat any variables with bounds-safe interfaces, as being at the bounds-safe interface types.

Issue #474, opened 2018-04-07
Extend bounds declaration checking to understand equality.


dtarditi:
The compiler statically checks that bounds declarations are valid.  Bounds declarations must be implied by existing bounds declaration and other information in the environment.    Currently, the compiler doesn't understand that assignment and initialization implies facts about equality of expressions that can be used during checking of bounds declarations.   

This causes the compiler to issue spurious warnings, and is why I've left checking of bounds declarations turned off for unchecked scopes.

This work item is to extend the compiler to collect quality information for individual assignment statements (or declarations with initializers) and use it during checking of bounds declarations.  For example, given
```
array_ptr<int> b : count(5) = ...
array_ptr<int> c : count(5) = b;
```
we are trying to prove for the declaration of `c` that the bounds of `b` imply 
```
bounds(c, c + 5)
```
is true.   The bounds of b is `bounds(b, b + 5)`.    If we know that `c == b`, it is simple to conclude that `bounds(b, b + 5)` implies `bounds(c, c + 5)` is valid.

It is future work to extend this checking to gather and use equality facts across statements/declarations.

dtarditi:
Work is complete.

Issue #472, opened 2018-04-04
Always warn when compiler can't prove bounds declarations are valid


dtarditi:
The compiler attempts to prove that bounds declaration valid.  Right now, the compiler only warns in checked scopes when it can't prove bounds declarations to be valid.   The compiler should always warn, even in unchecked scopes.

dtarditi:
Work is complete.

PR #476, opened 2018-04-10
Use simple facts about expression equality.  Expand checking of bounds declarations.

dtarditi:
This change extends the compiler to collect and use simple facts about expression equality when checking bounds declarations.  It also turns on checking of bounds declarations for unchecked scopes by default.  It was only on by default for checked scopes.  The checking of bounds declarations is still simple, so it may produce (many) unnecessary warnings.   It can be turned off using  `-Wno-check-bounds-decls` (everywhere), `-Wno-check-bounds-decl-unchecked-scope` (for unchecked scopes), and `-Wno-check-bounds-decls-checked-scope` (for checked scopes), and.  This change addresses issues #474 and #472.

Equality facts are gathered at assignments and initializers.  They are not propagated across statements. For example, given `x = y`, when checking the bounds for `x` are implied by the bounds for `y`, we assume `x == y`.  In general, given `e1 = e2`, we check that `e1` and `e2` are non-modifying expressions and collect the fact that `e1 == e2`.   (We still need to check that the assignment doesn't modify an lvalue used by` e2`, and will add that in the future.  We may miss issuing warnings that we should issue,)
- We pass a list of sets of equivalent expressions to CanonBounds, which checks whether two expressions compute the same value.  We try to compare expressions and if that fails, we consult the list of sets.  We check each expression to see if it is equivalent to an expression in a set (we don't recursively use the equality information).
- When setting up the equality fact `e1 == e2`, we add an lvalue-to-rvalue cast for the IR for `e1`, .   Given a declarator for a variable v with an initializer, we make sure to turn v into a lvalue-to-rvalue cast of a decl-ref.
- There was code in SemaBounds.cpp that puts an expression into a standard form by ignoring non-value changing operations.   That code is moved to CanonBounds.cpp.
- The types of subexpressions of an expression affect the meaning of an expression.  When comparing two expressions, check that the types of their subexpressions match.  This wasn't a concern before when the code was strictly comparing lexical  equality.  There is a TODO for work needed for bounds-safe interfaces, tracked by Github issue #475.

The change fixes a few bugs found during testing of the code for checking equivalence of expressions:
- CanonBounds.cpp needs to treat implicit and C-style casts as being equivalent if they cast to/from the same type.
- In SemaBounds.cpp, when we infer the rvalue bounds of a bounds cast expression of the form *_bounds_cast<T>(`e1`, bounds-expr), the bounds need to be for `(T) e1`, not `e1` (`e1` has the wrong type).
- In SemaExpr.cpp, when type checking bounds(`e1`, `e2`), we need to check that e1 and e2 point to compatible types, not exactly the same type.

Testing:
- Updated the Checked C repo tests.  This will be covered by a separate pull request:
  - Added more tests of expression equivalence (lexical_equality should be renamed, as it checks for expression equivalence using a few additional facts about C expressions).
  - Updated tests to handle the fact that checking of bounds declaration is on everywhere by default.  Replace a bunch of TODO's with an expected warning message. Also deleted some warnings that are no longer issued because the compiler can prove the bounds are valid.
  - Kept checking of bounds declarations on for most tests and only disabled it for some parsing and typechecking tests.
  - In some cases, made simple changes to avoid warnings about bounds declarations (the original cases are captured in the new Checked C clang repo file test/CheckedC/static-checking/bounds-decl-challenges.c)
- Update the Checked C clang tests:
  - In bounds-decl-checking.c, remove expected warnings that are now gone because the compiler understand equality facts.  Add code that checks the compiler understands equality facts involving arrays.  Include a negative test where the compiler concludes that an array is too small for the declared bounds for an array_ptr variable.
- Passed automated testing on Linux x64 (clang regression tests and LNT testing)

### DIFF
--- a/lib/AST/CanonBounds.cpp
+++ b/lib/AST/CanonBounds.cpp
@@ -24,8 +24,8 @@
 using namespace clang;
 using Result = Lexicographic::Result;
 
-Lexicographic::Lexicographic(ASTContext &Ctx, EqualityRelation *EV) :
-  Context(Ctx), EqualVars(EV), Trace(false) {
+Lexicographic::Lexicographic(ASTContext &Ctx, EquivExprLists *EquivExprs) :
+  Context(Ctx), EquivExprs(EquivExprs), Trace(false) {
 }
 
 Result Lexicographic::CompareInteger(signed I1, signed I2) {
@@ -205,7 +205,7 @@ Result Lexicographic::CompareExpr(const Expr *E1, const Expr *E2) {
    Stmt::StmtClass E2Kind = E2->getStmtClass();
    Result Cmp = CompareInteger(E1Kind, E2Kind);
    if (Cmp != Result::Equal)
-     return Cmp;
+     goto different;
 
    Cmp = Result::Equal; 
    switch (E1Kind) {
@@ -274,7 +274,7 @@ Result Lexicographic::CompareExpr(const Expr *E1, const Expr *E2) {
    }
 
    if (Cmp != Result::Equal)
-     return Cmp;
+     goto different;
 
    // Check children
    auto I1 = E1->child_begin();
@@ -287,7 +287,7 @@ Result Lexicographic::CompareExpr(const Expr *E1, const Expr *E2) {
      if (ordered) {
        if (Cmp == Result::Equal)
          continue;
-       return Cmp;
+       goto different;
       }
      assert(E1Child && E2Child);
      const Expr *E1ChildExpr = dyn_cast<Expr>(E1Child);
@@ -296,7 +296,7 @@ Result Lexicographic::CompareExpr(const Expr *E1, const Expr *E2) {
      if (E1ChildExpr && E2ChildExpr) {
        Cmp = CompareExpr(E1ChildExpr, E2ChildExpr);
        if (Cmp != Result::Equal)
-         return Cmp;
+         goto different;
      } else
        // assert fired - return something
        return Result::LessThan;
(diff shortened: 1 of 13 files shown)


lib/AST/CanonBounds.cpp
```diff
@@ -17,15 +17,127 @@
...
+  }
+  // Ignore operations that don't change runtime values: parens, some cast operations,
+  // array/function address-of and dereference operators, and address-of/dereference
+  // operators that cancel (&* and *&).
+  //
+  // The code for casts is adapted from Expr::IgnoreNoopCasts, which seems like doesn't
+  // do enough filtering (it'll ignore LValueToRValue casts for example).
+  // TODO: reconcile with CheckValuePreservingCast
+  static Expr *IgnoreValuePreservingOperations(ASTContext &Ctx, Expr *E) {
+    while (true) {
```
awruef: Maybe this could be written in an recursive style? If it works, that's probably best, and there's no reason to re-write it in a recursive style, so there's almost no purpose in leaving this note, except to say that I saw this and appreciated the "get it done" nature of the code but was a little sad at the repeated `continue` paradigm to recur down the `Expr`. I will get over being sad, though.

lib/Sema/SemaBounds.cpp
```diff
@@ -2088,9 +2015,37 @@ namespace {
...
+          TargetTy = S.getASTContext().getArrayDecayedType(D->getType());
+        } else {
+          Kind = CK_LValueToRValue;
+          TargetTy = D->getType();
+        }
+        Expr *TargetExpr = BoundsInference(S).CreateImplicitCast(TargetTy, Kind, TargetDeclRef);
+        EqualExpr.push_back(TargetExpr);
+        EqualExpr.push_back(Src);
+        EquivExprs.push_back(&EqualExpr);
+        /*
```
awruef: Maybe remove this commented-out block or un-comment it and guard it with `Trace` or something?

lib/AST/CanonBounds.cpp
dtarditi: The style was just from the original code that we started from.  I agree that a tail-recursive function would be better.    The continue style is error-prone.   We'll likely extend this code, at which point I'll convert it.

lib/Sema/SemaBounds.cpp
dtarditi: Yes, I think this information should be available under a flag.    I'll do that in a future PR because I think we want to put some effort into presenting the information.

We have a challenge when the compiler can't prove something true or false of letting the programmer understand what was going on.  When I was looking at the results of bounds declaration checking, for situations where there was a warning, it was helpful to know what equality facts the compiler knew.   I think we want an explanatory mode that tells the programmer what facts the compiler knows and what it tried to prove.   The compiler could issue additional notes including this information.

PR #485, opened 2018-04-26
Rename BOUNDS_CHECKED pragma to CHECKED_SCOPE.

dtarditi:
- Rename pragma and associated methods within the compiler.
- Update tests
- There are corresponding PRs for Checked C and the Checked C LNT repos: https://github.com/Microsoft/checkedc/pull/282 and https://github.com/Microsoft/checkedc-llvm-test-suite/pull/86.

Testing:
- Passed local testing on Windows and automated testing on Linux.


dtarditi:
@nmeum, flagging that this compiler change will require changes in your code base.

PR #498, opened 2018-06-11
Clean up code in CheckBoundsDeclarations class.

dtarditi:
- There's some code that appears to be unnecessary in VisitVarDecl. We should remove it.  VisitCastExpr should already be taking care of this.
- In addition, add comments describing the the difference between the Visit* and Traverse* methods. It's not immediately obvious what the difference is.
- Rename TraverseVarDecl to TraverseTopLevelVarDecl for clarity.

Testing:
- Passed local testing on Windows.
- Passed automated testing on x64 Linux: debug clang and LNT tests.

### DIFF
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -2150,6 +2150,11 @@ namespace {
       PointerWidth(S.Context.getTargetInfo().getPointerWidth(0)),
       ReturnBounds(ReturnBounds) {}
 
+    // Traverse methods iterate recursively over AST tree nodes, visiting all
+    // children of the node too.
+    //
+    // Visit methods do work on individual nodes, such as checking bounds
+    // declarations or inserting bounds checks.
     void TraverseStmt(Stmt *S, bool InCheckedScope) {
       if (!S)
         return;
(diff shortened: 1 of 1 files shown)


lib/Sema/SemaBounds.cpp
```diff
@@ -2762,8 +2765,10 @@ void Sema::CheckFunctionBodyBoundsDecls(FunctionDecl *FD, Stmt *Body) {
 }
 
 void Sema::CheckTopLevelBoundsDecls(VarDecl *D) {
-  if (!D->isLocalVarDeclOrParm())
-    CheckBoundsDeclarations(*this, nullptr).TraverseVarDecl(D, getCurScope()->isCheckedScope());
+  if (!D->isLocalVarDeclOrParm()) {
+    auto Checker = CheckBoundsDeclarations(*this, nullptr);
```
sxl463: Do we need to update the existed code right after this change? Just like this snippet: 
void Sema::CheckTopLevelBoundsDecls(VarDecl *D) {
  if (!D->isLocalVarDeclOrParm())
    CheckBoundsDeclarations(*this, nullptr).TraverseVarDecl(D, getCurScope()->isCheckedScope());
}

dtarditi:
Thanks for the CR - I removed the use of auto and just used an initializer instead.

PR #499, opened 2018-06-12
Add initialization checking for array/struct/union variables with checked pointer members

sxl463:
If a variable has a member or element with checked pointer type that may be used to access memory, it must be initialized when declared. We add initialization checking for array/struct/union variable declarations.

Testing:

Passed local testing on Windows.
Passed automated testing on 
DevTest Debug x64 Linux
DevTest Debug x64 Windows
DevTest Debug X86 Windows 
DevTest Release Linux X64 LNT

### DIFF
--- a/include/clang/AST/Type.h
+++ b/include/clang/AST/Type.h
@@ -1917,7 +1917,7 @@ public:
   /// \brief Whether this type is or contains a checked type
   bool isOrContainsCheckedType() const;
 
-  /// \brief Whether this type(Struct/Union) contains a checked value
+  /// \brief check whether an object of a struct or union type contains a checked value
   bool containsCheckedValue() const;
 
   /// \brief Whether this type is or contains an unchecked type.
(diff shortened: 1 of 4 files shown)


lib/AST/Type.cpp
```diff
@@ -4109,6 +4109,62 @@ bool Type::isOrContainsUncheckedType() const {
...
+  }
+  case Type::ConstantArray:
+  case Type::DependentSizedArray:
+  case Type::IncompleteArray:
+  case Type::VariableArray: {
+
+    return current->getPointeeOrArrayElementType()->containsCheckedValue();
+  }
+
+  case Type::FunctionProto: {
```
dtarditi: Function types should not be included.   We are checking whether a checked pointer value is stored in an object in memory.  A function type may have a checked pointer type as an argument or a return value.  That doesn't mean that a checked pointer value is stored in an object in memory.

include/clang/AST/Type.h
```diff
@@ -1917,6 +1917,9 @@ class Type : public ExtQualsTypeCommonBase {
   /// \brief Whether this type is or contains a checked type
   bool isOrContainsCheckedType() const;
 
+  /// \brief Whether this type(Struct/Union) contains a checked value
```
dtarditi: I would word this slightly more precisely: check whether an object of a struct or union type contains a checked value.

Issue #502, opened 2018-06-13
Check bounds declarations after assignments to variables used in bounds declarations.


dtarditi:
An expression may assign to a variable used in a bounds declaration:
```
f(_Array_ptr<int> p : count(len), int len) {
   len = …;
}
```
We need to check that bounds declarations are still valid after assignment expressions that contain these assignments.

There are three parts to this work:
1. Determine when an expression assigns to a variable used in an in-scope bounds declaration.
2. For each expression statement, determine the set of bounds declarations that must be checked and facts relevant to checking (such as equality) that are true at the end of the statement.
3. Check that the bounds declarations are true using the relevant facts.

Analysis of facts in the presence of complicated intra-expression control-flow (#206) is beyond the scope of this work item.

kkjeer:
This was implemented by #853.

PR #503, opened 2018-06-14
Expressions that modify variables used in bounds declarations.

dtarditi:
We are currently checking that bounds declarations are valid after assignments to variables with bounds declarations.  We also need to check that bounds declarations are valid after assignments to variables _used_ in bounds declarations.   This work is described in issue #502.

This change implements the first part of the work in issue #502.   It computes a mapping from expressions that modify variables to sets of in-scope bounds declarations that use the modified variables.

Determining whether an expression modifies a variable used by an in-scope bounds declaration is a lexically-scoped computation     When checking bounds declarations, we are currently traversing the AST in a lexically-scoped fashion.  We plan to change the traversal of the AST to be based on control flow, so that we can incorporate dataflow facts.   It isn't easy in clang to determine the lexical scope that an expression occurs in.   Clang doesn't have back pointers from expressions to their parent expressions or from statements to enclosing lexical scopes.  That makes it difficult to answer the question of what set of bounds declarations are in scope.  To deal with this, we compute a map that can be consulted at any point in a function.

We compute the information for each function, but do not use it.  For compilation units that don't use Checked C extensions, the maps will be empty.  We want to get coverage on the different kinds of ASTs that we might encounter, so the code is left on during bounds declaration checking.

Testing:
- I wrote a new set of Checked C clang tests that contain expressions that modify variables used in bounds declarations.  I then used this to dump the results of the analysis and manually inspect that the analysis is producing expected results.  The compiler will eventually produce error messages in some cases.   These are in tests/CheckedC/static-checking/bounds-side-effects.c.
- Passes existing Checked C and Checked C clang tests on Windows.
- Passed debug Linux clang tests.
- Other testing in progress.


lib/Sema/CheckedCAlias.cpp
```diff
@@ -333,3 +336,300 @@ void Sema::TrackMemberBoundsDependences(FieldDecl *FD, BoundsExpr *BE) {
...
+namespace {
+// Update map from variables to bounds expressions that use those variables.
+class UpdateDependences : public RecursiveASTVisitor<UpdateDependences> {
+private:
+   bool Add;              // whether to add or remove dependence.
+   VarDecl *BoundsDecl;   // variable with bounds declaration.
+   Sema::BoundsDependencyTracker::DependentMap &Map;
+
+   void AddDependence(VarDecl *D) {
+    if (D == BoundsDecl)   // Do not add self-dependences.
```
sxl463: Just a personal view, I prefer letting constant variables be left values, like:
BoundsDecl == D;
By doing this we can avoid the equal sing missing bug, buggy code like D = BoundsDecl will never appear again.

lib/Sema/SemaBounds.cpp
```diff
@@ -2759,6 +2759,15 @@ namespace {
 }
 
 void Sema::CheckFunctionBodyBoundsDecls(FunctionDecl *FD, Stmt *Body) {
+  ModifiedBoundsDependencies Tracker;
+  // Compute a mapping from expressions that modify lvalues to in-scope bounds
+  // declarations that depend upon those expressions.  We plan to change
+  // CheckBoundsDeclaration to traverse function bodies in an order determined
```
sxl463: Here "traverse function bodies" sounds like an inter-procedural process.

lib/Sema/SemaDecl.cpp
```diff
@@ -12922,13 +12922,19 @@ void Sema::ActOnBoundsDecl(DeclaratorDecl *D, BoundsAnnotations Annots,
...
   }
 
+  D->setBoundsExpr(getASTContext(), BoundsExpr);
+  D->setInteropTypeExpr(getASTContext(), IType);
+
   // if this is a member bounds dedclaration, update information mapping
   // members to what bounds declarations depend upon them.
   if (FieldDecl *FD = dyn_cast<FieldDecl>(D))
     TrackMemberBoundsDependences(FD, BoundsExpr);
+  else if (VD && !VD->isLocalVarDeclOrParm()) {
```
sxl463: LLVM code uses "Param" as the abbreviation for parameter. I think we can update "Parm" to "Param". Sounds better.

dtarditi:
I have addressed your code review feedback:
- Made a class member const and changed a comparison to have the const value on the left-hand side.
- Changed the wording of a comment for clarity.

PR #510, opened 2018-06-29
Fix compiler build warnings

dtarditi:
I noticed that we had compiler warnings during the compiler build on Linux.  This fixes the warnings:

Testing:
- Passed automated testing on Linux x64.


lib/Sema/CheckedCAlias.cpp
```diff
@@ -390,7 +389,7 @@ class UpdateDependences : public RecursiveASTVisitor<UpdateDependences> {
   UpdateDependences(VarDecl *BoundsDecl,
                     Sema::BoundsDependencyTracker::DependentMap &Map,
                     bool Add) :
-    BoundsDecl(BoundsDecl), Map(Map), Add(Add) {}
+    Add(Add), BoundsDecl(BoundsDecl), Map(Map) {}
```
sxl463: Why do we need to change the initialization order here?

dtarditi: There is a gcc compiler warning about fields not being initialized in the same order that they were declared.   The idea is to make it easier for programmers to read code that does initialization and to notice if initializers have been forgotten.

lib/Sema/CheckedCAlias.cpp
```diff
@@ -413,11 +412,11 @@ void Sema::BoundsDependencyTracker::Add(VarDecl *D) {
   UpdateDependences(D, Map, true).TraverseStmt(BE);
 }
 
-int Sema::BoundsDependencyTracker::EnterScope() {
+unsigned Sema::BoundsDependencyTracker::EnterScope() {
```
sxl463: I noticed we use "unsigned int" somewhere(e.g.  line 4819 in Sema.h), while in some other places we use "unsigned", do we need to keep unification?

Further, do we really need to globally change int to unsigned, even for some return types? What will happen if the programmers use the return values of those APIs for comparison?
For example, in the following program:
int main(){
	int a = -1;
	unsigned int b = 1;
	if(a > b)
		printf("a > b, a = %d, b = %u\n", a, b);  // the compiler will do (unsigned int) (a) first
	else
		printf("a <= b, a = %d, b = %u\n", a, b);
	exit(0);
}

I remembered this warning long long ago when I started learning C++ so I won't use them until necessarilly
"Avoid using unsigned ints in C and C++ unless you're dealing with raw memory contents or performing bit manipulations like shifting or masking."  cited from the following article:
http://soundsoftware.ac.uk/c-pitfall-unsigned.html

dtarditi: There are differing philosphies about signed vs. unsigned.   When working in a large existing code base, we should stick with the practice of that code base.

clang and LLVM make liberal use of unsigned types.    I think they use the philosophy that if an integer can't be negative, it should be unsigned.   I changed to use unsigned because (1) `size()` returns an unsigned number, which caused a compiler warning and (2) the value can't be  negative.
 
I agree that I should stick with `unsigned` and avoid `unsigned int`.    Existing clang code uses `unsigned`.

sxl463: This explanation makes sense to me.

PR #511, opened 2018-06-29
Add usage example and instructions to README

AnnaKornfeldSimpson:
Based on Andrew's email and my steps getting things working in Ubuntu, I added instructions and a usage example to the README documentation. This only covers producing `compile_commands.json` from cmake configuration although Andrew's email mentioned some potential other options.

### DIFF
--- a/tools/checked-c-convert/README.md
+++ b/tools/checked-c-convert/README.md
@@ -26,7 +26,7 @@ CMake's configuration tool can provide a `compile_commands.json` file listing th
 
 ### Utility to generate source list from compile_commands
 There's a small Python2 script in tools/checked-c-convert/utils that will take the `compile_commands` and generate a `checked-c-convert` command line invocation that contains all the source files needed as arguments. Run it as:
-`python2.7 run.py path/to/compile_commands.json ../checked-c-convert`
+`python2.7 run.py path/to/compile_commands.json checked-c-convert`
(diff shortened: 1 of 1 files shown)


tools/checked-c-convert/README.md
```diff
@@ -17,8 +17,16 @@ manually specified files with a manually specified command line, or, using
 a `compile_commands.json` compilation database. 
 
 ### Command Line
+Invoke `checked-c-convert` on the command line as:
+
+`$ checked-c-convert -p path/to/directory/containing/compile_commands.json path/to/source/file/to/convert.c`
 
 ### `compile_commands.json` database
+CMake's configuration tool can provide a `compile_commands.json` file listing the instructions for the compiler during build. Produce this with `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` added to the configuration call to cmake.
+
```
dtarditi: I have a wording suggestion. I think it'd be more precise to say that "CMake's configuration tool can generate".   I think it'd also e helpful to note that as of CMake 3.12, this is only supported by the Makefile and Ninja generators for CMake (https://cmake.org/cmake/help/v3.12/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html).  That way no one makes the mistake of trying to using this with an unsupported generator.

PR #513, opened 2018-07-03
Traverse expressions during bounds declaration checking using a CFG

dtarditi:
To check the validity of bounds declarations, the compiler traverses all expressions in a function and checks validity after each assignment.   The compiler currently uses the abstract syntax tree for the traversal.  This is a problem because we want to use dataflow information during checking.   This changes the compiler to use a control-flow graph (CFG) for the traversal.   The compiler constructs the clang CFG for a function body.  It then traverses the CFG in reverse post order (RPO) and checks each "CFG element" in each  basic block.

I ran into several issues doing this:
- Clang builds a CFG where individual elements of basic blocks may be expressions.   In the case of control flow within an expression, for call expressions, and in a few other cases, clang may make a subexpression of an expression a CFG node element.   We only want to traverse an expression once during bounds declaration checking.      Clang doesn't have back pointers from expressions to their parents, so the compiler can't easily determine when it is looking at a subexpression.  The compiler collects the set of CFG elements that are subexpressions of other CFG expressions and ignores them during the traversal.
- It is difficult to determine when an expression is in a checked scope or not based on the CFG.  The compiler builds the set of all expressions in a function that are in a checked scope and uses that set to answer the question.
- clang is sometimes unable to build a valid CFG.   I believe this will be very rare.   In this case, we fall back to the current AST traversal order.  We won't have dataflow facts available when this happens.   There are two situations where clang isn't able to build a CFG: the AST is incorrect (for example, there was an error building it) or the program uses the Microsoft `__finally` extension:
- I hit an assert in the clang CFG builder for one of the clang test cases (loop-control.c).  The problem is this for loop:
```
  for ( ; ; (void)({ if (first) { first = 0; continue; } 0; })) {}
```
This is a for-loop with  an increment expression that uses a GCC expression statement extension with a continue.   The CFG builder asserts given this loop.  No assert is hit during testing currently because the CFG isn't being built for this test case.  Adding -Wall to the command line for this case  causes a CFG to be built and the debug compiler to assert.

I investigated the bug and the problem is that clang is building an invalid CFG.  This causes the assert to fail.  There's a check immediately after the assert for an invalid CFG.    The assert should be moved after the check.  The CFG builder should be fixed too, but that is outside the scope of this change.
- The order in which expressions is checked changed in the test case inferred-bounds/member-base. There is a for-loop that introduces a cycle in the CFG.  This introduces some possible choices in RPO traversal.   Both body and exit must be traversed before f, but the order in which they must be traversed is not specified.   The traversal examines exit first and then body.   The fix is to put the for loop in a separate function, so the order in which to check bounds expressions is not ambiguous.
```
   |
  \|/
   f <------
   |  \     |
  \|/  _\|  |
exit   body
```
Testing:
- Passed Linux LNT and debug build automation tests.
- Passed Windows x86 automation.
- Passed local Windows x64 build.


lib/Sema/SemaBounds.cpp
```diff
@@ -1379,6 +1384,9 @@ namespace {
     Sema &S;
     bool DumpBounds;
     uint64_t PointerWidth;
+    Stmt *Body;
+    CFG *Cfg;
+
```
sxl463: Why do we need this empty line here? Anything special?

lib/Sema/SemaBounds.cpp
```diff
@@ -2145,11 +2153,143 @@ namespace {
 
 
   public:
-    CheckBoundsDeclarations(Sema &S, BoundsExpr *ReturnBounds) : S(S),
+    CheckBoundsDeclarations(Sema &S, Stmt *Body, CFG *Cfg, BoundsExpr *ReturnBounds) : S(S),
```
sxl463: The initialization S(S) looks a little weird to me.

lib/Sema/SemaBounds.cpp
```diff
@@ -2145,11 +2153,143 @@ namespace {
...
+
+    void IdentifyChecked(Stmt *S, StmtSet &CheckedStmts, bool InCheckedScope) {
+      if (!S)
+        return;
+
+      if (InCheckedScope)
+        if (isa<Expr>(S) || isa<DeclStmt>(S) || isa<ReturnStmt>(S))
+          CheckedStmts.insert(S);
+
+      if (CompoundStmt *CS = dyn_cast<CompoundStmt>(S))
```
sxl463: LLVM recommends adding "const" when using dyc_cast for type testing
if (const type *Val = dyn_cast<T>(oldVal))

lib/Sema/SemaBounds.cpp
```diff
@@ -2145,11 +2153,143 @@ namespace {
...
+      }
+   }
+
+  // Identify CFG elements that are statements that are substatements of other
+  // CFG elements.  (CFG elements are the components of basic blocks).  When a
+  // CFG is constructed, subexpressions of top-level expressions may be placed
+  // in separate CFG elements.  This is done for subexpressions of expressions
+  // with control-flow, for example. When checking bounds declarations, we want
+  // to process a subexpression with its enclosing expression. We want to
+  // ignore CFG elements that are substatements of other CFG elements.
```
sxl463: I suggest putting a simple example here to help users understand, like:
Some complex conditional expressions may bring extra control-flow...
x = (x == 0) ? f1() : f2();

lib/Sema/SemaBounds.cpp
```diff
@@ -2145,11 +2153,143 @@ namespace {
...
+   void FindNestedElements(StmtSet &NestedStmts) {
+      // Create the set of top-level CFG elements.
+      StmtSet TopLevelElems;
+      CFG::iterator Iter = Cfg->begin();
+      CFG::iterator IterEnd = Cfg->end();
+      for ( ; Iter != IterEnd; ++Iter) {
```
sxl463: Why don't we use the more clear C++ 17 iteration here? like:
for (const auto& iter : CFG)
...

lib/Sema/SemaBounds.cpp
```diff
@@ -2145,11 +2153,143 @@ namespace {
...
+     StmtSet CheckedStmts;
+     IdentifyChecked(Body, CheckedStmts, false);
+     PostOrderCFGView POView = PostOrderCFGView(Cfg);
+     PostOrderCFGView::iterator PO_Iter= POView.begin();
+     PostOrderCFGView::iterator PO_End = POView.end();
+     for ( ; PO_Iter != PO_End; ++PO_Iter) {
+       const CFGBlock *Block = *PO_Iter;
+       CFGBlock::const_iterator ElemIter = Block->begin();
+       CFGBlock::const_iterator ElemEnd = Block->end();
+       for (; ElemIter != ElemEnd; ++ElemIter) {
```
sxl463: Same as line 2206, I thinkg using "for (const auto& PO_iter : POView)" can simplify the code

dtarditi:
Good suggestions.  I've added a comment and made the other code improvements you suggested.

PR #515, opened 2018-07-07
Newline each input file in the generated script

AnnaKornfeldSimpson:
Changes the output script produced to give each argument (both `-arguments` and file names) its own line. This will allow diffs and make it much easier to see how many files a particular configuration would send to the convert tool.

If the resulting script is read for purposes other than being a shell script, having the `\` present may require a little extra work, but I think the benefits of having line numbers would be worth the effort. Anyone else have thoughts?

### DIFF
--- a/tools/checked-c-convert/utils/run.py
+++ b/tools/checked-c-convert/utils/run.py
@@ -57,7 +57,7 @@ def runMain(args):
   args.extend(DEFAULT_ARGS)
   args.extend(list(s))
   f = open('convert.sh', 'w')
-  f.write(" \\ \n".join(args))
+  f.write(" \\\n".join(args))
   f.close()
   subprocess.check_call(args)
 
--- a/tools/checked-c-convert/utils/run.py
+++ b/tools/checked-c-convert/utils/run.py
@@ -56,8 +56,8 @@ def runMain(args):
   args.append(prog_name)
   args.extend(DEFAULT_ARGS)
   args.extend(list(s))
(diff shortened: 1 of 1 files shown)


tools/checked-c-convert/utils/run.py
```diff
@@ -56,8 +56,8 @@ def runMain(args):
   args.append(prog_name)
   args.extend(DEFAULT_ARGS)
   args.extend(list(s))
-  f = open('bla', 'w')
-  f.write(" ".join(args))
+  f = open('convert.sh', 'w')
+  f.write(" \\\n".join(args))
```
awruef: What about `os.linesep`?

AnnaKornfeldSimpson: Neat, a new Python thing that I didn't know about! https://docs.python.org/2/library/os.html#os.linesep says not to use it when using this to write to text files, though. So I assume it's only supposed to be for files opened with 'wb' ?

awruef: On further reading, I suspect this is something that I have just been cargo culting without fully understanding how it works. You're right. I was just worried that on Windows, if this ever switched to outputting `.bat` files, the newline representation would have to be changed, but the documentation I read that you linked seems to indicate that is not the case.

AnnaKornfeldSimpson: Yes I thought the same about Windows until I read the docs.

Issue #517, opened 2018-07-10
Need automatic way to change header includes to checked versions


AnnaKornfeldSimpson:
When running checked-c-convert or otherwise doing translation for a number of source files, it would be really helpful if there were an automated tool or script to do this for the small number of *_checked headers in https://github.com/Microsoft/checkedc/tree/f464045e92f67c8ed144de475af6d9df653849a1/include

AnnaKornfeldSimpson:
If the header file associated with that source file is also changed to <filename>.checked.h, replacing that in <filename>.checked.c would be great as well.

dtarditi:
@AnnaKornfeldSimpson added a script to do this, so closing this issue.

PR #529, opened 2018-07-16
CheckedC Header convert script

AnnaKornfeldSimpson:
This utility script takes a file with a list of c filenames, and adjusts their `#include<>` of standard headers to be the `_checked` versions from https://github.com/Microsoft/checkedc/tree/master/include
This *rewrites* the listed c files to use these headers. If the script is working correctly, no compiler behavior should become worse as a result of its use.

Additional Details:
- Right now, this script is designed to take a file formatted in the manner of a call to checked-c-convert, as output as "convert.sh" by the utility in this folder. This means there's some extra characters that need to be stripped off the end of lines.
- Since there are likely to be many c files for the program to iterate through and the c files may be long, there is a preprocessing step to build the "find/replace" function for the header files. This allows looking for all header replacements in a single pass through the file.

Addresses #517


AnnaKornfeldSimpson:
Although this is helpful, it does not entirely do the job because it ignores anything included in other header files. Header files are not listed in the compiler_commands.json so it is harder to collect a relevant list. One option would be to just `find` all the `.h` files in the repo and then use them as an input file. This would require only minor changes to the script to accept `.h` files as well as `.c` files in the main loop. There might be other options as well; any ideas?

AnnaKornfeldSimpson:
Commit above changes the script to accept `.h` files as well as `.c` files since plenty of standard headers are included in `.h` files. Addresses my previous comment.

tools/checked-c-convert/utils/headerConvertScript.py
```diff
@@ -0,0 +1,67 @@
+"""
+This program takes a list of .c files, e.g. in a convert.sh file,
+and converts any included headers for which checkedc has _checked 
+to the checked versions. The list of checked headers is found in
+the llvm/projects/checkedc-wrapper/checkedc/include directory.
+
+NOTE: This does not  any local / project-specific headers
```
dtarditi: Is there a word missing here?

tools/checked-c-convert/utils/headerConvertScript.py
```diff
@@ -0,0 +1,67 @@
...
+Those must still be updated manually.
+"""
+
+import os, re, sys
+import argparse
+
+# This assumes the file is inside the tools/checked-c-convert/utils
+# directory of a tools/checkedc-clang repo, which is inside an llvm repo
+# containing the checkedc-wrapper. Adjust if the headers are elsewhere.
+CHECKEDC_HEADER_DIR=os.path.abspath(
```
dtarditi: I suggest adding an optional argument that sets this and using this value only if nothing is set.

PR #536, opened 2018-07-27
Nightly Release Package Build for Linux

AnnaKornfeldSimpson:
Adds a script that does a nightly release package build for Linux, the same way we do for Windows. Include that script at the bottom of the main Linux build and test script.
This package is configured as Release - without Debug, regardless of how the build that runs the other tests is configured. Thus it does re-build the compiler, but only the Release version.


AnnaKornfeldSimpson:
Fixed the build configuration.

automation/UNIX/build-package.sh
```diff
@@ -0,0 +1,53 @@
...
+set -o pipefail
+set -x
+
+cd ${LLVM_OBJ_DIR}
+
+if [ "${BUILD_PACKAGE}" == "No" ]; then cmdsucceeded; fi
+
+echo "Building installation package for clang"
+
+if [ "${TEST_TARGET_ARCH}" == "X86" ]; then
```
dtarditi: Clang is a cross-compiler by default that can target multiple architectures.  This setting tells clang to only target X86.  I think you want to delete this if statement.

AnnaKornfeldSimpson: I copied that from run-cmake.sh lines 9-13. Does it play a role there or should that be cleared out too?

dtarditi:
I think it is advisable to set a value for BUILD_PACKAGE by default. The build automation invokes build-and-test.sh, which sources config-vars.sh before calling setup-files.sh.  It does that so bash places the variable changes in the environment.

From what I remember, bash doesn't like it if you use undefined variables and fails your script.  So comparing against Yes will still fail of BUILD_PACKAGE is undefined.

automation/UNIX/build-package.sh
dtarditi: I think run-cmake.sh needs to be updated too.

automation/UNIX/run-cmake.sh
```diff
@@ -6,16 +6,12 @@ set -ue
 set -o pipefail
 set -x
 
-if [ "${TEST_TARGET_ARCH}" == "X86" ]; then
-  CMAKE_ADDITIONAL_OPTIONS="-DLLVM_TARGET_ARCH=X86"
+if [[ ("$LNT" != "" || "$BUILD_PACKAGE" == "Yes") && ("$BUILDCONFIGURATION" == "Release" || "$BUILDCONFIGURATION" == "ReleaseWithDebInfo") ]]; then
```
dtarditi: When building a package, you also want to set -DLLVM_INSTALL_TOOLCHAIN_ONLY=ON.  Otherwise the install package will be much larger than necessary.

PR #539, opened 2018-08-03
Fix some bugs related to retention of existing Checked C annotations

awruef:
I'm working along fixing some bugs reported, and along the way I encountered and fixed some errors related to retaining Checked C types in programs input to the re-writer. I've added tests that will help catch those failures in the future. I think that fixing the remaining issues will require a small change to insert casts in a situation where the callers parameter is unconstrained, but the callee's parameter is constrained. That will happen in a separate pull request.

### DIFF
--- a/test/CheckedCRewriter/function_tests.c
+++ b/test/CheckedCRewriter/function_tests.c
@@ -166,7 +166,7 @@ int *ok_mut_clone(int *a, int b) {
   *a = b;
   return a;
 }
-//CHECK: int *ok_mut_clone(int *a, int b) {
+//CHECK: int* ok_mut_clone(int *a, int b) {
 
 ok_mut_t get_mut_2(void) {
   return &ok_mut_clone;
--- a/tools/checked-c-convert/CheckedCConvert.cpp
+++ b/tools/checked-c-convert/CheckedCConvert.cpp
@@ -731,45 +731,44 @@ bool CastPlacementVisitor::VisitFunctionDecl(FunctionDecl *FD) {
   assert(cDefn != nullptr);
 
   if (cDecl->numParams() == cDefn->numParams()) { 
-    bool didAny = false;
+    // Track whether we did any work and need to make a substitution or not.
+    bool didAny = cDecl->numParams() > 0;
     std::string s = "";
     std::vector<std::string> parmStrs;
     // Compare parameters. 
-    if (cDecl->numParams() == cDefn->numParams())
-      for (unsigned i = 0; i < cDecl->numParams(); ++i) {
-        auto Decl = getHighest(cDecl->getParamVar(i), Info);
-        auto Defn = getHighest(cDefn->getParamVar(i), Info);
-        assert(Decl);
-        assert(Defn);
+    for (unsigned i = 0; i < cDecl->numParams(); ++i) {
+      auto Decl = getHighest(cDecl->getParamVar(i), Info);
+      auto Defn = getHighest(cDefn->getParamVar(i), Info);
+      assert(Decl);
+      assert(Defn);
 
-        didAny = true;
-        // If this holds, then we want to insert a bounds safe interface. 
-        bool anyConstrained = Defn->anyChanges(Info.getConstraints().getVariables());
-        if (Defn->isLt(*Decl, Info) && anyConstrained) {
+      // If this holds, then we want to insert a bounds safe interface. 
+      bool anyConstrained = Defn->anyChanges(Info.getConstraints().getVariables());
+      if (Defn->isLt(*Decl, Info) && anyConstrained) {
+        std::string scratch = "";
+        raw_string_ostream declText(scratch);
+        Definition->getParamDecl(i)->print(declText);
+        std::string ctype = Defn->mkString(Info.getConstraints().getVariables(), false);
+        std::string bi = declText.str() + " : itype("+ctype+") ";
+        parmStrs.push_back(bi);
+      } else {
+        // Do what we used to do.
+        if (anyConstrained) { 
+          std::string v = Defn->mkString(Info.getConstraints().getVariables());
+          if (PVConstraint *PVC = dyn_cast<PVConstraint>(Defn)) {
+            if (PVC->getItypePresent()) {
+              v = v + " : " + PVC->getItype();
+            }
+          }
+          parmStrs.push_back(v);
+        } else {
           std::string scratch = "";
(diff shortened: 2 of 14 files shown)


tools/checked-c-convert/ProgramInfo.h
```diff
@@ -130,7 +122,8 @@ class FunctionVariableConstraint;
 class PointerVariableConstraint : public ConstraintVariable {
 public:
 	enum Qualification {
-		ConstQualification
+		ConstQualification,
+    StaticQualification
```
dtarditi: There's a spacing issue here.

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -732,64 +732,84 @@ bool CastPlacementVisitor::VisitFunctionDecl(FunctionDecl *FD) {
...
+                v = v + " : " + PVC->getItype();
+              }
+            }
+            parmStrs.push_back(v);
+          } else {
+            std::string scratch = "";
+            raw_string_ostream declText(scratch);
+            Definition->getParamDecl(i)->print(declText);
+            parmStrs.push_back(declText.str());
+                      }
```
dtarditi: There's a white space issue here.

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -732,64 +732,84 @@ bool CastPlacementVisitor::VisitFunctionDecl(FunctionDecl *FD) {
...
+    std::vector<std::string> parmStrs;
+    // Compare parameters. 
+    if (cDecl->numParams() == cDefn->numParams())
+      for (unsigned i = 0; i < cDecl->numParams(); ++i) {
+        auto Decl = getHighest(cDecl->getParamVar(i), Info);
+        auto Defn = getHighest(cDefn->getParamVar(i), Info);
+        assert(Decl);
+        assert(Defn);
+
+        didAny = true;
```
dtarditi: I don't understand why you moved the assignment to didAny to here.   This seems to always set it.

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -732,64 +732,84 @@ bool CastPlacementVisitor::VisitFunctionDecl(FunctionDecl *FD) {
...
+        if (Defn->isLt(*Decl, Info) && anyConstrained) {
+          std::string scratch = "";
+          raw_string_ostream declText(scratch);
+          Definition->getParamDecl(i)->print(declText);
+          std::string ctype = Defn->mkString(Info.getConstraints().getVariables(), false);
+          std::string bi = declText.str() + " : itype("+ctype+") ";
+          parmStrs.push_back(bi);
+        } else {
+          // Do what we used to do.
+          if (anyConstrained) {
```
dtarditi: This comment should describe the case that this code is handling.

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -732,64 +732,84 @@ bool CastPlacementVisitor::VisitFunctionDecl(FunctionDecl *FD) {
...
+            parmStrs.push_back(declText.str());
+                      }
+        }
+      }
+
+    // Compare returns. 
+    auto Decl = getHighest(cDecl->getReturnVars(), Info);
+    auto Defn = getHighest(cDefn->getReturnVars(), Info);
+
+    // Insert a bounds safe interface at the return address.
```
dtarditi: Instead of saying "at the return address", should the comment say "insert a bounds-safe interface for the return"?

tools/checked-c-convert/CheckedCConvert.cpp
```diff
@@ -732,64 +732,84 @@ bool CastPlacementVisitor::VisitFunctionDecl(FunctionDecl *FD) {
...
+    std::string returnVar = "";
+    std::string endStuff = "";
+    bool anyConstrained = Defn->anyChanges(Info.getConstraints().getVariables());
+    if (Defn->isLt(*Decl, Info) && anyConstrained) {
+      std::string ctype = Defn->mkString(Info.getConstraints().getVariables());
+      returnVar = Defn->getTy();
+      endStuff = " : itype("+ctype+") ";
+      didAny = true;
+    } else {
+      // Do what we used to do at the return address.
```
dtarditi: This comment should describe the case that is being handled.

tools/checked-c-convert/ProgramInfo.h
```diff
@@ -150,14 +143,20 @@ class PointerVariableConstraint : public ConstraintVariable {
   // If for all U in arrSizes, any U -> (a,b) where a = O_SizedArray or 
   // O_UnSizedArray, arrPresent is true.
   bool arrPresent;
+  // Is there an itype associated with this constraint? If there is, how was it
```
dtarditi: There should be comments on the public accessors too.

test/CheckedCRewriter/function_tests.c
```diff
@@ -166,7 +166,7 @@ int *ok_mut_clone(int *a, int b) {
   *a = b;
   return a;
 }
-//CHECK: int *ok_mut_clone(int *a, int b) {
+//CHECK: int* ok_mut_clone(int *a, int b) {
```
dtarditi: I don't understand why the spacing changed here.

tools/checked-c-convert/CheckedCConvert.cpp
awruef: Roughly, there's now only one case where we won't re-write a declaration in this function - which is when there is a parameter mis-match indicating we don't exactly know what to do. I should remove `didAny` and replace it with something more sensible.

tools/checked-c-convert/CheckedCConvert.cpp
awruef: Yeah, probably.

tools/checked-c-convert/ProgramInfo.h
awruef: I agree, I added it.

test/CheckedCRewriter/function_tests.c
awruef: Where we used to leave return values alone, we started changing them sometimes. This sounds unsatisfying as I type it, but it's the world we live in now. I could try to get to the bottom of it now, or later.

awruef:
I addressed the feedback minus the spacing thing.

Issue #205, opened 2017-03-17
Add support for _Return_value


dtarditi:
Checked C allows return bounds expression for functions to use the keyword `_Return_value` to denote the return value of a function.   It can be used in bounds expressions and bounds can be declared for it.

The proposed work items are:
- Extend the IR with a way to represent Return_value and Current_expr_value  The like approach is to create a new kind of expression similar to PositionalParameterExpr, that is used only in bounds expressions.
- Extend the syntax to allow them to be parsed where an identifier can be parsed.  Construct the new IR expression to represent it.
- _Return_value should only be allowed in the context of parsing a return bounds expression.
- Current_expr_value should only be allowed in the context of parsing a bounds expression (excluding a return bounds expression).

Note that I need the IR representation for _Current_expr_value soon (in the next few days), so if anyone picks this up soon, they should coordinate the IR work me.

dtarditi:
I am replacing _Current_expr_value with expression temporaries.   This leads to a cleaner, more efficient semantics.  With _Current_expr_value, we had to adjust uses of it when inferring bounds expressions.  The adjustment had to to offset the change between a subexpression value and the parent expression value.  With expression temporaries, we no longer have to make this adjustment.  An expression temporary is a temporary variable inserted by the compiler. 

The point at which assignment of a value to a temporary variable happens can be controlled precisely  This is unlike C semantics for nested assignments to source-level variables. which informally says that "the assignments have to happen at some point before the evaluation of the expression is complete" (more precisely, happen in an order that respects sequence points).

sulekhark:
This issue got automatically closed by mistake: GitHub closed issue number N in the `checkedc-clang` repository when a 3C PR that has a comment "Fixes issue N" (where N refers to the issue number in CCI's repository), was merged to `master` branch in the `checkedc-clang` repository. Reopening it.

PR #544, opened 2018-08-11
Add support for _Return_value expression.

dtarditi:
The _Return_value expression allows programmers to name the value returned by a function in a return bounds declaration.  This changes extends the clang IR to represent it and _Current_expr_value. It also adds lexing, parsing, and type system support for _Return_value.   This addresses the first part of issue #205.  Support for checking of bounds declarations using _Return_value will be implemented next.

We create a new expression class BoundsValueExpr to represent these expressions.  This class is modelled on CXXThisExpr.   Constructing a_Return_value expression requires knowing the return type of the function that the return bounds is associated with.  We must have the _Return_value expression constructed before we can build the function type, since function types incorporate the return bounds expressions.  Unfortunately, we don't know the function return type until the middle of GetFullTypeForDeclarator in SemaType.cpp.  Because clang intertwines AST building, type checking, and other static semantic checking, we have to "defer parse" return bounds expressions.   "Defer parse" means that the compiler recognize the tokens that make up a return bounds expression, saves them, and parses them later to build the IR for the return bounds expressions.

To allow parsing of the return bounds to happen in the middle of SemaType.cpp, we introduce a callback from semantic checking to the parser.  This lets us preserve the interface between the two.  We have to bring the parameters into scope, since they've already gone out of scope by the time GetFullTypeForDeclarator runs.   We aren't the first people to have to do this.  Other parsing phases such as parsing of attributes have to do this too.

After parsing the return bounds expression, we attach it to the declarator whose type is being constructed.    This lets us get the line number information right for error messages involving redeclarations with conflicting return bounds.  (Long explanation: memoization of function types also memoizes bounds expressions embedded in ty pes.  It does this by ignoring line numbers and abstracting function parameters to parameter locations.   The end result is that line numbers for expressions embedded in types are meaningless.  Clang has a complicated mechanism for tracking source line information for types called TypeSourceInfo, but we haven't implemented support for embedded expressions in types and are not likely to anytime soon).

This change also contains some unrelated clean up:
1. Fix the line ending for the test case for bug526 to be line feeds.
2. Delete identifiers for _Dynamic_bounds_cast and _Assume_bounds_cast.  These are keywords, so we don't need identifiers.

Testing:
-  There will be a separate pull request to the Checked C repo for parsing and typechecking tests for_Return_value.
- Added clang tests for AST dumping and checking of bounds declarations.  The checking of bounds declarations involving _Return_value doesn't work yet, so the tests document this.
- Passes automated Checked C and clang testing with these changes.


lib/Parse/ParseExpr.cpp
```diff
@@ -3577,6 +3580,42 @@ Parser::DeferredParseBoundsExpression(std::unique_ptr<CachedTokens> Toks,
...
+  ParseScope PrototypeScope(TheParser, PrototypeScopeFlag);
+  TheParser->Actions.ActOnSetupParametersAgain(TheParser->Actions.CurScope, Params);
+  bool Err = TheParser->DeferredParseBoundsExpression(std::move(Toks), Result, D);
+  PrototypeScope.Exit();
+  return Err;
+}
+
+ExprResult Parser::ParseReturnValueExpression() {
+  assert(Tok.is(tok::kw__Return_value) &&
+         "Not bounds  value expression");
```
Prabhuk: Double space between bounds and value in the message "Not bounds value expression"

lib/Sema/DeclSpec.cpp
```diff
@@ -215,8 +216,8 @@ DeclaratorChunk DeclaratorChunk::getFunction(bool hasProto,
                                   TrailingReturnType.isInvalid();
   I.Fun.TrailingReturnType      = TrailingReturnType.get();
   I.Fun.ReturnAnnotsColonLoc    = ReturnAnnotsColonLoc.getRawEncoding();
-  I.Fun.ReturnBounds            = ReturnAnnotsExpr.getBoundsExpr();
-  I.Fun.ReturnInteropType       = ReturnAnnotsExpr.getInteropTypeExpr();
+  I.Fun.ReturnBounds            =  ReturnBounds.release();
```
Prabhuk: Remove space before "ReturnBounds.release()"

include/clang/Sema/DeclSpec.h
```diff
@@ -1377,11 +1377,11 @@ struct DeclaratorChunk {
...
 
     /// The annotations for the value returned by the function.  We store them
-    // as individual fields, not BoundsAnnotations, because BoundsAnnotations has
-    // a default constructor.  This is for good reason. For equally good reasons,
-    // DeclaratorChunk doesn't have a default constructor.  This means that using
-    // a default initialized instance of BoundsAnnotations is not allowed here.
-    BoundsExpr *ReturnBounds;
+    // as individual fields because the return bounds are deferred-parsed.
+    // Note: ReturnBounds is actually a unique_ptr.  However unique_ptr requires
+    // a constructor and this struct can't have one, so so we cast it to a
```
Prabhuk: Double space before "However" and duplicate "so".

lib/Parse/ParseDecl.cpp
```diff
@@ -1959,29 +1959,9 @@ Parser::DeclGroupPtrTy Parser::ParseDeclGroup(ParsingDeclSpec &DS,
...
-                        SkipUntilFlags::StopBeforeMatch);
-            }
-          }
-        }
-
-        // Case 2: the return bounds expression is misplaced for a complex
-        // function declarator. Diagnosis this, suggest a fix, and bail out.
+        // It is easy to misplace it.  Handle the case where the return bounds
+        // expression is misplaced for a complex function declarator.
+        // Diagnosis this, suggest a fix, and bail out.
```
Prabhuk: Diagnosis -> Diagnose

PR #546, opened 2018-08-11
 Add support for polymorphic bounds safe interface functions

Prabhuk:
Added new keyword _Itype_for_any which is a function specifier.

Pending:
1. Parsing logic of _For_any and _Itype_for_any seems to fail in linux
2. Spec needs to be updated

### DIFF
--- a/include/clang/AST/ASTContext.h
+++ b/include/clang/AST/ASTContext.h
@@ -1198,10 +1198,12 @@ public:
   /// pointer to blocks.
   QualType getBlockDescriptorExtendedType() const;
 
-  /// Returns a type variable type based on the depth of the "for any" scope 
-  /// where type variable is declared, and the position of the type variable in
-  /// _For_any qualifier.
-  QualType getTypeVariableType(unsigned int depth, unsigned int position) const;
+  /// Returns a type variable type based on the depth of "for any" or
+  /// "itype for any" scope where type variable is declared, and the
+  /// position of the type variable in _For_any or _Itype_for_any
+  ///.qualifier respectively
+  QualType getTypeVariableType(unsigned int depth, unsigned int position,
+                               bool isBoundsInterfaceType) const;
 
   void setcudaConfigureCallDecl(FunctionDecl *FD) {
     cudaConfigureCallDecl = FD;
@@ -2486,12 +2488,16 @@ public:
   /// at least as much checking as the other type T2.  Return true if it does
   /// or false if it does not or the types differ in some other way than
   /// checkedness.
+  /// Note:: Checked C - In bounds safe interface scopes, this function
+  /// returns true if T2 is a TypeVariableType and T1 is a pointer to void type
   bool isAtLeastAsCheckedAs(QualType T1, QualType T2) const;
 
   /// \brief Determine whether a pointer, array, or function type T1
   /// is the same as the other pointer, array, or function type T2 if
   /// checkedness is ignored.  Return true if does or false if the types
   /// differ in some other way than checkedness.
+  /// Note:: Checked C - In bounds safe interface scopes, this function
+  /// returns true if T2 is a TypeVariableType and T1 is a pointer to void type
   bool isEqualIgnoringChecked(QualType T1, QualType T2) const;
 
   /// \brief Return true if this type is a checked type that is not
--- a/include/clang/AST/ASTContext.h
+++ b/include/clang/AST/ASTContext.h
@@ -2489,7 +2489,7 @@ public:
   /// or false if it does not or the types differ in some other way than
   /// checkedness.
   /// Note:: In bounds safe interface scopes, this function
-  /// returns true if T2 is a TypeVariableType and T1 is a pointer to void type
+  /// returns true if T1 is a TypeVariableType and T2 is a pointer to void type
   bool isAtLeastAsCheckedAs(QualType T1, QualType T2) const;
 
   /// \brief Determine whether a pointer, array, or function type T1
@@ -2497,7 +2497,7 @@ public:
   /// checkedness is ignored.  Return true if does or false if the types
   /// differ in some other way than checkedness.
   /// Note:: In bounds safe interface scopes, this function
-  /// returns true if T2 is a TypeVariableType and T1 is a pointer to void type
+  /// returns true if T1 is a TypeVariableType and T2 is a pointer to void type
   bool isEqualIgnoringChecked(QualType T1, QualType T2) const;
 
   /// \brief Return true if this type is a checked type that is not
--- a/include/clang/AST/ASTContext.h
+++ b/include/clang/AST/ASTContext.h
@@ -2488,7 +2488,7 @@ public:
   /// at least as much checking as the other type T2.  Return true if it does
   /// or false if it does not or the types differ in some other way than
   /// checkedness.
-  /// Note:: Checked C - In bounds safe interface scopes, this function
+  /// Note:: In bounds safe interface scopes, this function
   /// returns true if T2 is a TypeVariableType and T1 is a pointer to void type
   bool isAtLeastAsCheckedAs(QualType T1, QualType T2) const;
 
@@ -2496,7 +2496,7 @@ public:
   /// is the same as the other pointer, array, or function type T2 if
   /// checkedness is ignored.  Return true if does or false if the types
   /// differ in some other way than checkedness.
-  /// Note:: Checked C - In bounds safe interface scopes, this function
+  /// Note:: In bounds safe interface scopes, this function
   /// returns true if T2 is a TypeVariableType and T1 is a pointer to void type
   bool isEqualIgnoringChecked(QualType T1, QualType T2) const;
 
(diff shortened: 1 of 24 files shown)


Prabhuk:
This PR is ready for review.

include/clang/AST/ASTContext.h
```diff
@@ -2486,12 +2488,16 @@ class ASTContext : public RefCountedBase<ASTContext> {
...
   /// checkedness.
+  /// Note:: Checked C - In bounds safe interface scopes, this function
+  /// returns true if T2 is a TypeVariableType and T1 is a pointer to void type
   bool isAtLeastAsCheckedAs(QualType T1, QualType T2) const;
 
   /// \brief Determine whether a pointer, array, or function type T1
   /// is the same as the other pointer, array, or function type T2 if
   /// checkedness is ignored.  Return true if does or false if the types
   /// differ in some other way than checkedness.
+  /// Note:: Checked C - In bounds safe interface scopes, this function
```
dtarditi: You don't need to prefix this with "Checked C", as this function is specific to the Checked C implementation.

include/clang/AST/ASTContext.h
```diff
@@ -2486,12 +2488,16 @@ class ASTContext : public RefCountedBase<ASTContext> {
   /// at least as much checking as the other type T2.  Return true if it does
   /// or false if it does not or the types differ in some other way than
   /// checkedness.
+  /// Note:: Checked C - In bounds safe interface scopes, this function
```
dtarditi: You don't need to prefix this with "Checked C", as this function is specific to the Checked C implementation.

lib/AST/ASTContext.cpp
```diff
@@ -8146,17 +8147,26 @@ QualType ASTContext::mergeFunctionTypes(QualType lhs, QualType rhs,
...
   if (lproto && rproto) { // two C99 style function prototypes
     assert(!lproto->hasExceptionSpec() && !rproto->hasExceptionSpec() &&
            "C++ shouldn't be here");
     // Compatible functions must have the same number of parameters
     if (lproto->getNumParams() != rproto->getNumParams())
       return QualType();
 
-    // Compatible functions must have the same number of type variables.
-    if (lproto->getNumTypeVars() != rproto->getNumTypeVars())
-      return QualType();
```
dtarditi: Place put a space after the 'if' keyword.

lib/AST/ASTContext.cpp
```diff
@@ -8194,6 +8204,12 @@ QualType ASTContext::mergeFunctionTypes(QualType lhs, QualType rhs,
         } else if (!rReturnAnnots.IsEmpty() && lReturnAnnots.IsEmpty()) {
           ReturnAnnots = rReturnAnnots;
           allLTypes = false;
+        } else if (lproto->isItypeGenericFunction() && !rproto->isItypeGenericFunction()) {
```
dtarditi: I think you need to check that rproto is also not a generic function.  Generic function types should not be compatible with itype generic function type.  We will allow itype generic function types to be converted to generic types at assignments, call sites, and returns. We will not allow the reverse.

I think you should leave the code "as is" and open an issue to handle this after this PR is complete.    
This is already a large pull request.    You'll need to write some tests and alter some of the assignment compatibility code.  It's a special case that can wait.

lib/Parse/ParseDecl.cpp
```diff
@@ -7204,6 +7220,91 @@ void Parser::ParseCheckedPointerSpecifiers(DeclSpec &DS) {
         Diag(StartLoc, DiagID) << PrevSpec;
 }
 
+void Parser::ParseItypeforanySpecifier(DeclSpec &DS) {
```
dtarditi: I think you should common this code with parsing for regular forall functions.   You can use formatting specifiers in clang error messages to customize parts of error message for specific cases.

lib/Parse/ParseExpr.cpp
```diff
@@ -1432,9 +1432,10 @@ ExprResult Parser::ParseCastExpression(bool isUnaryExpression,
   // are compiling for OpenCL, we need to return an error as this implies
   // that the address of the function is being taken, which is illegal in CL.
 
-  // Check to see if resExpr is a generic function call. In this case, we must
-  // parse a list of type names that follows generic function decl identifier.
-  if (ParseGenericFunctionExpression(Res)) return ExprError();
+  // Check to see if resExpr is a itype genric or generic function call.
```
dtarditi: There's a typo: genric -> generic.

lib/Parse/Parser.cpp
```diff
@@ -1204,7 +1204,8 @@ Decl *Parser::ParseFunctionDefinition(ParsingDeclarator &D,
   // Tell the actions module that we have entered a function definition with the
   // specified Declarator for the function.
   Sema::SkipBodyInfo SkipBody;
-  Scope* funcDeclScope = D.getDeclSpec().isForanySpecified() ? getCurScope()->getParent() : getCurScope();
+  Scope* funcDeclScope = (D.getDeclSpec().isForanySpecified() || D.getDeclSpec().isItypeforanySpecified())?
```
dtarditi: Need a space before the '?'

lib/Sema/DeclSpec.cpp
```diff
@@ -975,7 +975,7 @@ bool DeclSpec::setFunctionSpecUnchecked(SourceLocation Loc,
 bool DeclSpec::setFunctionSpecForany(SourceLocation Loc,
                                         const char *&PrevSpec,
                                         unsigned &DiagID) {
-  if (FS_forany_specified) {
+  if (FS_itypeforany_specified || FS_forany_specified) {
```
dtarditi: Is the case for itypeforany_specified a conflicting declspec, not a duplicate declspec?

lib/Sema/DeclSpec.cpp
```diff
@@ -985,6 +985,19 @@ bool DeclSpec::setFunctionSpecForany(SourceLocation Loc,
   return false;
 }
 
+bool DeclSpec::setFunctionSpecItypeforany(SourceLocation Loc,
+                                              const char *&PrevSpec,
+                                              unsigned &DiagID) {
+  if (FS_itypeforany_specified || FS_forany_specified) {
```
dtarditi: Same here, in terms of this being a conflict, not a duplicate.

lib/Sema/SemaDecl.cpp
```diff
@@ -8729,6 +8737,17 @@ Sema::ActOnFunctionDeclarator(Scope *S, Declarator &D, DeclContext *DC,
     }
   }
 
+  if (D.getDeclSpec().isItypeforanySpecified()) {
```
dtarditi: Can you common this code with the existing case with IsForanySpecified?

lib/Serialization/ASTWriterStmt.cpp
```diff
@@ -402,9 +402,22 @@ void ASTStmtWriter::VisitDeclRefExpr(DeclRefExpr *E) {
   Record.push_back(E->hasTemplateKWAndArgsInfo());
   Record.push_back(E->hadMultipleCandidates());
   Record.push_back(E->refersToEnclosingVariableOrCapture());
-  bool isGenericFunction = E->GetTypeArgumentInfo() != nullptr;
+  bool isGenericFunction = false;
+  if(E->getDecl() && isa<FunctionDecl>(E->getDecl()))
```
dtarditi: I think the brace should be on the prior line.

lib/AST/ASTContext.cpp
```diff
@@ -8923,6 +8951,18 @@ bool ASTContext::isEqualIgnoringChecked(QualType T1, QualType T2) const {
     return true;
 
   Type::TypeClass T1TypeClass = T1->getTypeClass();
+  if(T1TypeClass == Type::Pointer)
+  {
```
dtarditi: I think the brace should be on the prior line.

include/clang/Basic/DiagnosticParseKinds.td
```diff
@@ -1204,12 +1204,27 @@ def err_forany_type_variable_identifier_expected : Error<
 def err_forany_unexpected_token : Error<
   "unexpected token in _For_any specifier">;
 
+def err_itypeforany_comma_or_parenthesis_expected : Error<
```
dtarditi: Clang development guidelines favor sharing diagnostic messages when possible, to minimize the number of different messages to manage.   Diagnostic message can contain formatting specifiers that allow the message to be customized for a specific circumstance.  We should try to follow those guidelines.

Prabhuk:
The PR comments are addressed. 
Follow-up tasks: 
1. Create a new issue for itype_for_any and for_any function type compatibility.
2. Change the signature of malloc's signature in stdlib_checked header file and fix the uses in checked scopes in tests.

include/clang/AST/ASTContext.h
```diff
@@ -2486,12 +2488,16 @@ class ASTContext : public RefCountedBase<ASTContext> {
   /// at least as much checking as the other type T2.  Return true if it does
   /// or false if it does not or the types differ in some other way than
   /// checkedness.
+  /// Note:: In bounds safe interface scopes, this function
```
dtarditi: T1 and T2 in this comment should be reversed.

Issue #553, opened 2018-08-28
Function pointers to generic functions with bounds-safe interfaces


AnnaKornfeldSimpson:
Unexpected complexity for generic functions! Some people want to define their own memory management functions. To do this, they'll be defining a function pointer to a malloc-like function. For example:
```
typedef void * my_malloc_fun(size_t s) : byte_count(s) itype(_Array_ptr<void>);
typedef _Ptr<my_malloc_fun> Use_This_Malloc_Fun;
``` 
in the header and then
`static Use_This_Malloc_Fun x_malloc = &malloc;`
in the code file.  Produces an error about how the `&malloc` needs type arguments, which doesn't make a lot of sense since it isn't being called here.
The most natural place to mark the generic-ness is at the lowest level:
`typedef _Itype_for_any(T) void * my_malloc_fun(size_t s) : byte_count(s) itype(_Array_ptr<T>);`
and then call `x_malloc<concrete_type>` as one would with malloc.
However, this does not address the complaint about &malloc needing type arguments.

dtarditi:
This is fixed.   We still need to handle taking the address of a generic function applied to type variables.  That is tracked by issue #559.

PR #558, opened 2018-09-05
Parsing function call expressions; confusion between function call and address of expressions addressed

Prabhuk:
While parsing, "address of" operation on generic functions were treated as callsites which led to the parser expecting type arguments. 
This is resolved by making sure that the generic functions are parsed for type arguments only when "address of" operator is not part of the expression that is parsed.

Related to #553

### DIFF
--- a/lib/Parse/ParseExpr.cpp
+++ b/lib/Parse/ParseExpr.cpp
@@ -1431,15 +1431,16 @@ ExprResult Parser::ParseCastExpression(bool isUnaryExpression,
     return ExprError();
   }
 
+  // Check to see if Res is an itype generic or generic function call.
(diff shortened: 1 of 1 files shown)


Prabhuk:
This PR is ready for review.

dtarditi:
This looks good, but I think there is more work to do.   We also need to allow taking the address of a generic function applied to type variables.   That will ben non-trivial, so it should be implemented separately.  The updated specification has some requirements on `&g<int>` with respect to calling conventionsl.    Could you file a work item for the additional work?

We should allow taking the address of a generic function and a generic function applied to type variables:
```
_Forall(T) g(T x);
&g  (produces a value with a generic function type)
&g<int> (produces a value specialized to a generic function type.
```

PR #564, opened 2018-09-14
Notes on building Linux Kernel with Checked C compiler

Prabhuk:
Notes and references for building the kernel.


dtarditi:
Thanks for writing this up!  I think this is a better fit for the wiki.  I've created a page here:   https://github.com/Microsoft/checkedc-clang/wiki/Building-the-Linux-kernel

Issue #596, opened 2018-12-13
Improve equivalence checking of pointer arithmetic expressions


dtarditi:
The Checked C specification calls for expanding pointer arithmetic to bytewise arithmetic before checking equivalence of expressions.   We're not doing that, which is causing various pointer expressions to not be considered equivalent when they are.

For example, given an expression `e1` that is a pointer to T, `(Array_ptr<char>) e1 + (sizeof(T) * i)` is equivalent to `e + i`.   

This kind of arithmetic arises when a programmer declared bounds for a variable using `count`, but a programmer does explicit size calculation as part of allocation.  For example:
```_Array_ptr<int> x : count(i) = malloc(sizeof(int) * i)
```

dtarditi:
With the changes to address work item #594, the compiler will emit additional warning messages when compiling the PtrDist and Olden benchmarks that have been converted to Checked C.  The additional messages are due to this issue when malloc or calloc are used.  

Here are the benchmarks that had additional warnings:

- Olden
  - EM3D: make_graph.c (use of calloc)
  - MST: hash.c, makegraphic. (use of calloc, also requires bundled blocks).

- PtrDist
  - Anagram: anagram.c (use of calloc)
  - YACR2
     -  channel.c: use of malloc
      - hcg.c: use of malloc
      - maze.c: use of free.   Note: there are unrelated warningss due to lack of relational arithmetic. Need to check for possible off by 1 issue
      -  vcg.c: use of malloc and free

mgrang:
@dtarditi Can we close this bug now that this has been implemented by @ppashakhanloo ?

PR #597, opened 2018-12-14
Bounds for call expressions where the bounds depend on call results.

dtarditi:
This change implements inference of bounds for call expressions for cases where the bounds depend on the result of the call expressions.   This changes adds inference for `count` and `byte_count` return bounds, including cases where a bounds-safe interface adds these as return bounds.  An additional case that is not yet implemented is when the return bounds uses the `_Return_value` expression.

We didn't implement the inference before because we needed a way to represent the result of evaluating an expression.  In PR #561, we extended the clang IR with temporary variables.    A temporary variable can be bound to the result of an expression and then used later within the enclosing top-level expression.   This change documents the temporary variable extension to the IR made in PR #561.

This change adds temporary variables for call expressions where the bounds depends on the results of the call..   It moves inference for bounds for call expressions to a separate function that also takes the temporary variable (if any) that will hold the result of the call.   It fills in cases for count and byte_count.   It extends the lowering from clang IR to LLVM IR to handle the case where a scalar expression result is bound to a temporary variable.

Before this change, we were returning Bounds(any) for call expressions where we could not infer bounds.  This disabled bounds checking for the result of a call (for example, `f(len)[]5]` would not have bounds checking. It also allowed static checking of bounds declarations where the calls was the right-hand side of assignment or initializer to pass.

This change adds equality facts for initializers (T lhs = rhs) and assignment (lhs = rhs) where the right-hand side binds a temporary variable v.  We add the fact that lhs == v. 

With this change, the compiler now issues additional warnings about being unable to prove the validity of bounds.  This happens for some benchmarks in PtrDst and Olden.  The compiler doesn't understand when some pointer arithmetic expressions are equivalent.   I open issue #596 to track this.

I updated the Checked C clang rewriter to handle additional places where temporary variables can appear.

Testing:
- Update existing AST tests for bounds inference for calls to have the inferred bounds.
- Add new tests of checking bounds declarations to the Checked C clang repo.
- Update Checked C repo tests for cases where new warning messages appear.  Tagged them with GitHub issues.
- Passing existing Checked C, Checked C clang, and LLVM test suite suites.


docs/checkedc/Implementation-Notes.md
```diff
@@ -136,6 +136,117 @@ The implication is that most dataflow analyses of variables
...
+The temporary variable is only guaranteed to have a valid
+value after the binding expression has been evaluated.  The
+binding expression is subject to the same evaluation rules
+as other C expressions.   
+
+- The `BoundsValueExpr` expression reads the value of a temporary variable.
+Its constructor takes a `CHKCBindTemporaryExpr` as an
+argument.  We do not create a variable name for the
+temporary expression.  The name used in diagnostic messages
+is `value of e`, where `e` is the expression whose result is being computed
```
reubeno: Nice!  Thanks for thinking about the readability of these error messages.

docs/checkedc/Implementation-Notes.md
```diff
@@ -136,6 +136,117 @@ The implication is that most dataflow analyses of variables
...
+and bound to the temporary.  The `CHKBBindTemporaryExpr`
+is not a subexpression of the `BoundsValueExpr`.  Insstead, it
```
reubeno: Super-minor nit: typo in 'instead'.

reubeno:
```suggestion
 is not a subexpression of the `BoundsValueExpr`.  Instead, it
```

awruef:
The changes to the re-writer look fine, too.

include/clang/Sema/Sema.h
```diff
@@ -4894,6 +4894,14 @@ class Sema {
   /// address of members used in bounds.
   void CheckAddressTakenMembers(UnaryOperator *AddrOf);
 
+  /// \brief Check whether E contains a return value expression.
+  bool ContainsReturnValueExpr(Expr *E);
+
+  /// \brief Wrap a call expression in a Checked C temporay binding
+  /// expression, if a temporary is needed to describe the bounds
+  /// of the result of the call expression.
+  ExprResult CreateTemporaryForCall(ExprResult R);
```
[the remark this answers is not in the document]
dtarditi: That's a good suggestion to clarity.  I'll rename the function.

lib/Sema/CheckedCAlias.cpp
```diff
@@ -633,3 +633,35 @@ void Sema::ComputeBoundsDependencies(ModifiedBoundsDependencies &Tracker,
...
+    CheckForReturnValue(Sema &SemaRef) :
+      SemaRef(SemaRef), FoundReturnValue(false) {}
+
+    bool ContainsReturnValueExpr() {
+      return FoundReturnValue;
+    }
+
+    bool VisitBoundsValueExpr(BoundsValueExpr *BVE) {
+      if (BVE->getKind() == BoundsValueExpr::Kind::Return)
+        FoundReturnValue = true;
```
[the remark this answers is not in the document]
dtarditi: Yes, you can have a Visit* method return false to short circuit of the AST walk.  I'll update the code.

dtarditi:
@reubeno, thanks for the code review!   I've addressed all your points.

PR #598, opened 2019-01-04
Implementation of Opaque Types (Work In Progress)

Prabhuk:
Opaque types are one of the three extensions proposed by Checked C to offer type safety.
One of the primary use case of this extension would be to replace void pointers. 
Please refer to [spec](https://github.com/Microsoft/checkedc/blob/master/spec/bounds_safety/void-ptr-replacements.tex) for more details.

Note: This work is still under development.


mshockwave:
Hi,
One of the unresolved issues is the lack of `LinkageInfo` for our new `TypeOpaqueType`. The linkage and visibility of a Type is computed by `LinkageComputer::computeTypeLinkageInfo`(in file lib/AST/Type.cpp), currently there is no associated switch-case for `Type::TypeOpaque`.

Originally I plan to use the same `LinkageInfo` as `Type::Typedef` , but it turned out that this method is only for canonical types and `Type::Typedef` is not one of them. I propose to use 
```
getDeclLinkageAndVisibility(cast<TypeOpaqueType>(T)->getDecl())
```
Since the associated Decl(i.e. `TypeOpaqueDecl`) would be a more reliable source to determinate linkage (and maybe visibility)

P.S. Although it hasn't break anything for now, but some related compilation warnings keep popping out and I think we need face linkage and visibility sooner or later.
P.S.S. For those who don't know, I'm Min-Yih Hsu.

dtarditi:
This PR is still, so I'm closing it.

PR #610, opened 2019-06-11
Add support for ARM32 LNT testing

mgrang:


mgrang:
The test scimark2 seems to run for a long time thereby being flagged as a FAIL.
``
Results
----------------
FAIL : 1
PASS : 1539
Results available at: db_None/v4/nts/1
++ grep FAIL /usr/local/magrang/master/build/LNT-Results-Release-Linux/arm32/result.log
FAIL: MultiSource/Benchmarks/SciMark2-C/scimark2.execution_time (309 of 1540)
FAIL : 1
++ echo 'LNT testing failed.'
LNT testing failed.
++ exit 1
``

mgrang:
Note: TEST_TARGET_ARCH needs to be a space delimited string of all the supported test targets.
For example, to enable X86_64 and ARM targets, set:
TEST_TARGET_ARCH="X86_64 ARM"

dtarditi:
In your original PR, you deleted the TEST-TARGET-ARCH from the LLVM_OBJ_DIR.  I asked you to not make that change.   Given that TEST-TARGET-ARCH can now be a semicolon separated list of clang targets, I think you need to make that change.  Otherwise various other things are going to fail in automation.  This change makes sense anyway because clang is a multi-targeting compiler by default.  When I originally designed the scripts, I hadn't thought about the case where we might emulate a target.  That allows us to test multiple architectures on one host architecture.

There are comments describing the environment variables in automation\Windows\build-and-test.bat.  Could you update the comment about TEST_TARGET_ARCH to indicate that on Linux, it can be a semi-colon separated list of clang-targets

We should eventually rename it to TEST_TARGET_ARCHS.  However that entails changes to our AzureDevOps configurations, so I prefer to not rename it for now.

Could you add a comment to the UNIX/build-and-test.sh to note that the environment variables are described in the Windows/build-and-test.bat?

I think the script needs to do a better job of testing that TEST_TARGET_ARCH contains valid values.  With automation, it is better to test for invalid configurations up front than to let the job run, build the compiler, and then fail tests.

mgrang:
This PR has been replaced by #616. So deleting this one.

PR #611, opened 2019-06-11
Support for custom includes and defines

Machiry:
The current conversion technique ignores the compiler flags, specifically `include paths` and `preprocessor variables`, these are important for clang to compile the corresponding source file.

This patch tries to add support so that compiler flags will be included while invoking the CheckedC-convert tool. Furthermore, this also adds support to handle custom working directory.

### DIFF
--- a/tools/checked-c-convert/utils/convert-commands.py
+++ b/tools/checked-c-convert/utils/convert-commands.py
@@ -11,10 +11,13 @@ It contains some work-arounds for cmake+nmake generated compile_commands.json
 files, where the files are malformed. 
 """
 SLASH = os.sep
+# to separate multiple commands in a line
+CMD_SEP = " &"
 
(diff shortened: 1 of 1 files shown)


tools/checked-c-convert/utils/convert-commands.py
```diff
@@ -56,21 +70,40 @@ def runMain(args):
...
-  args.extend(list(s))
   f = open('convert.sh', 'w')
-  f.write(" \\\n".join(args))
+  for compiler_args, target_directory, src_file in s:
+    args = []
+    # get the command to change the working directory
+    change_dir_cmd = ""
+    if len(target_directory) > 0:
+      if os.name == "nt":
+        change_dir_cmd = "dir " + target_directory + " &"
```
dtarditi: I think you want to use cd for Windows also, not dir.

Issue #586, opened 2018-11-07
Disallow ptr<void>.


dtarditi:
I think we should disallow `ptr<void>`.  The rationale is that a value v of type ptr<T> has bounds given by `bounds((array_ptr<T>) v, (array_ptr<T>) v + 1)`.  Pointer arithmetic on array_ptr<T> does not make sense because `void` is an incomplete type.

In our current implementation, we're assuming that `ptr<void>` is actually a pointer to a single byte.  We allow bounds-safe interfaces with _Ptr<void> as the interface type to be declared.   This seems like an area where programmer could get into trouble because they don't understand that _Ptr<void> is only guaranteed to point to a single byte.

PR #612, opened 2019-06-12
Fix to correct handling of void pointers in itype

Machiry:
## Problem
For `itype`, we expect the tool to use CheckedC types instead of regular `C` types. However, in the case of  `void*`, the tool emits `itype(void*)` instead of `itype(_Ptr<void>)`.

This is because, the tool considers all `void*` as wild as those pointers can be used to hold __any__ pointer types. Consequently, all `void*` types are emitted as regular `C` types. However, in the case of `itype`, we should use `itype(_Ptr<void>)`.

## Fix
The fix tries to specialize the `mkString` function, that emits the type string, to handle the cases when the function is called to be used within `itype`

### DIFF
--- a/tools/checked-c-convert/CheckedCConvert.cpp
+++ b/tools/checked-c-convert/CheckedCConvert.cpp
@@ -748,7 +748,7 @@ bool CastPlacementVisitor::VisitFunctionDecl(FunctionDecl *FD) {
         std::string scratch = "";
         raw_string_ostream declText(scratch);
         Definition->getParamDecl(i)->print(declText);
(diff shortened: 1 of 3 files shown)


dtarditi:
We have a separate work item to disallow `ptr<void>` (issue #586).      The problem is that `_Ptr<void>` and `itype(_Ptr<void>)` are almost always incorrect when used in practice.  We need bounds for  `void` pointers so that we can understand the required extent of the object being pointed to.

For the conversion tool, I believe the right solution for now is to just not  emit `itype(_Ptr<void>)`.

dtarditi:
This change is now stale, given the extensive changes in the conversion tool.

Issue #391, opened 2017-10-05
Disallow casts to null-terminated pointers in checked scopes.


dtarditi:
We should not allow casts from unchecked, ptr, or array_ptr values to null-terminated in checked scopes.  Those pointers might not pointed to a null-terminated array, or the null terminator may be overwritten.

dtarditi:
I believe we are already disallow implicit casts.  This work item is to disallow explicit casts.

mgrang:
This was fixed in https://github.com/microsoft/checkedc-clang/pull/626 and tests were added in https://github.com/microsoft/checkedc/pull/367.

PR #626, opened 2019-07-09
Disallow explicit cast to nt_array_ptr in checked scopes (#391)

jzhou76:
Disallow cast from other checked pointer types to nt_array_ptr in
checked scopes because the source pointer might not point to a
NULL_terminated array. Casting from an unchecked pointer to a
nt_array_ptr pointer should also be prohibited; this has already been
handled as no unchecked pointers are allowed in checked scopes.

Also added a new error message in
clang/include/clang/Basic/DiagnosticSemaKinds.td for casting to
nt_array_ptr in checked scopes.

The test file tests/typechecking/checked_scope_basic.c was updated
with a new function test_cast_to_nt_array_ptr to test
casting to nt_array_ptr.

The change passed the new test code and the regression tests for
checkedc and clang.

### DIFF
--- a/lib/Sema/SemaCast.cpp
+++ b/lib/Sema/SemaCast.cpp
@@ -2703,17 +2703,14 @@ void CastOperation::CheckCStyleCast(bool IsCheckedScope) {
 
     // Disallow cast from other Checked Pointer types to nt_arary_ptr because 
     // the SrcType might not point to a NULL-terminated array.
-    if (DestType->isPointerType()) {
-      if (cast<PointerType>(DestType)->getKind() == CheckedPointerKind::NtArray) {
-        if (SrcType->isPointerType() &&
-            cast<PointerType>(SrcType)->getKind() != CheckedPointerKind::NtArray) {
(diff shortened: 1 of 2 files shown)


lib/Sema/SemaCast.cpp
```diff
@@ -2700,6 +2700,22 @@ void CastOperation::CheckCStyleCast(bool IsCheckedScope) {
       SrcExpr = ExprError();
       return;
     }
+
+    // Disallow cast from other Checked Pointer types to nt_arary_ptr because 
+    // the SrcType might not point to a NULL-terminated array.
+    if (DestType->isPointerType()) {
+      if (cast<PointerType>(DestType)->getKind() == CheckedPointerKind::NtArray) {
```
dtarditi: Could you use the function `isCheckedPointerNtArrayType` for testing whether a type is or is not an nt_array_ptr type?

jzhou76: I have updated the code as suggested.

PR #628, opened 2019-07-12
Added the _MMSafe_ptr pointer in the Clang AST.

jzhou76:
Added a new type of checked pointer  _MMSafe_ptr (also used as
mmsafe_ptr for brevity). This new type of pointer can only point
to a struct object, and all the restrictions imposed on the _Ptr
type of pointer apply to it (these are yet to be implemented). 
It is meant to provide safe memory management: dereferencing 
pointers of this kind is guaranteed to be free from Use-after-Free errors.

- Added this new pointer type in the CheckedPointerKind enum class in
  include/clang/AST/Type.h.
- Added a isCheckedPointerMMSafeType() function to
  include/clang/AST/Type.h.
- Added this new pointer type in declaration and type specifier related
  source files including include/clang/Basic/Specifiers.h,
  include/clang/DeclSpec.h, lib/Parse/ParseDecl.cpp,
  lib/Sema/DeclSpec.cpp, lib/Sema/SemaType.cpp (this file als contains
  the code that builds _MMSafe_ptr as a PointerType.).
- Added _MMSafe_ptr as a keyword in include/clang/Basic/TokenKinds.def.
- Added check on the type of the pointee of a _MMSafe_ptr pointer; 
  only struct types are allowed to be pointed by this new type of pointer.

Also wrote tests to test parsing this new type of pointer in
tests/typechecking/pointer_types.h in the checkedc repository.
The current changes of this repo and the checkedc repo passed all
the checkedc and clang tests.


jzhou76:
This PR is not ready to be merged. We have decided to create _MMSafe_ptr as a **real** new type of pointer which is fundamentally different from the other three new types of pointers ptr, array_ptr, and nt_array_ptr. _MMSafe_ptr has an extra field of ID besides a raw pointer, while all the other three new pointers are essentially a raw pointer plus all necessary checks. Some changes in this PR, such as adding the _MMSafe_ptr as a keyword, are correct and will be kept; but the change in lib/Sema/SemaType.cpp (around line 1560) that added _MMSafe_ptr to be a Pointer type is incorrect because later function `BuildPointerType()` would build _MMSafe_ptr as a normal `Pointer` type in Clang.

lib/Sema/SemaType.cpp
```diff
@@ -2014,6 +2014,13 @@ QualType Sema::BuildPointerType(QualType T, CheckedPointerKind kind,
     return QualType();
   }
 
+  // In Checked C, _MMSafe_ptr is only allowed to point to struct types.
+  if (kind == CheckedPointerKind::MMSafe&& !T->isStructureType()) {
```
jzhou76: This is a bug here. Currently this code would fail to compile when we have a `_MMSafe_ptr<T> p` in a generic function.
The compiler complains that 'T' is not a struct type. I'll fix this soon.

mgrang: Space needed between CheckedPointerKind::MMSafe&&

jzhou76: Thanks. I noticed that right after the commit. I'll fix it in next commit.

jzhou76:
@dtarditi this PR is ready for reviewing.

PR #636, opened 2019-07-19
Add support for generic structs

abeln:
This PR adds generic structs to Checked C.

Generic structs can be defined and then used like so
```C
struct List _For_any(T) {
  T *head;
  struct List<T> *next;
};

struct List<int> *lst;
int *h = lst->head;
lst = lst->next;
```

Here's a summary of what's in these changes:

1. Changed the parser so we can handle both type parameters and type arguments (type application).
2. Declarations and definitions of generic structs are represented as `RecordDecls` that store a list of type parameters.
3. Type applications (uses of generic structs) are also represented with `RecordDecls` containing the "base" `RecordDecl` as well as a list of type arguments (e.g. `List<int>` points to both `List` and `[int]`).
4. We eagerly instantiate type applications when possible, but sometimes we must delay instantiations to support recursive references (e.g. `List<T>` which contains a pointer to the `next`, which is another `List<T>`).
5. We rule out problematic recursive dependencies between structs by adapting the notion of "expanded cycles" used by C# for generics.

You can take a look at the accompanying PR with the tests.

What is _not_ in this PR:
1. Support for nested structs (haven't looked at those).
2. Serialization of generic structs.
3. Interop with C code (will come in new PR)

### DIFF
--- a/include/clang/AST/ASTContext.h
+++ b/include/clang/AST/ASTContext.h
@@ -286,6 +286,14 @@ private:
   llvm::DenseMap<std::pair<const RecordDecl *, ArrayRef<const Type *> >, RecordDecl *>
     CachedTypeApps;
 
+  // Mapping from record decls to list of delayed type applications.
+  // The key is a declaration or definition of the generic RecordDecl, and the
+  // corresponding values all have the given RecordDecl as base.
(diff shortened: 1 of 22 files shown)


abeln:
Will re-open after pushing to a branch within microsoft/checkedc-clang

PR #638, opened 2019-07-19
Add generic structs to Checked C

abeln:
This PR adds generic structs to Checked C.

Generic structs can be defined and then used like so
```C
struct List _For_any(T) {
  T *head;
  struct List<T> *next;
};

struct List<int> *lst;
int *h = lst->head;
lst = lst->next;
```

Here's a summary of what's in these changes:

1. Changed the parser so we can handle both type parameters and type arguments (type application).
2. Declarations and definitions of generic structs are represented as `RecordDecls` that store a list of type parameters.
3. Type applications (uses of generic structs) are also represented with `RecordDecls` containing the "base" `RecordDecl` as well as a list of type arguments (e.g. `List<int>` points to both `List` and `[int]`).
4. We eagerly instantiate type applications when possible, but sometimes we must delay instantiations to support recursive references (e.g. `List<T>` which contains a pointer to the `next`, which is another `List<T>`).
5. We rule out problematic recursive dependencies between structs by adapting the notion of "expanded cycles" used by C# for generics.

You can take a look at the accompanying PR with the tests: https://github.com/microsoft/checkedc/pull/370

What is _not_ in this PR:
1. Support for nested structs (haven't looked at those).
2. Serialization of generic structs.
3. Interop with C code (will come in new PR)


abeln:
@dtarditi fyi all CI tests now pass for this PR

lib/Parse/ParseDeclCXX.cpp
```diff
@@ -2009,6 +2036,33 @@ void Parser::ParseClassSpecifier(tok::TokenKind TagTokKind,
...
+///
+/// generic-struct-instantiation
+///   '<' type-name-list '>'
+///
+/// type-name-list
+///   type-name type-name-list-suffix [opt]
+///
+///  type-name-list-suffix
+///    ',' type-name type-name-list-suffix [opt]
+DeclResult Parser::ParseRecordTypeApplication(RecordDecl* Base) {
```
mgrang: `T* x ==> T *x`

lib/Parse/ParseExpr.cpp
```diff
@@ -3297,8 +3293,18 @@ ExprResult Parser::ParseGenericTypeArgumentList(ExprResult Res, SourceLocation L
   }
   ConsumeToken(); // consume '>' token
 
-  auto TypeArgs = ArrayRef<DeclRefExpr::GenericInstInfo::TypeArgument>(typeArgumentInfos);
-  return Actions.ActOnTypeApplication(Res, Loc, TypeArgs);
+  return std::make_pair<>(false, typeArgumentInfos);
```
mgrang: I think you can omit the <> in std::make_pair?

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
 
 }
 
+// Type Instantiation
+RecordDecl* Sema::ActOnRecordTypeApplication(RecordDecl *Base, ArrayRef<TypeArgument> TypeArgs) {
```
mgrang: `T* x ==> T *x`

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+  assert(Base->isGeneric() && "Base decl must be generic in a type application");
+  assert(Base->getCanonicalDecl() == Base && "Base must be canonical decl");
+ 
+  auto &ctx = Base->getASTContext();
+
+  // Unwrap the type arguments from a 'TypeArgument' to the underlying 'Type *'.
+  auto RawArgs = new (ctx) llvm::SmallVector<const Type *, 4>();
+  for (auto TArg : TypeArgs) {
```
mgrang: You can omit the braces for the for loop.

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+    RawArgs->push_back(TArg.typeName.getTypePtr());
+  }
+
+  // If possible, just retrieve the application from the cache.
+  // This is needed not only for performance, but for correctness to handle
+  // recursive references in type applications (e.g. a list which contains a list as a field).
+  if (auto Cached = ctx.getCachedTypeApp(Base, *RawArgs)) {
```
mgrang: Same here. Can omit braces.

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+    return Cached;   
+  }
+
+  // Notice we pass dummy location arguments, since the type application doesn't exist in user code.
+  RecordDecl* Inst = RecordDecl::Create(ctx, Base->getTagKind(), Base->getDeclContext(), SourceLocation(), SourceLocation(),
```
mgrang: `T* x ==> T *x`

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+  assert(Incomplete->isInstantiated() && "Only instantiated record decls can be completed");
+  assert(Incomplete->field_empty() && "Can't complete record decl with non-empty fields");
+
+  auto Defn = Incomplete->baseDecl()->getDefinition();
+  assert(Defn && "The record definition should be populated at this point");
+  for (auto Field : Defn->fields()) {
+    QualType InstType = SubstituteTypeArgs(Field->getType(), Incomplete->typeArgs());
+    assert(!InstType.isNull() && "Subtitution of type args failed!");
+    // TODO(abeln): are TypeSouceInfo and InstType in sync?
+    FieldDecl* NewField = FieldDecl::Create(Field->getASTContext(), Incomplete, SourceLocation(), SourceLocation(),
```
mgrang: `T* x ==> T *x`

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+    }
+
+    bool VisitTypedefType(const TypedefType *Type) {
+      return Visit(Type->desugar().getTypePtr());
+    }
+
+    bool VisitRecordType(const RecordType *Type) {
+      auto RDecl = Type->getDecl();
+      if (!RDecl->isInstantiated()) return false;
+      for (auto TArg : RDecl->typeArgs()) {
```
mgrang: Can omit braces.

lib/Sema/SemaExpr.cpp
```diff
@@ -5692,7 +5692,7 @@ ExprResult Sema::ActOnCallExpr(Scope *Scope, Expr *Fn, SourceLocation LParenLoc,
       }
       // In an unchecked scope, pass void types as the type arguments.
       if (DeclRefExpr *DR = dyn_cast<DeclRefExpr>(Fn)) {
-        SmallVector<DeclRefExpr::GenericInstInfo::TypeArgument, 4>
+        SmallVector<TypeArgument, 4>
```
mgrang: Please move typeArgumentInfos to this line.

lib/Sema/SemaDecl.cpp
```diff
@@ -15271,7 +15271,8 @@ Decl *Sema::ActOnTag(Scope *S, unsigned TagSpec, TagUseKind TUK,
                      SourceLocation ScopedEnumKWLoc,
                      bool ScopedEnumUsesClassTag, TypeResult UnderlyingType,
                      bool IsTypeSpecifier, bool IsTemplateParamOrArg,
-                     SkipBodyInfo *SkipBody) {
+                     SkipBodyInfo *SkipBody,
+                     ArrayRef<TypedefDecl*> TypeParams) {
```
mgrang: `T* ==> T *`

lib/Sema/DeclSpec.cpp
```diff
@@ -1019,7 +1019,7 @@ bool DeclSpec::setFunctionSpecChecked(SourceLocation Loc,
   return false;
 }
 
-bool DeclSpec::setFunctionSpecForany(SourceLocation Loc,
+bool DeclSpec::setSpecForany(SourceLocation Loc,
```
mgrang: While you are at it, please fix the indentation for the arguments of this function.

lib/Sema/DeclSpec.cpp
```diff
@@ -1037,7 +1037,7 @@ bool DeclSpec::setFunctionSpecForany(SourceLocation Loc,
   return false;
 }
 
-bool DeclSpec::setFunctionSpecItypeforany(SourceLocation Loc,
+bool DeclSpec::setSpecItypeforany(SourceLocation Loc,
```
mgrang: Ditto.

lib/Sema/CheckedCSubst.cpp
```diff
@@ -162,28 +399,47 @@ class TypeApplication : public TreeTransform<TypeApplication> {
...
 
-   const FunctionProtoType *FPT = TransformedQT->getAs<FunctionProtoType>();
-   FunctionProtoType::ExtProtoInfo EPI = FPT->getExtProtoInfo();
-   EPI.GenericFunction = false;
-   EPI.ItypeGenericFunction = false;
-   EPI.NumTypeVars = 0;
-   QualType Result =
-     Context.getFunctionType(FPT->getReturnType(), FPT->getParamTypes(), EPI);
-    return Result;
+   if (const FunctionProtoType * FPT = TransformedQT->getAs<FunctionProtoType>()) {
```
mgrang: `T * x ==> T *x`

lib/Sema/CheckedCSubst.cpp
```diff
@@ -162,28 +399,47 @@ class TypeApplication : public TreeTransform<TypeApplication> {
     }
     return Result;
   }
+
+  Decl* TransformDecl(SourceLocation Loc, Decl* D) {
```
mgrang: `T* x ==> T *x`

include/clang/AST/ASTContext.h
```diff
@@ -3074,6 +3089,29 @@ class ASTContext : public RefCountedBase<ASTContext> {
...
+  RecordDecl *getCachedTypeApp(const RecordDecl *Base, ArrayRef<const Type *> TypeArgs);
+
+  /// Add the instantiated record type 'Inst' as the result of the type application 'Base<TypeArgs>'.
+  /// Cached applications shouldn't be overwritten, so this should be called at most once per key.
+  void addCachedTypeApp(const RecordDecl *Base, ArrayRef<const Type *> TypeArgs, RecordDecl *Inst);
+
+  /// Get the list of type applications that have 'Base' as their base RecordDecl.
+  ArrayRef<RecordDecl *> getDelayedTypeApps(RecordDecl *Base);
+
+  /// Add 'TypeApp' to the cache using its base field as key.
```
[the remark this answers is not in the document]
sunnychatterjee: >>>llvm::DenseMap<RecordDecl *, llvm::SmallVector<RecordDecl *, 4>* >

I'm curious, if it's a standard practice to use hardcode numeric value inside of llvm::SmallVector. Typically, I'd expect to use a constexpr in such cases.



    DelayedTypeApps;

include/clang/AST/Decl.h
```diff
@@ -3772,12 +3772,18 @@ class RecordDecl : public TagDecl {
 protected:
   RecordDecl(Kind DK, TagKind TK, const ASTContext &C, DeclContext *DC,
              SourceLocation StartLoc, SourceLocation IdLoc,
-             IdentifierInfo *Id, RecordDecl *PrevDecl);
+             IdentifierInfo* Id, RecordDecl* PrevDecl,
+             ArrayRef<TypedefDecl*> TypeParams = ArrayRef<TypedefDecl*>(nullptr, (size_t)0),
```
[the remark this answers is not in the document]
sunnychatterjee: Can we use an explicit static_cast<size_t>(0) in this case rather than a C-style cast?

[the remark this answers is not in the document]
sunnychatterjee: Nit: The order of the new arguments could be: GenericBaseDecl, TypeParams, TypeArgs as opposed to TypeParams, GenericBaseDecl, TypeArgs.

include/clang/AST/Decl.h
```diff
@@ -3966,9 +3972,70 @@ class RecordDecl : public TagDecl {
   /// nullptr is returned if no named data member exists.
   const FieldDecl *findFirstNamedDataMember() const;
 
+  // Checked C
+
+  /// Whether the record is generic.
+  bool isGeneric();
+  /// Returns the record's type parameters.
+  /// If there are no type parameters, then the array will be empty.
+  ArrayRef<TypedefDecl*> typeParams();
```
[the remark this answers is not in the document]
sunnychatterjee: bool isGeneric() can be a 'const' method, no?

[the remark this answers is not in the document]
sunnychatterjee: There are other methods below it that can be marked as `const` as well.

include/clang/AST/Decl.h
```diff
@@ -3966,9 +3972,70 @@ class RecordDecl : public TagDecl {
...
+  // is either generic or instantiated (we don't support partial instantiations).
+
+  /// Whether this struct is generic.
+  bool IsGeneric;
+  /// The number of type parameters the record has.
+  size_t NumTypeParams;
+  /// New []'d array of pointers to TypedefDecls for the type
+  /// parameters of this record.  This is null if not a generic record.
+  TypedefDecl **TypeParams;
+  /// Sets the type parameters for the record.
```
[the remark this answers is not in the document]
sunnychatterjee: Clarification: // Whether the record is a (fully) instantiated generic.

Do we support partial specializations at all?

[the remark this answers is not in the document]
sunnychatterjee: Nit: "baseDecl()" --> Suggest renaming to "genericBaseDecl"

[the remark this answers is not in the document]
sunnychatterjee: A couple of questions:
1. Is there a reason to not use an STL container like vector/list/deque for storing the TypeParams as opposed to storing the "NumTypeParams" and "TypeParams" separately.
2. Can we use std::unique_ptr<TypedefDecl *> for the TypeParams so we don't have to manage memory?

include/clang/AST/Decl.h
```diff
@@ -3966,9 +3972,70 @@ class RecordDecl : public TagDecl {
...
+  /// The underlying generic RecordDecl that was instantiated.
+  RecordDecl *BaseDecl;
+  /// The number of type arguments in this instantiated struct (must match the number of parameters
+  /// in the underlying generic base struct).
+  size_t NumTypeArgs;
+  /// New []'d array of pointers to the type arguments. This is null if this RecordDecl
+  /// isn't the result of an instantiation.
+  TypeArgument *TypeArgs;
+  /// Sets the type arguments for an instantiation, as well as 'IsInstantiated' and 'NumTypeArgs'.
+  void setTypeArgs(const ASTContext& C, RecordDecl *BaseDecl, ArrayRef<TypeArgument> NewTypeArgs);
```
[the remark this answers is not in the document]
sunnychatterjee: This comment is similar to TypeParams. Is there a reason to not use an STL container for storing TypeArgs as opposed to using NumArgs and TypeArgs?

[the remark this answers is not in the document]
sunnychatterjee: void setTypeArgs(const ASTContext& C, RecordDecl *BaseDecl, ArrayRef<TypeArgument> NewTypeArgs);

If this is purely a setter, you should consider using: const RecordDecl *BaseDecl.

include/clang/Sema/DeclSpec.h
```diff
@@ -393,8 +393,8 @@ class DeclSpec {
 
   TypedefDecl **TypeVarInfo;
   unsigned NumTypeVars : 15;
-  bool GenericFunction : 1;
-  bool ItypeGenericFunction : 1;
+  bool GenericFunctionOrStruct : 1;
```
sunnychatterjee: Is there a reason to combine the bit fields for generic function and struct and not keep them separate for more granularity? Similarly, for the methods below.

lib/AST/ASTContext.cpp
```diff
@@ -11182,3 +11182,46 @@ APFixedPoint ASTContext::getFixedPointMin(QualType Ty) const {
...
   return APFixedPoint::getMin(getFixedPointSemantics(Ty));
 }
+
+RecordDecl *ASTContext::getCachedTypeApp(const RecordDecl *Base, ArrayRef<const Type *> TypeArgs) {
+  auto it = CachedTypeApps.find(std::make_pair(Base, TypeArgs));
+  if (it == CachedTypeApps.end()) return nullptr;
+  return it->second;
+}
+
+void ASTContext::addCachedTypeApp(const RecordDecl *Base, ArrayRef<const Type *> TypeArgs, RecordDecl *Inst) {
```
sunnychatterjee: Can RecordDecl *Inst be marked as const?

sunnychatterjee: Is it safe to assume "Base" and "Inst" are never going to be nullptrs? Perhaps, we should either dynamically check for null (defense in depth) or add asserts.

Similar observation for methods below in the PR.

lib/AST/ASTContext.cpp
```diff
@@ -11182,3 +11182,46 @@ APFixedPoint ASTContext::getFixedPointMin(QualType Ty) const {
   assert(Ty->isFixedPointType());
   return APFixedPoint::getMin(getFixedPointSemantics(Ty));
 }
+
+RecordDecl *ASTContext::getCachedTypeApp(const RecordDecl *Base, ArrayRef<const Type *> TypeArgs) {
+  auto it = CachedTypeApps.find(std::make_pair(Base, TypeArgs));
```
sunnychatterjee: Nit: Can we use: "const auto" for the iterator? Similar observation in the methods below.

lib/AST/ASTContext.cpp
```diff
@@ -11182,3 +11182,46 @@ APFixedPoint ASTContext::getFixedPointMin(QualType Ty) const {
...
+}
+
+void ASTContext::addDelayedTypeApp(RecordDecl *TypeApp) {
+  assert(TypeApp->isInstantiated() && TypeApp->isDelayedTypeApp() && "Expected a delayed type application");
+  auto Base = TypeApp->baseDecl();
+  assert(Base->getCanonicalDecl() == Base && "Base should be a canonical decl");
+  llvm::SmallVector<RecordDecl *, 4> *Delayed = nullptr;
+  auto Iter = DelayedTypeApps.find(Base);
+  if (Iter == DelayedTypeApps.end()) {
+    Delayed = new (*this) llvm::SmallVector<RecordDecl *, 4>();
```
sunnychatterjee: Clarification: Who deletes the memory for "Delayed"? Can we get rid of the "new" for "SmallVector"?

lib/AST/ASTDumper.cpp
```diff
@@ -1567,7 +1567,7 @@ void ASTDumper::VisitCastExpr(const CastExpr *Node) {
 void ASTDumper::VisitDeclRefExpr(const DeclRefExpr *Node) {
   if (Node->GetTypeArgumentInfo() &&
       !Node->GetTypeArgumentInfo()->typeArgumentss().empty()) {
-    for (DeclRefExpr::GenericInstInfo::TypeArgument tn :
+    for (TypeArgument tn :
```
sunnychatterjee: You can use: for (const auto tn : ...)

lib/AST/Decl.cpp
```diff
@@ -4154,8 +4154,11 @@ unsigned EnumDecl::getODRHash() {
 RecordDecl::RecordDecl(Kind DK, TagKind TK, const ASTContext &C,
                        DeclContext *DC, SourceLocation StartLoc,
                        SourceLocation IdLoc, IdentifierInfo *Id,
-                       RecordDecl *PrevDecl)
-    : TagDecl(DK, TK, C, DC, IdLoc, Id, PrevDecl, StartLoc) {
+                       RecordDecl *PrevDecl, ArrayRef<TypedefDecl*> TypeParams,
+                       RecordDecl *BaseDecl,
+                       ArrayRef<TypeArgument> TypeArgs)
+    : TagDecl(DK, TK, C, DC, IdLoc, Id, PrevDecl, StartLoc),
+      IsGeneric(false), NumTypeParams(0), IsInstantiated(false), NumTypeArgs(0), IsDelayed(false) {
```
sunnychatterjee: IsGeneric(false), NumTypeParams(0), IsInstantiated(false), NumTypeArgs(0), IsDelayed(false): A better way would be to use member-wise initialization for all these data members when declaring them rather than do it at the constructor scope.

lib/AST/Decl.cpp
```diff
@@ -4314,6 +4320,68 @@ const FieldDecl *RecordDecl::findFirstNamedDataMember() const {
...
+
+ArrayRef<TypedefDecl*> RecordDecl::typeParams() {
+  return { TypeParams, NumTypeParams };
+}
+
+void RecordDecl::setTypeParams(const ASTContext& C, ArrayRef<TypedefDecl*> NewTypeParams) {
+  assert(!isGeneric() && "Can't reset type parameters for record");
+  NumTypeParams = NewTypeParams.size();
+  IsGeneric = NumTypeParams > 0;
+  assert(!(isGeneric() && isInstantiated()) && "Record can't be both generic and instantiated");
```
sunnychatterjee: Clarification: Can this happen in reality? asserts don't fire on retail bits, so I was wondering if we need to throw an error for this case.

lib/Parse/ParseExpr.cpp
```diff
@@ -3255,18 +3255,14 @@ ExprResult Parser::ParseBoundsExpression() {
...
-// generic type.  However, we don't have the AST support for this yet.
-//
-// Returns false if parsing succeeded and true if an error occurred.
-ExprResult Parser::ParseGenericTypeArgumentList(ExprResult Res, SourceLocation Loc) {
-  SmallVector<DeclRefExpr::GenericInstInfo::TypeArgument, 4> typeArgumentInfos;
+// Parse a generic type argument list of the form type_name_1, ..., type_name_n '>'.
+// Notice the initial '<' must have already been parsed.
+// This is re-used for both generic functions and generic structs.
+// Return false if parsing succeeds, in which case the 'typeArgs' is also populated.
+// Return true if parsing fails.
```
sunnychatterjee: Returning false for success and true for failure is not very intuitive. Is there a reason for this choice?

lib/Parse/ParseExpr.cpp
```diff
@@ -3255,18 +3255,14 @@ ExprResult Parser::ParseBoundsExpression() {
...
+std::pair<bool, Parser::TypeArgVector> Parser::ParseGenericTypeArgumentList(SourceLocation Loc) {
+  Parser::TypeArgVector typeArgumentInfos;
+  auto err = std::make_pair<>(true, Parser::TypeArgVector());
   bool firstTypeArgument = true;
```
sunnychatterjee: bool firstTypeArgument: This can also be declared as auto.

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+RecordDecl* Sema::ActOnRecordTypeApplication(RecordDecl *Base, ArrayRef<TypeArgument> TypeArgs) {
+  assert(Base->isGeneric() && "Base decl must be generic in a type application");
+  assert(Base->getCanonicalDecl() == Base && "Base must be canonical decl");
+ 
+  auto &ctx = Base->getASTContext();
+
+  // Unwrap the type arguments from a 'TypeArgument' to the underlying 'Type *'.
+  auto RawArgs = new (ctx) llvm::SmallVector<const Type *, 4>();
+  for (auto TArg : TypeArgs) {
+    RawArgs->push_back(TArg.typeName.getTypePtr());
```
sunnychatterjee: Consider using: emplace_back

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
 
+// Type Instantiation
+RecordDecl* Sema::ActOnRecordTypeApplication(RecordDecl *Base, ArrayRef<TypeArgument> TypeArgs) {
+  assert(Base->isGeneric() && "Base decl must be generic in a type application");
+  assert(Base->getCanonicalDecl() == Base && "Base must be canonical decl");
+ 
+  auto &ctx = Base->getASTContext();
+
+  // Unwrap the type arguments from a 'TypeArgument' to the underlying 'Type *'.
+  auto RawArgs = new (ctx) llvm::SmallVector<const Type *, 4>();
```
sunnychatterjee: Is heap allocation of "RawArgs" really needed?

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+  // If possible, just retrieve the application from the cache.
+  // This is needed not only for performance, but for correctness to handle
+  // recursive references in type applications (e.g. a list which contains a list as a field).
+  if (auto Cached = ctx.getCachedTypeApp(Base, *RawArgs)) {
+    return Cached;   
+  }
+
+  // Notice we pass dummy location arguments, since the type application doesn't exist in user code.
+  RecordDecl* Inst = RecordDecl::Create(ctx, Base->getTagKind(), Base->getDeclContext(), SourceLocation(), SourceLocation(),
+    Base->getIdentifier(), Base->getPreviousDecl(), ArrayRef<TypedefDecl*>(nullptr, (size_t)0) /* TypeParams */, Base, TypeArgs);
```
sunnychatterjee: Use: static_cast<size_t>(0)

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+
+// Definitions for expanding cycles check
+namespace {
+  // A graph node is a triple '(BaseRecordDecl, TypeArgIndex, Expanding)' (here stored as two nested 'std::pairs' so we can keep them in a 'DenseSet').
+  // The semantics of a triple are as follows: a triple is in the set if, starting from one of the type arguments of 'Base', 
+  // it's possible to arrive at the type argument with index 'TypeArgIndex' which is defined in 'BaseRecordDecl'.
+  // 'Expanding' indicates whether any of the edges taken to arrive to (BaseRecordDecl, TypeArgIndex) is expanding ('Expanding = 1')
+  // or if they're all non-expanding ('Expanding = 0'). 
+  using Node = std::pair<std::pair<RecordDecl *, int>, char>;
+  const char NON_EXPANDING = 0;
```
sunnychatterjee: You can use: constexpr

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+  // applications where the variable appears. The resulting edges are "expanding" or "non-expanding"
+  // depending on whether the variable appears at the top level as a type argument.
+  //
+  // The new edges aren't returned; instead, they're added as a side effect to the
+  // 'Worklist' argument in the constructor.
+  class ExpandingEdgesVisitor : public TypeVisitor<ExpandingEdgesVisitor> {
+    // The worklist where the new nodes will be inserted.
+    std::stack<Node> &Worklist;
+    // The type variable that we're looking for in embedded type applications.
+    const TypeVariableType *TypeVar;
```
sunnychatterjee: Initialize data members.

lib/Sema/CheckedCSubst.cpp
```diff
@@ -162,28 +399,47 @@ class TypeApplication : public TreeTransform<TypeApplication> {
     }
     return Result;
   }
+
+  Decl* TransformDecl(SourceLocation Loc, Decl* D) {
+    RecordDecl *RDecl;
+    if ((RDecl = dyn_cast<RecordDecl>(D)) && RDecl->isInstantiated()) {
+      llvm::SmallVector<TypeArgument, 4> NewArgs;
+      ArrayRef<TypeArgument> OrigArgs = RDecl->typeArgs();
+      for (auto TArg = OrigArgs.begin(); TArg != OrigArgs.end(); ++TArg) {
```
sunnychatterjee: Can we use auto-for loops here?

lib/Serialization/ASTWriterStmt.cpp
```diff
@@ -495,7 +495,7 @@ void ASTStmtWriter::VisitDeclRefExpr(DeclRefExpr *E) {
 
   if (isGenericFunction || isItypeGenericFunction) {
     Record.push_back(E->GetTypeArgumentInfo()->typeArgumentss().size());
-    for (DeclRefExpr::GenericInstInfo::TypeArgument tn :
+    for (TypeArgument tn :
```
sunnychatterjee: You can use "auto" inside the for loop.

include/clang/AST/Decl.h
```diff
@@ -3772,12 +3772,18 @@ class RecordDecl : public TagDecl {
 protected:
   RecordDecl(Kind DK, TagKind TK, const ASTContext &C, DeclContext *DC,
              SourceLocation StartLoc, SourceLocation IdLoc,
-             IdentifierInfo *Id, RecordDecl *PrevDecl);
+             IdentifierInfo* Id, RecordDecl* PrevDecl,
```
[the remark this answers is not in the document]
abeln: Done

include/clang/AST/Decl.h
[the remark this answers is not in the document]
abeln: Done (both suggestions)

lib/AST/ASTContext.cpp
dtarditi: There's a region of memory associated with the AST Context.  The data until the AST Context is deallocated..  This is using placement-new to put the vector in that context.  

However, as you point out, it may be advisable to not allocate the memory in the first place.  SmallVector is designed to only heap allocate if the specified number of elements (4) is exceeded.  I believe the definition of DelayedTypesApps needs to be changed.

include/clang/AST/Decl.h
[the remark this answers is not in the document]
abeln: Done

lib/AST/Decl.cpp
dtarditi: This would be a logical error within the compiler - should never happen in reality.

include/clang/AST/Decl.h
```diff
@@ -3772,12 +3772,18 @@ class RecordDecl : public TagDecl {
...
+             ArrayRef<TypedefDecl*> TypeParams = ArrayRef<TypedefDecl*>(nullptr, (size_t)0),
+             RecordDecl *BaseDecl = nullptr,
+             ArrayRef<TypeArgument> TypeArgs = ArrayRef<TypeArgument>(nullptr, (size_t)0));
 
 public:
   static RecordDecl *Create(const ASTContext &C, TagKind TK, DeclContext *DC,
                             SourceLocation StartLoc, SourceLocation IdLoc,
-                            IdentifierInfo *Id, RecordDecl* PrevDecl = nullptr);
+                            IdentifierInfo *Id, RecordDecl* PrevDecl = nullptr,
```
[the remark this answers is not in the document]
abeln: Done

include/clang/AST/Decl.h
```diff
@@ -3966,9 +3972,70 @@ class RecordDecl : public TagDecl {
...
+  /// Whether this struct is generic.
+  bool IsGeneric;
+  /// The number of type parameters the record has.
+  size_t NumTypeParams;
+  /// New []'d array of pointers to TypedefDecls for the type
+  /// parameters of this record.  This is null if not a generic record.
+  TypedefDecl **TypeParams;
+  /// Sets the type parameters for the record.
+  /// This sets 'TypeParams', but also 'NumTypeParams' and 'IsGeneric'.
+  void setTypeParams(const ASTContext& C, ArrayRef<TypedefDecl*> NewTypeParams);
```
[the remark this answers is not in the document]
abeln: Done

lib/Parse/ParseExpr.cpp
dtarditi: The pattern is used in the existing parsing code, unfortunately.  I agree with your comment that it is unintuitive.

lib/Parse/ParseExpr.cpp
dtarditi: We should see what the clang style guide says on this one.  auto is just as long to type as  bool :).

lib/Sema/CheckedCSubst.cpp
dtarditi: I agree that it is likely not needed - good catch.

lib/Sema/CheckedCSubst.cpp
dtarditi: Do we even need the cast?

include/clang/AST/Decl.h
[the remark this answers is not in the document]
abeln: Switching to vectors simplified things considerably. Thanks.

include/clang/AST/Type.h
```diff
@@ -1445,6 +1446,13 @@ class BoundsAnnotations {
   void Profile(llvm::FoldingSetNodeID &ID, const ASTContext &Ctx) const;
 };
 
+/// Checked C: an argument in an instantiation of a generic function
+/// or record.
+struct TypeArgument {
+  QualType typeName;
+  TypeSourceInfo* sourceInfo;
```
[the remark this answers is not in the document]
abeln: Done

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -9990,5 +9990,9 @@ def err_bounds_type_annotation_lost_checking : Error<
   def err_type_args_limited : Error<
     "type arguments only supported for variables and function names right now">;
 
+  // TODO(abeln): describe cycle in error message
```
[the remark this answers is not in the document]
abeln: Done

include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -9990,5 +9990,9 @@ def err_bounds_type_annotation_lost_checking : Error<
...
+  def err_expanding_cycle : Error<
+	"expanding cycle in struct definition">;
```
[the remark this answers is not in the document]
abeln: Done

include/clang/Parse/Parser.h
```diff
@@ -2642,7 +2647,7 @@ class Parser : public CodeCompletionHandler {
   void ParseAtomicSpecifier(DeclSpec &DS);
   void ParseCheckedPointerSpecifiers(DeclSpec & DS);
   void ParseForanySpecifier(DeclSpec &DS);
-  bool ParseGenericFunctionSpecifierHelper(DeclSpec &DS,
+  bool ParseForanySpecifierHelper(DeclSpec &DS,
```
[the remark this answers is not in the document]
abeln: Done

include/clang/Sema/Sema.h
```diff
@@ -2249,7 +2249,8 @@ class Sema {
                  bool &IsDependent, SourceLocation ScopedEnumKWLoc,
                  bool ScopedEnumUsesClassTag, TypeResult UnderlyingType,
                  bool IsTypeSpecifier, bool IsTemplateParamOrArg,
-                 SkipBodyInfo *SkipBody = nullptr);
+                 SkipBodyInfo *SkipBody = nullptr,
+                 ArrayRef<TypedefDecl*> TypeParams = ArrayRef<TypedefDecl*> { nullptr, 0} );
```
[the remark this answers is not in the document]
abeln: Done

include/clang/Sema/Sema.h
```diff
@@ -4872,11 +4873,50 @@ class Sema {
...
   BoundsExpr *CheckNonModifyingBounds(BoundsExpr *Bounds, Expr *E);
 
-  ExprResult ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
-                     ArrayRef<DeclRefExpr::GenericInstInfo::TypeArgument> Args);
+  ExprResult ActOnFunctionTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
+                     ArrayRef<TypeArgument> Args);
 
-  QualType SubstituteTypeArgs(QualType QT,
-                ArrayRef<DeclRefExpr::GenericInstInfo::TypeArgument> TypeArgs);
+  RecordDecl* ActOnRecordTypeApplication(RecordDecl* Base, ArrayRef<TypeArgument> TypeArgs);
```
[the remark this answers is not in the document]
abeln: Done

lib/AST/Decl.cpp
```diff
@@ -4167,13 +4170,17 @@ RecordDecl::RecordDecl(Kind DK, TagKind TK, const ASTContext &C,
...
   setParamDestroyedInCallee(false);
   setArgPassingRestrictions(APK_CanPassInRegs);
+  setTypeParams(C, TypeParams);
+  setTypeArgs(C, BaseDecl, TypeArgs);
 }
 
 RecordDecl *RecordDecl::Create(const ASTContext &C, TagKind TK, DeclContext *DC,
                                SourceLocation StartLoc, SourceLocation IdLoc,
-                               IdentifierInfo *Id, RecordDecl* PrevDecl) {
+                               IdentifierInfo *Id, RecordDecl* PrevDecl, ArrayRef<TypedefDecl*> TypeParams,
```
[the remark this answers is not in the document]
abeln: Done

lib/AST/Decl.cpp
```diff
@@ -4314,6 +4320,68 @@ const FieldDecl *RecordDecl::findFirstNamedDataMember() const {
...
 
+// Checked C
+
+// Type Parameters
+
+bool RecordDecl::isGeneric() {
+  return IsGeneric;
+}
+
+ArrayRef<TypedefDecl*> RecordDecl::typeParams() {
```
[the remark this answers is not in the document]
abeln: Done

lib/AST/Decl.cpp
```diff
@@ -4314,6 +4320,68 @@ const FieldDecl *RecordDecl::findFirstNamedDataMember() const {
...
+  return { TypeParams, NumTypeParams };
+}
+
+void RecordDecl::setTypeParams(const ASTContext& C, ArrayRef<TypedefDecl*> NewTypeParams) {
```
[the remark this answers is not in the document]
abeln: This method no longer exists

lib/AST/Decl.cpp
```diff
@@ -4314,6 +4320,68 @@ const FieldDecl *RecordDecl::findFirstNamedDataMember() const {
...
+  this->IsDelayed = IsDelayed;
+}
+
+void RecordDecl::setTypeArgs(const ASTContext& C, RecordDecl *NewBaseDecl, ArrayRef<TypeArgument> NewTypeArgs) {
+  assert(!isInstantiated() && "Can't reset type parameters for record");
+  BaseDecl = NewBaseDecl;
+  NumTypeArgs = NewTypeArgs.size();
+  IsInstantiated = NumTypeArgs > 0;
+  assert(!(isGeneric() && isInstantiated()) && "Record can't be both generic and instantiated");
+  assert(((IsInstantiated && BaseDecl) || !(IsInstantiated || BaseDecl)) && "Must provide both base decl and type arguments, or neither");
```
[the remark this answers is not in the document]
abeln: Done (moved to RecorDecl constructor)

lib/Parse/ParseDeclCXX.cpp
```diff
@@ -1646,6 +1646,15 @@ void Parser::ParseClassSpecifier(tok::TokenKind TagTokKind,
   const PrintingPolicy &Policy = Actions.getASTContext().getPrintingPolicy();
   Sema::TagUseKind TUK;
 
+  // Checked C - handle generic structs.
+  if (Tok.is(tok::kw__For_any)) {
+    // TODO(abeln): add error handling here
```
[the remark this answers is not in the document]
abeln: Done

lib/Parse/ParseDeclCXX.cpp
```diff
@@ -1913,15 +1922,33 @@ void Parser::ParseClassSpecifier(tok::TokenKind TagTokKind,
...
         DSC == DeclSpecContext::DSC_template_param ||
             DSC == DeclSpecContext::DSC_template_type_arg,
-        &SkipBody);
+        &SkipBody,
+        DS.typeVariables());
+
+    // Checked C: a reference to a struct can be followed by a list of type arguments,
+    // if the struct is generic.
+    if (TagOrTempResult.isUsable() && TUK == Sema::TUK_Reference) {
+      RecordDecl* Decl = llvm::dyn_cast<RecordDecl>(TagOrTempResult.get()->getCanonicalDecl());
```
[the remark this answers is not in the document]
abeln: Done

lib/Parse/ParseDeclCXX.cpp
```diff
@@ -1913,15 +1922,33 @@ void Parser::ParseClassSpecifier(tok::TokenKind TagTokKind,
 
     stripTypeAttributesOffDeclSpec(attrs, DS, TUK);
 
+    // Checked C: if the struct has type parameters, then they introduced a new scope, so we
+    // must make sure to add the struct to the parent scope instead.
+    // The type parameters scope is removed later in 'ParseDeclOrFunctionDefInternal'.
+    assert(getCurScope()->isForanyScope() == DS.isForanySpecified());
+    Scope* structScope = getCurScope()->isForanyScope() ? getCurScope()->getParent() : getCurScope();
```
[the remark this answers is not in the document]
abeln: Done

lib/Parse/ParseDeclCXX.cpp
abeln: Done

lib/Parse/ParseExpr.cpp
abeln: Yep. Done.

lib/Sema/CheckedCSubst.cpp
abeln: Done

lib/Sema/CheckedCSubst.cpp
abeln: I personally like the braces, because they make the scope clearer: c.f.

for (auto TArg : TypeArgs)
  doSomething(); // ok: run with the loop
  doSomethingElse(); // looks like it's run as part of the loop, but isn't

lib/Sema/CheckedCSubst.cpp
abeln: Ditto

lib/Sema/CheckedCSubst.cpp
abeln: Done.

lib/Sema/CheckedCSubst.cpp
abeln: Done

lib/Sema/CheckedCSubst.cpp
abeln: Ditto

lib/Sema/SemaExpr.cpp
abeln: Done.

lib/Sema/DeclSpec.cpp
abeln: Done

lib/Sema/DeclSpec.cpp
abeln: Done.

lib/Sema/SemaDecl.cpp
abeln: Done

lib/Sema/CheckedCSubst.cpp
abeln: Done

lib/Sema/CheckedCSubst.cpp
abeln: Done

abeln:
@mgrang addressed your comments. Thanks for the feedback!

include/clang/AST/ASTContext.h
```diff
@@ -279,6 +279,21 @@ class ASTContext : public RefCountedBase<ASTContext> {
   llvm::DenseMap<const MaterializeTemporaryExpr *, APValue *>
     MaterializedTemporaryValues;
 
+  /// Mapping from (generic record decl, type arguments) pairs to instantiated record decls.
+  /// e.g. (List, int) -> List<int>
+  /// This keeps tracks of all type applications both so we can preserve the uniqueness invariant
+  /// for decls and types, and also so we can typecheck gnarly recursive applications.
```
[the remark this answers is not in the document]
abeln: Done.

include/clang/AST/ASTContext.h
```diff
@@ -279,6 +279,21 @@ class ASTContext : public RefCountedBase<ASTContext> {
...
+  llvm::DenseMap<std::pair<const RecordDecl *, ArrayRef<const Type *> >, RecordDecl *>
+    CachedTypeApps;
+
+  // Mapping from record decls to list of delayed type applications.
+  // The key is a declaration or definition of the generic RecordDecl, and the
```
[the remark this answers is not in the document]
abeln: Done.

include/clang/AST/Decl.h
```diff
@@ -3966,9 +3972,70 @@ class RecordDecl : public TagDecl {
...
+
+  /// Whether this struct is the result of instantiating a generic struct
+  /// (so the current struct is fully - instantiated no longer generic).
```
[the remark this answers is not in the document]
abeln: This comment no longer exists.

include/clang/AST/Decl.h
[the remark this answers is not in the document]
abeln: The method no longer exists. We now use std::vector to store the TypeParams.

lib/Parse/ParseDecl.cpp
```diff
@@ -4200,7 +4200,7 @@ void Parser::ParseStructUnionBody(SourceLocation RecordLoc,
   // We are guessing that most structure declarations have 4 or fewer members
   // with bounds expressions on them.
   SmallVector<BoundsExprInfo, 4> deferredBoundsExpressions;
-
+
```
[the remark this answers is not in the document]
abeln: Done

lib/Parse/ParseDecl.cpp
```diff
@@ -4353,6 +4353,34 @@ void Parser::ParseStructUnionBody(SourceLocation RecordLoc,
         FD->setInvalidDecl();
     }
   }
+
+  // Checked C: complete delayed type applications.
```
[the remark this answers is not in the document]
abeln: Done

lib/Parse/ParseDecl.cpp
```diff
@@ -4353,6 +4353,34 @@ void Parser::ParseStructUnionBody(SourceLocation RecordLoc,
...
+      //     struct List<T> *tail1;
+      //     struct List<T> *tail2;
+      //   }
+      //
+      // If all type applications are done immediately after the corresponding field
+      // is processed, then when we encounter 'List<T>' corresponding to 'tail1', we need
+      // to instantiate a 'List' record for which we don't know the set of fields
+      // (specifically, we don't know that it contains a second 'tail2' field).
+      //
+      // The solution is to initially create "dummy" 'RecordDecls' for 'tail1' and 'tail2'.
```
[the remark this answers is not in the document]
abeln: Yes, only one is created. Clarified.

lib/Parse/ParseDeclCXX.cpp
```diff
@@ -2009,6 +2036,33 @@ void Parser::ParseClassSpecifier(tok::TokenKind TagTokKind,
   }
 }
 
+/// Checked C: parse an application of the 'Base' 'RecorDecl' to a number of type
```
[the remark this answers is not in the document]
abeln: Done

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+void Sema::CompleteTypeAppFields(RecordDecl *Incomplete) {
+  assert(Incomplete->isInstantiated() && "Only instantiated record decls can be completed");
+  assert(Incomplete->field_empty() && "Can't complete record decl with non-empty fields");
+
+  auto Defn = Incomplete->baseDecl()->getDefinition();
+  assert(Defn && "The record definition should be populated at this point");
+  for (auto Field : Defn->fields()) {
+    QualType InstType = SubstituteTypeArgs(Field->getType(), Incomplete->typeArgs());
+    assert(!InstType.isNull() && "Subtitution of type args failed!");
+    // TODO(abeln): are TypeSouceInfo and InstType in sync?
```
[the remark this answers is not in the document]
abeln: Done

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+  // one of the fields refers back to the application.
+  ctx.addCachedTypeApp(Base, *RawArgs, Inst);
+
+  auto Defn = Base->getDefinition();
+  if (Defn && Defn->isCompleteDefinition()) {
+    // If the underlying definition exists and has all its fields already populated, then we can complete
+    // the type application.
+    CompleteTypeAppFields(Inst);
+  } else {
+    // If the definition isn't available, it might be because we're typing a recursive field declaration: e.g.
```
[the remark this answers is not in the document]
abeln: Done

lib/Serialization/ASTWriterStmt.cpp
abeln: Done

lib/Sema/CheckedCSubst.cpp
abeln: Yes. Done.

lib/Sema/CheckedCSubst.cpp
abeln: Done

lib/Sema/CheckedCSubst.cpp
abeln: Done

lib/Sema/CheckedCSubst.cpp
abeln: Yes, we need it. Switched to static_cast as suggested

lib/Sema/CheckedCSubst.cpp
abeln: Not needed. Allocated it on the stack.

lib/Sema/CheckedCSubst.cpp
abeln: I don't fully understand move semantics, but this is just copying a pointer, so it seems to me there should be no advantage (no constructor is invoked)?

lib/Parse/ParseExpr.cpp
abeln: auto it is

lib/Parse/ParseExpr.cpp
abeln: What David said :)

lib/AST/Decl.cpp
abeln: Done

lib/AST/ASTContext.cpp
abeln: Done

lib/AST/ASTDumper.cpp
abeln: Done

lib/AST/ASTContext.cpp
abeln: Done

lib/AST/ASTContext.cpp
abeln: > Can RecordDecl *Inst be marked as const?
Maybe, but it'd require quite a few changes: the const leaks out to the users and their callers and their callers, etc. I'll leave it be for now.

> Is it safe to assume "Base" and "Inst" are never going to be nullptrs? Perhaps, we should either dynamically check for null (defense in depth) or add asserts.
> 
> Similar observation for methods below in the PR.

Added assertions.

include/clang/Sema/DeclSpec.h
abeln: We can reuse the current logic in DeclSpec with minimal changes. If we need to disambiguate between them in the future, I propose we split them at that point.

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+}
+
 namespace {
+// LocRebuilderTransform is an uncustomized 'TreeTransform' that is used
+// solely for re-building 'TypeLocs' within 'TypeApplication'.
+// We use this vanilla transform instead of a recursive call to 'TypeApplication::TransformType' because
+// we sometimes substitute a type variable for another type variable, and in those cases we
+// want to re-build 'TypeLocs', but not do further substitutions.
+// e.g.
+//   struct Box _For_any(U) { T *x; }
```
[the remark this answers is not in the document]
abeln: Done

lib/Sema/CheckedCSubst.cpp
```diff
@@ -75,18 +76,254 @@ ExprResult Sema::ActOnTypeApplication(ExprResult TypeFunc, SourceLocation Loc,
...
+  const TypeVariableType *GetTypeVar(TypedefDecl *TDef) {
+    const TypeVariableType *TypeVar = llvm::dyn_cast<TypeVariableType>(TDef->getUnderlyingType());
+    assert(TypeVar && "Expected a type variable as the parameter of a generic record");
+    return TypeVar;
+  }
+
+  // A 'TypeVisitor' that determines whether a type references a given type variable.
+  // e.g. ContainsTypeVar('T').Visit('List<T>') -> true
+  //      ContainsTypeVar('T').Visit('List<int>') -> false
+  class ContainsTypeVarVisitor : public TypeVisitor<ContainsTypeVarVisitor, bool> {
```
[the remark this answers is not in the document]
abeln: > I think you should change this pattern to not rely on the return type. I'm not sure that the return value will be propagated for all kinds of types. The typical pattern that is used is to have a field that is set to true if the type variable is encountered. Then after invoking the traversel, you check that field.

I tried this, but since we re-use the visitor multiple times in a loop, we'd also have to add a method to "reset" the results of the previous run. I think having the visitor be stateless makes for a nicer api.

> 
> I think this wil let you delete some methods. I believe it will also handle some missing cases, like function types with a recursive type application embedded in them.

Unfortunately, TypeVisitor doesn't structurally recurse on the components of a type by default. The default implementation just returns false. We have to handle _every_ case we care about by hand.

include/clang/AST/Decl.h
```diff
@@ -3772,12 +3772,18 @@ class RecordDecl : public TagDecl {
 protected:
   RecordDecl(Kind DK, TagKind TK, const ASTContext &C, DeclContext *DC,
              SourceLocation StartLoc, SourceLocation IdLoc,
-             IdentifierInfo *Id, RecordDecl *PrevDecl);
+             IdentifierInfo* Id, RecordDecl* PrevDecl,
+             ArrayRef<TypedefDecl*> TypeParams = ArrayRef<TypedefDecl*>(nullptr, (size_t)0),
+             RecordDecl *BaseDecl = nullptr,
```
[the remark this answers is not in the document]
abeln: Renamed to GenericBaseDecl

include/clang/AST/ASTContext.h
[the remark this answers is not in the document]
abeln: > its base -> TypeApp's base. Nit: "as key" -> "as the key".
Done


> Nit: I'm curious, if it's a standard practice to use hardcoded numeric values ('4' in this case) inside of llvm::SmallVector. Typically, I'd expect to use a constexpr in such cases. There are other instances of this value being used elsewhere in this PR.

It does seem to be common practice to hardcode constants in llvm::SmallVector (see https://github.com/microsoft/checkedc-clang/search?q=%22llvm%3A%3ASmallVector%22&unscoped_q=%22llvm%3A%3ASmallVector%22).

abeln:
@dtarditi @sunnychatterjee  @mgrang  This is ready for review. PTAL.
I followed almost all the suggestions in the comments (but see https://github.com/microsoft/checkedc-clang/pull/638#discussion_r307526632).
I also added clang-specific tests in https://github.com/microsoft/checkedc-clang/pull/638/files#diff-56d34b0536d04260879053b75ae1601e

lib/Sema/CheckedCSubst.cpp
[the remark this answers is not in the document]
dtarditi: I didn't realize that TypeVisitor doesn't recursively traverse types that are components of otehr types.  In that case, I think you should switch to RecursiveASTVisitor, which does.   The current code pattern is dangerous because it is easy to miss a case.   If that happens, we won't notice the use of the type variable.  

For the re-use, we should just create a new object for each iteration of the loop.  The object doesn't need to be heap-allocated, it can be stack-allocated.

dtarditi:
One additional thing: for the TODO's that you've added, could you open a GitHub issue to track them and mark the TODO's with the issue number.  I think we'll want to come back to some of them, particularly the efficiency-related one.

lib/Parse/ParseDecl.cpp
```diff
@@ -3005,14 +3004,14 @@ static void SetupFixedPointError(const LangOptions &LangOpts,
...
 ///       declaration-specifiers: [C99 6.7]
-///         storage-class-specifier declaration-specifiers[opt]
-///         type-specifier declaration-specifiers[opt]
-/// [C99]   function-specifier declaration-specifiers[opt]
-/// [C11]   alignment-specifier declaration-specifiers[opt]
-/// [GNU]   attributes declaration-specifiers[opt]
-/// [Clang] '__module_private__' declaration-specifiers[opt]
-/// [ObjC1] '__kindof' declaration-specifiers[opt]
-///
+///            storage-class-specifier declaration-specifiers[opt]
```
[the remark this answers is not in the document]
abeln: Done

lib/Sema/CheckedCSubst.cpp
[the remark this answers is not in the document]
abeln: Switched to RecursiveASTVisitor. I was able to remove a few of the cases, but not all of them (e.g. RecursiveASTVisitor doesn't handle TypeDefType and RecordType).

I also had to move the expanding cycles visitors to their own file, because the obj file for CheckedCSubst had grown too large and couldn't be generated. I was getting
```
fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj split file
``

Splitting the file seems to solve the problem.

abeln:
@dtarditi PTAL
In particular, can you double check that https://github.com/microsoft/checkedc-clang/pull/638#discussion_r311791676 is reasonable? (had never seen the "obj file too large" error).

lib/Sema/CheckedCSubst.cpp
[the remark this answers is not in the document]
abeln: As per David's suggestion, I moved the expanding cycles code back to CheckedCSubst, and modified the Cmake file.

abeln:
@dtarditi moved back the code to CheckedCSubst as requested. Tests pass locally. Currently running dev pipeline.

PR #643, opened 2019-07-29
Improve equivalence checking of pointer arithmetic expressions

ppashakhanloo:
A new equivalence checking function has been added in this change
(EqualPointerArithOffset). This function is used for comparing two
bounds (i.e., Base + Offset) where one bound is following pointer
arithmetic form and the other one is in bytewise arithmetic form.
The assumption is that overflow checking is performed inside the *alloc function.

This pull request also contains new tests.

### DIFF
--- a/lib/Sema/SemaBounds.cpp
+++ b/lib/Sema/SemaBounds.cpp
@@ -1791,8 +1791,7 @@ namespace {
           return ProofResult::False;
         }
         if (IsLowerOffsetVariable() && R.IsLowerOffsetVariable())
-          if (EqualValue(S.Context, LowerOffsetVariable, R.LowerOffsetVariable, EquivExprs) ||
-              EqualPointerArithOffset(S.Context, Base, R.Base, LowerOffsetVariable, R.LowerOffsetVariable, EquivExprs))
+          if (EqualValue(S.Context, Base, R.Base, LowerOffsetVariable, R.LowerOffsetVariable, EquivExprs))
             return ProofResult::True;
         if (R.IsLowerOffsetVariable() && IsLowerOffsetConstant() &&
             R.LowerOffsetVariable->getType()->isUnsignedIntegerType() && LowerOffsetConstant.getExtValue() == 0)
@@ -1824,8 +1823,7 @@ namespace {
           return ProofResult::False;
         }
         if (IsUpperOffsetVariable() && R.IsUpperOffsetVariable())
-          if (EqualValue(S.Context, UpperOffsetVariable, R.UpperOffsetVariable, EquivExprs) ||
-              EqualPointerArithOffset(S.Context, Base, R.Base, UpperOffsetVariable, R.UpperOffsetVariable, EquivExprs))
+          if (EqualValue(S.Context, Base, R.Base, UpperOffsetVariable, R.UpperOffsetVariable, EquivExprs))
             return ProofResult::True;
         if (IsUpperOffsetVariable() && R.IsUpperOffsetConstant() &&
             UpperOffsetVariable->getType()->isUnsignedIntegerType() && R.UpperOffsetConstant.getExtValue() == 0)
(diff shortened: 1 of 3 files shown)


test/CheckedC/static-checking/bounds-decl-checking-cs.c
```diff
@@ -2,9 +2,8 @@
...
 // true or false.
 //
-// RUN: %clang -cc1 -fcheckedc-extension -verify %s
+// RUN: %clang_cc1 -fcheckedc-extension -verify %s
 
 #pragma CHECKED_SCOPE ON
 
-#include "bounds-decl-checking.c"
-
+#include "bounds-decl-checking.c"
```
mgrang: Please add back the newline here.

test/CheckedC/static-checking/bounds-decl-checking-cs.c
```diff
@@ -2,9 +2,8 @@
 // produces warnings about bounds declarations that are not provably
 // true or false.
 //
-// RUN: %clang -cc1 -fcheckedc-extension -verify %s
+// RUN: %clang_cc1 -fcheckedc-extension -verify %s
```
mgrang: Is there a reason to change `%clang -cc1` to `%clang_cc1`? AFAIK both of these are equivalent, no?

ppashakhanloo: Apparently, they are not the same. I needed to include <stddef.h>. It does not seem to work with `clang -cc1`. It says it cannot find stddef.h.
Please let me know if there is something better I could do here.

test/CheckedC/static-checking/bounds-decl-checking-cs.c
ppashakhanloo: Done.

test/CheckedC/static-checking/bounds-decl-checking-cs.c
mgrang: `clang` is the driver while `clang_cc1` is the clang frontend. Basically, clang driver will take in all the flags/options and will expand them into a clang -cc1 command line. So if you want to use `clang -cc1` then you would need to provide the path to the includes. So I guess what you did here is correct in that we do not need to go through the `clang` indirection and can directly invoke `clang_cc1`.

lib/Sema/SemaBounds.cpp
```diff
@@ -2058,6 +2060,126 @@ namespace {
       return BaseRange::Kind::ConstantSized;
     }
 
+    // In this function, the goal is to compare two expressions: `Base1 + Offset1` and `Base2 + Offset2`
```
dtarditi: I think a more general approach would be to try to put each arithmetic expression into a standard form involving integer arithmetic.  If that is possible, you can then compare the standard forms.   The standard form would be base + (const op e), where op is either signed multiplication or unsigned multiplication, and base, const, and e are all interpreted as integers.   You could have variables that represent each of the components of the standard form.

Given `array_ptr<int>` p:
- For `((char *) p) + e1`, the standard form is `base = ((char *) p, const = 1, e = e1.`  op depends on the signedness of e1.
- For `p + e2`. the standard form is `base = p, const = sizeof(int), e = e2`.  op depends on the signedness of `e2`.
- For `((unsigned char *) p) + e1`, the standard from is `base = (unsigned char *) p, const = 1, e = e1`.  op depends on the signedness of `e1`.

To check that two expressions to produce identical results, we would check that the multiplication operators have the same signedness.   At this time, I don't think we should try to handle the case where one multiplication operator is signed and the other is unsigned (we would need additional facts to prove that the two operations produce the same value: that the signed integer is >= 0 and that the multiplication does not overflow).

ppashakhanloo:
@dtarditi I made some changes to the code. It is ready for review.
There is now a new function CreateStandardForm(...) which basically expands the pointer arithmetic to bytewise arithmetic. I added some new tests and updated some comments as well.

lib/Sema/SemaBounds.cpp
```diff
@@ -2058,6 +2058,97 @@ namespace {
...
+    // 1. For (char *)p + e1 or (unsigned char *)p + e1, ConstantPart = 1, VariablePart = e1.
+    // 2. For p + e1, ConstantPart = sizeof(T), VariablePart = e1.
+    //
+    // Note that another way to interpret the functionality of this function is that it expands
+    // pointer arithmetic to bytewise arithmetic.
+    static bool CreateStandardForm(ASTContext &Ctx, Expr *Base, Expr *Offset, llvm::APSInt &ConstantPart, bool &IsOpSigned, Expr *&VariablePart) {
+      ConstantPart = llvm::APSInt(Ctx.getTargetInfo().getPointerWidth(0), false);
+
+      if (Base->getType()->getPointeeOrArrayElementType()->getCanonicalTypeInternal()->isCharType()) {
+        if (BinaryOperator *BO = dyn_cast<BinaryOperator>(Offset->IgnoreParenCasts())) {
```
dtarditi: I think you want to call IgnoreParens() here.   The code should not ignore other casts.

lib/Sema/SemaBounds.cpp
```diff
@@ -2058,6 +2058,97 @@ namespace {
...
+          if (BO->getRHS()->IgnoreParenCasts()->isIntegerConstantExpr(ConstantPart, Ctx))
+            VariablePart = BO->getLHS()->IgnoreParenCasts();
```
dtarditi: I think you can delete the call to IgnoreParenCasts().   The code should not ignore casts, and it does not need to ignore parenthesis.

lib/Sema/SemaBounds.cpp
```diff
@@ -2058,6 +2058,97 @@ namespace {
...
+    // 2. For p + e1, ConstantPart = sizeof(T), VariablePart = e1.
+    //
+    // Note that another way to interpret the functionality of this function is that it expands
+    // pointer arithmetic to bytewise arithmetic.
+    static bool CreateStandardForm(ASTContext &Ctx, Expr *Base, Expr *Offset, llvm::APSInt &ConstantPart, bool &IsOpSigned, Expr *&VariablePart) {
+      ConstantPart = llvm::APSInt(Ctx.getTargetInfo().getPointerWidth(0), false);
+
+      if (Base->getType()->getPointeeOrArrayElementType()->getCanonicalTypeInternal()->isCharType()) {
+        if (BinaryOperator *BO = dyn_cast<BinaryOperator>(Offset->IgnoreParenCasts())) {
+          if (BO->getRHS()->IgnoreParenCasts()->isIntegerConstantExpr(ConstantPart, Ctx))
```
dtarditi: You can remove the IgnoreParenCasts() here.

ppashakhanloo: Done

lib/Sema/SemaBounds.cpp
ppashakhanloo: Done

lib/Sema/SemaBounds.cpp
ppashakhanloo: Done

lib/Sema/SemaBounds.cpp
```diff
@@ -2058,6 +2058,97 @@ namespace {
...
+    // Given array_ptr<T> p:
+    // 1. For (char *)p + e1 or (unsigned char *)p + e1, ConstantPart = 1, VariablePart = e1.
+    // 2. For p + e1, ConstantPart = sizeof(T), VariablePart = e1.
+    //
+    // Note that another way to interpret the functionality of this function is that it expands
+    // pointer arithmetic to bytewise arithmetic.
+    static bool CreateStandardForm(ASTContext &Ctx, Expr *Base, Expr *Offset, llvm::APSInt &ConstantPart, bool &IsOpSigned, Expr *&VariablePart) {
+      ConstantPart = llvm::APSInt(Ctx.getTargetInfo().getPointerWidth(0), false);
+
+      if (Base->getType()->getPointeeOrArrayElementType()->getCanonicalTypeInternal()->isCharType()) {
```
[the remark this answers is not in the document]
ppashakhanloo: Done

lib/Sema/SemaBounds.cpp
```diff
@@ -2058,6 +2058,97 @@ namespace {
...
+          ConstantPart = llvm::APSInt(ConstantPart, false);
+          if (ConstantPart.getBitWidth() < Ctx.getTargetInfo().getPointerWidth(0))
+            ConstantPart = ConstantPart.extend(Ctx.getTargetInfo().getPointerWidth(0));
+        } else
+          goto exit;
+        return true;
+
+      exit:
+        VariablePart = Offset->IgnoreParenCasts();
+        ConstantPart = 1;
```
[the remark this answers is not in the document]
ppashakhanloo: Great, I was very unhappy with the way that I had found! I replaced it with your suggestion.

ppashakhanloo:
@dtarditi I have addressed all the above comments. The code is ready for review.
There are some more updates that can be found in checkedc repo. I opened a pull request at [microsoft/checkedc](https://github.com/microsoft/checkedc/pull/376).

Issue #237, opened 2017-03-28
Add non-null checks to pointer arithmetic


lenary:
After some inspection, the call to emit a check should go roughly here
 
https://github.com/Microsoft/checkedc-clang/blob/master/lib/CodeGen/CGExprScalar.cpp#L2454-L2456

The reason this isn't trivial is that any `array_ptr<T> arr : count(x)` bounds are translated into `array_ptr<T> arr : bounds(arr, arr + x)`, and we have to do code generation for both the lower and the upper bounds expression. For the upper expression in particular, this will mean we add a non-null check to `arr` (the "base" pointer). We know this check is redundant, as we do a non-null check of `arr` before we do the range check. 

The sensible fix is probably to disable these non-null checks when generating the code for the upper and lower expressions. It is worth thinking about the soundness of this fix, in particular with respect to range checks that include pointer arithmetic on pointers that aren't the base pointer.

mgrang:
PR #647

mgrang:
> PR #647

PR #647 is subsumed by https://github.com/microsoft/checkedc-clang/pull/663

PR #647, opened 2019-07-30
Add a dynamic check for pointer arithmetic of null pointers

mgrang:
See issue #237

### DIFF
--- a/include/clang/Basic/CodeGenOptions.def
+++ b/include/clang/Basic/CodeGenOptions.def
@@ -364,6 +364,8 @@ CODEGENOPT(BranchTargetEnforcement, 1, 0)
 /// Whether to emit unused static constants.
 CODEGENOPT(KeepStaticConsts, 1, 0)
 
+CODEGENOPT(CheckedCRuntimeChecks, 1, 0) ///< Insert Checked C runtime checks.
+
 #undef CODEGENOPT
 #undef ENUM_CODEGENOPT
 #undef VALUE_CODEGENOPT
--- a/include/clang/Basic/CodeGenOptions.def
+++ b/include/clang/Basic/CodeGenOptions.def
@@ -364,8 +364,8 @@ CODEGENOPT(BranchTargetEnforcement, 1, 0)
 /// Whether to emit unused static constants.
 CODEGENOPT(KeepStaticConsts, 1, 0)
 
-/// Whether to add runtime checks for checkedc.
-CODEGENOPT(CheckedCRuntimeChecks, 1, 1)
+/// Whether to add null ptr checks for checkedc.
+CODEGENOPT(CheckedCNullPtrChecks, 1, 1)
 
 #undef CODEGENOPT
 #undef ENUM_CODEGENOPT
--- a/include/clang/Basic/CodeGenOptions.def
+++ b/include/clang/Basic/CodeGenOptions.def
@@ -365,7 +365,7 @@ CODEGENOPT(BranchTargetEnforcement, 1, 0)
 CODEGENOPT(KeepStaticConsts, 1, 0)
 
 /// Whether to add runtime checks for checkedc.
-CODEGENOPT(CheckedCRuntimeChecks, 1, 0)
+CODEGENOPT(CheckedCRuntimeChecks, 1, 1)
 
 #undef CODEGENOPT
 #undef ENUM_CODEGENOPT
--- a/include/clang/Basic/CodeGenOptions.def
+++ b/include/clang/Basic/CodeGenOptions.def
@@ -364,7 +364,7 @@ CODEGENOPT(BranchTargetEnforcement, 1, 0)
 /// Whether to emit unused static constants.
 CODEGENOPT(KeepStaticConsts, 1, 0)
 
-/// Whether to add null ptr checks for checkedc.
+/// Whether to add null pointer checks for Checked C.
 CODEGENOPT(CheckedCNullPtrChecks, 1, 1)
 
 #undef CODEGENOPT
--- a/include/clang/Basic/CodeGenOptions.def
+++ b/include/clang/Basic/CodeGenOptions.def
@@ -364,7 +364,8 @@ CODEGENOPT(BranchTargetEnforcement, 1, 0)
 /// Whether to emit unused static constants.
 CODEGENOPT(KeepStaticConsts, 1, 0)
 
-CODEGENOPT(CheckedCRuntimeChecks, 1, 0) ///< Insert Checked C runtime checks.
+/// Whether to add runtime checks for checkedc.
+CODEGENOPT(CheckedCRuntimeChecks, 1, 0)
 
 #undef CODEGENOPT
 #undef ENUM_CODEGENOPT
--- a/include/clang/Driver/Options.td
+++ b/include/clang/Driver/Options.td
@@ -2621,6 +2621,12 @@ def fintegrated_as : Flag<["-"], "fintegrated-as">, Flags<[DriverOption]>,
 def fno_integrated_as : Flag<["-"], "fno-integrated-as">,
                         Flags<[CC1Option, DriverOption]>, Group<f_Group>,
                         HelpText<"Disable the integrated assembler">;
+
+def fcheckedc_runtime_checks : Flag<["-"], "fcheckedc-runtime-checks">,
+  Group<f_Group>, Flags<[CC1Option]>;
+def fno_checkedc_runtime_checks : Flag<["-"], "fno-checkedc-runtime-checks">,
+  Group<f_Group>, Flags<[CC1Option]>;
+
 def : Flag<["-"], "integrated-as">, Alias<fintegrated_as>, Flags<[DriverOption]>;
 def : Flag<["-"], "no-integrated-as">, Alias<fno_integrated_as>,
       Flags<[CC1Option, DriverOption]>;
--- a/include/clang/Driver/Options.td
+++ b/include/clang/Driver/Options.td
@@ -2624,10 +2624,10 @@ def fno_integrated_as : Flag<["-"], "fno-integrated-as">,
 
 def fcheckedc_null_ptr_checks : Flag<["-"], "fcheckedc-null-ptr-checks">,
   Group<f_Group>, Flags<[CC1Option]>,
-  HelpText<"Enable runtime null ptr checks">;
+  HelpText<"Enable runtime null pointer checks">;
 def fno_checkedc_null_ptr_checks : Flag<["-"], "fno-checkedc-null-ptr-checks">,
   Group<f_Group>, Flags<[CC1Option]>,
-  HelpText<"Disable runtime null ptr checks">;
+  HelpText<"Disable runtime null pointer checks">;
 
 def : Flag<["-"], "integrated-as">, Alias<fintegrated_as>, Flags<[DriverOption]>;
 def : Flag<["-"], "no-integrated-as">, Alias<fno_integrated_as>,
--- a/include/clang/Driver/Options.td
+++ b/include/clang/Driver/Options.td
@@ -2622,10 +2622,12 @@ def fno_integrated_as : Flag<["-"], "fno-integrated-as">,
                         Flags<[CC1Option, DriverOption]>, Group<f_Group>,
                         HelpText<"Disable the integrated assembler">;
 
-def fcheckedc_runtime_checks : Flag<["-"], "fcheckedc-runtime-checks">,
-  Group<f_Group>, Flags<[CC1Option]>;
-def fno_checkedc_runtime_checks : Flag<["-"], "fno-checkedc-runtime-checks">,
-  Group<f_Group>, Flags<[CC1Option]>;
+def fcheckedc_null_ptr_checks : Flag<["-"], "fcheckedc-null-ptr-checks">,
+  Group<f_Group>, Flags<[CC1Option]>,
+  HelpText<"Enable runtime null ptr checks">;
+def fno_checkedc_null_ptr_checks : Flag<["-"], "fno-checkedc-null-ptr-checks">,
+  Group<f_Group>, Flags<[CC1Option]>,
+  HelpText<"Disable runtime null ptr checks">;
 
 def : Flag<["-"], "integrated-as">, Alias<fintegrated_as>, Flags<[DriverOption]>;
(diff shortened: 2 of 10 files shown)


mgrang:
`int main() {
  int *i = NULL;
  ++i; // No dynamic check added here.

  array_ptr<char> p : count(3) = NULL;
  ++p; // Dynamic check added here.
}
`

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -2342,8 +2342,17 @@ llvm::Value *ScalarExprEmitter::EmitIncDecConsiderOverflowBehavior(
...
 ScalarExprEmitter::EmitScalarPrePostIncDec(const UnaryOperator *E, LValue LV,
                                            bool isInc, bool isPre) {
-
   QualType type = E->getSubExpr()->getType();
+
+  // Add a dynamic check for pointer arithmetic of null pointers.
+  if (type->isCheckedPointerType() || type->isCheckedArrayType()) {
+    LValueBaseInfo BaseInfo;
+    TBAAAccessInfo TBAAInfo;
+    Address Addr = CGF.EmitPointerWithAlignment(E->getSubExpr(),
```
dtarditi: This will result in E's subexpression being evaluated twice.  This would be a problem for something where the subexpression might have a side-effect, such as `(a[f()])++`.   I think you need to push the insertion of the check into the cases below.   The only interesting cases are those where a pointer type can occur.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -2427,6 +2427,15 @@ ScalarExprEmitter::EmitScalarPrePostIncDec(const UnaryOperator *E, LValue LV,
 
   // Next most common: pointer increment.
   } else if (const PointerType *ptr = type->getAs<PointerType>()) {
```
mgrang: According to the spec, array_ptrs of function pointers are not allowed. So I guess we do not need a check in the case for function pointers.

mgrang:
This patch currently handles the case of pre/post inc/dec of array_ptrs. I am now working to extend this patch to handle binary ops as well. For example:

```
array_ptr<char> p = NULL;
p += 1;
p = p - 1;
```

mgrang:
Should we guard the insertion of non-null checks via a clang flag? The user can pass a flag like `-fno-runtime-checks` to skip insertion of run time checks for performance-critical code.

dtarditi:
Yes, I think a clang flag controlling isertion of these checks is a good idea.  We'll need it so that we can examine the performance impact of adding the checks.

lib/CodeGen/CGExprScalar.cpp
dtarditi: Agreed - we do not need a check for the case of function pointers.

mgrang:
> Yes, I think a clang flag controlling isertion of these checks is a good idea. We'll need it so that we can examine the performance impact of adding the checks.

Added in latest patchset.

lib/CodeGen/CGDynamicCheck.cpp
```diff
@@ -54,6 +54,9 @@ void CodeGenFunction::EmitDynamicNonNullCheck(const Address BaseAddr, const Qual
   if (!getLangOpts().CheckedC)
     return;
 
+  if (!CGM.getCodeGenOpts().CheckedCRuntimeChecks)
```
mgrang: Runtime checks for non-null will only be added if the flag `-fcheckedc-runtime-checks` is present. Since I have added this check here, it means this would affect all non-null checks (and not just those for pointer arithmetic). This is resulting in 3 unit test failures which we need to fix (by adding this flag to those tests). Is this behavior acceptable?

lenary:
Nice!

lib/CodeGen/CGDynamicCheck.cpp
dtarditi: It should be on by default, and something that we can disable.

dtarditi:
I think you should rename the flag to indicate that it is specific to null checks.  With the current name, someone might think we disable bounds checks too.

include/clang/Basic/CodeGenOptions.def
```diff
@@ -364,6 +364,9 @@ CODEGENOPT(BranchTargetEnforcement, 1, 0)
 /// Whether to emit unused static constants.
 CODEGENOPT(KeepStaticConsts, 1, 0)
 
+/// Whether to add null ptr checks for checkedc.
```
sunnychatterjee: Do we want the nullptr checks to be emitted by default?

include/clang/Driver/Options.td
```diff
@@ -2621,6 +2621,14 @@ def fintegrated_as : Flag<["-"], "fintegrated-as">, Flags<[DriverOption]>,
 def fno_integrated_as : Flag<["-"], "fno-integrated-as">,
                         Flags<[CC1Option, DriverOption]>, Group<f_Group>,
                         HelpText<"Disable the integrated assembler">;
+
+def fcheckedc_null_ptr_checks : Flag<["-"], "fcheckedc-null-ptr-checks">,
+  Group<f_Group>, Flags<[CC1Option]>,
+  HelpText<"Enable runtime null ptr checks">;
```
sunnychatterjee: Nit: "Enable runtime nullptr checks". Similarly, at other places.

include/clang/Basic/CodeGenOptions.def
mgrang: Yes, we want to emit them by default. Please see @dtarditi 's comment above.

include/clang/Driver/Options.td
mgrang: Thanks. I wanted to distinguish general "null pointers" from the C++ "nulllptr" keyword. That's the reason for the space in between. I guess I can spell it out as "null pointer".

mgrang:
This patch is causing 3 run time unit tests to assert. I am working on fixing it.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -2319,14 +2319,100 @@ static BinOpInfo createBinOpInfoFromIncDec(const UnaryOperator *E,
...
+    //   | `-DeclRefExpr '_Array_ptr<int> (int) : count(arg #0)' Function 'bar' '_Array_ptr<int> (int) : count(arg #0)'
+    //   `-IntegerLiteral 'int' 1
+
+    const auto *V = CGF.getBoundsTemporaryRValueMapping(TE).getScalarVal();
+    const auto Addr =
+      CGF.CreateDefaultAlignTempAlloca(V->getType(), "saved-rvalue");
+    CGF.EmitDynamicNonNullCheck(Addr, Ty);
+    return;
+
+  } else if (RV.isComplex()) {
```
mgrang: I also do not have any driving examples for complex and aggregate rvalues but I have added code here just in case. As an alternative, we could simply assert here and only handle scalar rvalues.

mgrang:
> This patch is causing 3 run time unit tests to assert. I am working on fixing it.

Fixed the 3 unit tests in the latest patchset.

mgrang:
I ran LNT testsuite locally and do not see any issues due to this patch. Although there are 10 tests failing with errors like `expected identifier or '('`. These fail even w/o this patch.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -2319,6 +2319,106 @@ static BinOpInfo createBinOpInfoFromIncDec(const UnaryOperator *E,
   return BinOp;
 }
 
+static Expr *peelOffPointerArithmetic(const BinaryOperator *B) {
+  if (B->isAdditiveOp() && B->getType()->isPointerType()) {
+    if (B->getLHS()->getType()->isPointerType()) {
+      return B->getLHS();
+    } else if (B->getRHS()->getType()->isPointerType()) {
```
sunnychatterjee: Nit: Coding style. In some of the other places like CGDynamicCheck.cpp, I'm seeing that for single statements under an if-else, we aren't using braces.  Should we drop the braces from the single statements under if-else here?

For example:

if (!(BaseTy->isCheckedPointerType() || BaseTy->isCheckedArrayType()))
    return;

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -2319,6 +2319,106 @@ static BinOpInfo createBinOpInfoFromIncDec(const UnaryOperator *E,
   return BinOp;
 }
 
+static Expr *peelOffPointerArithmetic(const BinaryOperator *B) {
```
sunnychatterjee: Nit: I guess this function is getting the pointer type for expressions like (a + i) or (i + a), where 'a' is the pointer type and 'i' is the integer offset. Can we add a comment describing this?

Also, maybe we can rename the function to be more clear on the intent? Something like: getPointerTypeFromPointerOffsetExpression or something similar.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -2319,6 +2319,106 @@ static BinOpInfo createBinOpInfoFromIncDec(const UnaryOperator *E,
...
+    if (B->getLHS()->getType()->isPointerType()) {
+      return B->getLHS();
+    } else if (B->getRHS()->getType()->isPointerType()) {
+      return B->getRHS();
+    }
+  }
+  return nullptr;
+}
+
+static Expr *peelOffOuterExpr(Expr *E) {
```
sunnychatterjee: 1. Can we rename this function to: skipCasts or something similar, which expresses the intent more clearly.
2. What about parenthesis? Do we create expressions for them in the ASTs as well? If so, we'd want to skip them as well.
3. The check for: if (const auto *BO = dyn_cast<BinaryOperator>(E)) can be more targeted, since you only handle additive operator. So, you can use: isAdditiveOp() here.

lib/CodeGen/CGExprScalar.cpp
mgrang: `peelOffPointerArithmetic` has been adapted from a similar function in `lib/StaticAnalyzer/Core/BugReporterVisitors.cpp`. I wanted to keep the same name to make the association clear and also to help while grep'ping for the function.

lib/CodeGen/CGExprScalar.cpp
mgrang: 1. `peelOffOuterExpr` has been adapted from a similar function in `lib/StaticAnalyzer/Core/BugReporterVisitors.cpp`. I wanted to keep the same name to make the association clear and also to help while grep'ping for the function.

2. Parenthesis are represented in the AST as node levels. For example: `(a * b) + c` is represented is:
```
                    +
                  /   \
                 *     c
               /   \
              a     b
```

3. `isAdditiveOp` is already checked inside `peelOffPointerArithmetic`.

lib/CodeGen/CGExprScalar.cpp
mgrang: Added comment describing the function.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -3162,6 +3267,9 @@ static Value *emitPointerArithmetic(CodeGenFunction &CGF,
     std::swap(pointerOperand, indexOperand);
   }
 
+  // Insert a dynamic check for arithmetic on null checked pointers.
+  emitDynamicNonNullCheck(CGF, pointerOperand);
```
dtarditi: This has similar problems, in that pointerOperand will be re-evaluated.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -2319,6 +2319,108 @@ static BinOpInfo createBinOpInfoFromIncDec(const UnaryOperator *E,
...
+    if (Expr *SubE = peelOffPointerArithmetic(BO))
+      return peelOffOuterExpr(SubE);
+
+  return E;
+}
+
+// Determine whether to emit a dynamic non-null check.
+static void emitDynamicNonNullCheck(CodeGenFunction &CGF, Expr *E) {
+  if (!CGF.CGM.getCodeGenOpts().CheckedCNullPtrChecks)
+    return;
```
dtarditi: I would structure this code differently.   It's an optimization to not emit a dynamic null check if E can be proved at compile-time to be non-null.  So I'd try to prove that E does not need a null check and return if the null check can be skipped.  Then I would go into the general case. of adding a check.

Please improve the comments on this method.  We should always explain what non-obvious arguments are.  For example, in this case E is the pointer-typed expression in a pointer-arithmetic expression.  We need to check the value of E is non-null, if E is a checked pointer type.

lib/CodeGen/CGExprScalar.cpp
```diff
@@ -2427,6 +2433,9 @@ ScalarExprEmitter::EmitScalarPrePostIncDec(const UnaryOperator *E, LValue LV,
 
   // Next most common: pointer increment.
   } else if (const PointerType *ptr = type->getAs<PointerType>()) {
+    // Insert a dynamic check for arithmetic on null checked pointers.
```
mgrang: I am not sure about adding a compile time check to see if `value` is indeed null so that the runtime check can be elided. LLVM has a function called `isKnownNonNull` which is used in a couple of places in Clang but in my case it always returns 0.

Consider this code:
```
array_ptr<char> s = NULL;
++s;
```

This is the IR:
```
 %s = alloca i8*, align 8
 store i8* null, i8** %s, align 8
 %0 = load i8*, i8** %s, align 8
 %_Dynamic_check.non_null = icmp ne i8* %0, null
```

I guess we need to check what is getting stored in `%s`. If it is indeed non-null, then we can elide the check. Is this a valid approach?

lib/CodeGen/CGExprScalar.cpp
sunnychatterjee: A suggestion: If you don't have motivating examples but have added the code for completeness, you can refer in comments to other parts of the codebase where such cases are being handled.

mgrang: This is no longer present in the latest patch.

mgrang:
This patch had merge conflicts with master. While trying to resolve the conflicts, git added other commits to this PR. At this point, I would simply create a new PR with the conflicts fixed. Please see https://github.com/microsoft/checkedc-clang/pull/663/files.

dtarditi:
Closing this PR because it is superseded by PR #663.

PR #651, opened 2019-08-01
Conform CGDynamicCheck.cpp to 80 char line limit [NFC]

mgrang:


mgrang:
These are only formatting changes. So I will go ahead and merge this patch.

PR #657, opened 2019-08-07
Forward dataflow analysis for collecting comparison facts

ppashakhanloo:
This pull request adds a forward dataflow analysis for collecting facts in every function for each CFG block.
I have added some new tests under test/CheckedC/dump_dataflow_facts.c.
~~Some comments still need to be added.~~

The main function that runs the analysis is AvailableFactsAnalysis::Analyze() which implements the iterative worklist algorithm. This function must be called before checking the bounds declarations.

For each CFG block _B_, the following sets are computed:
- Gen[B]: For every _if-node_, if the condition is a binary operator where opcode is one of <, <=, >, or >=, add the condition to Gen[B]
- Kill[B]: First, collect all the variables that are defined in block _B_, and if any expression in the program contains any of these variables, add it to Kill[B]
- In[B]: The intersection of the Out set of Predecessors of B
- Out[B]: First, compute the set union between In[B] and Gen[B]. Next, remove any comparison in Kill[B].

For using the collected facts, the following functions can be used:
- void Next(); // moving to the next block in CFG
- void GetFacts(std::set<std::pair<Expr *, Expr *>>& InequalityFacts); // getting the collected facts of the current CFG block

Representation of the comparison facts:
- Each fact is represented by a pair of two Expr*s. By (Expr1, Expr2), we mean Expr1 <= Expr2 or Expr1 < Expr2.


ppashakhanloo:
@dtarditi I believe this PR is ready for a review of this stage.

lib/Sema/SemaBounds.cpp
```diff
@@ -3285,6 +3288,318 @@ namespace {
...
+  void Reset() {
+    CurrentIndex = 0;
+  }
+
+  void Next() {
+    CurrentIndex++;
+  }
+
+  // This function fills `ComparisonFacts` with pairs (Expr1, Expr2) where
+  // Expr1 < Expr2, Expr1 <= Expr2, Expr2 > Expr1, or Expr2 >= Expr1.
```
dtarditi: I think this comment is confusing.    The invariant is that Expr1 <= Expr2.

lib/Sema/SemaBounds.cpp
```diff
@@ -3285,6 +3288,318 @@ namespace {
...
+  // Given two sets S1 and S2, this function returns true if the two sets
+  // are not equal.
+  // Equality means that their sizes are the same, and every member of S1 is found in
+  // S2 and vice versa.
+  bool Differ(ComparisonSet& S1, ComparisonSet& S2) {
+    if (S1.size() != S2.size())
+      return true;
+    if (S1.size() == 0 && S2.size() == 0)
+      return false;
+    if (S1.size() > S2.size())
```
dtarditi: This logic seems wrong.  By the first check in the function, S1.size() must equal S2.size().

lib/Sema/SemaBounds.cpp
```diff
@@ -3285,6 +3288,318 @@ namespace {
...
+        case BinaryOperatorKind::BO_GE:
+        case BinaryOperatorKind::BO_GT:
+          ISet.insert(Comparison(BO->getRHS(), BO->getLHS()));
+          break;
+        default:
+          break;
+      }
+    } else
+      return;
+    for (auto I = St->child_begin(); I != St->child_end(); ++I)
```
dtarditi: This logic is wrong - you shouldn't recurse.     There's no guarantee that comparison produces a true or false value consumed by the if-statement.  For example, consider `if (f(a < b))`.

lib/Sema/SemaBounds.cpp
```diff
@@ -3285,6 +3288,318 @@ namespace {
...
+      ExtractComparisons(*I, ISet);
+  }
+
+  // Given a set of Comparisons `InputSet`, this function negates them
+  // by swaping the element of pairs, and records them in `OutputSet`.
```
dtarditi: Typo: swaping->swapping.

test/CheckedC/dump-dataflow-facts.c
```diff
@@ -0,0 +1,213 @@
+// Tests for dumping of datafow analysis for collecting facts
+//
+// RUN: %clang_cc1 -DDEBUG_DATAFLOW -fcheckedc-extension %s 2>&1 | FileCheck %s
```
dtarditi: You need to add a compiler flag.    After the analysis, you'd check the compiler flag and dump the results accordingly.  You can look at the recent commit by @mgrang to see how to do this.

test/CheckedC/dump-dataflow-facts.c
```diff
@@ -0,0 +1,213 @@
...
+int f(int a);
+
+void fn_1(void) {
+  int a, b, c;
+  if (b < c)
+    if (f((a=5)+3))
+      b = c;
+}
+
+// CHECK: {
```
dtarditi: This should include block number information.  Otherwise this will be very difficult to debug in the future.

Could you add more test cases for more complicated expressions?

lib/Sema/SemaBounds.cpp
```diff
@@ -3285,6 +3288,318 @@ namespace {
...
+      CurrentBlock->OutThen = Difference(Union(CurrentBlock->In, CurrentBlock->GenThen), CurrentBlock->Kill);
+      CurrentBlock->OutElse = Difference(Union(CurrentBlock->In, CurrentBlock->GenElse), CurrentBlock->Kill);
+
+      // Recompute the Affected Blocks
+      if (Differ(OldOutThen, CurrentBlock->OutThen) || Differ(OldOutElse, CurrentBlock->OutElse))
+        for (CFGBlock::const_succ_iterator I = CurrentBlock->Block->succ_begin(), E = CurrentBlock->Block->succ_end(); I != E; ++I)
+          WorkList.push(GetByCFGBlock(*I));
+    }
+
+  //if(DEBUG_DATAFLOW)
```
dtarditi: This needs to be moved out to a separate routine.  The dumps should be controlled by a compiler flag.

lib/Sema/SemaBounds.cpp
```diff
@@ -3285,6 +3288,318 @@ namespace {
...
+    CollectExpressions(I.second, Exprs);
+    for (auto InnerExpr : Exprs)
+      if (const DeclRefExpr *DRE = dyn_cast<DeclRefExpr>(InnerExpr))
+        if (const VarDecl *VD = dyn_cast<VarDecl>(DRE->getDecl()))
+          if (VD == V)
+            return true;
+    return false;
+  }
+
+  // This function looks recursively for comparisons in statement `S`.
```
dtarditi: I think the proper invariant is this: this function computes a list of comparisons E1 <= E2 that will be true if the expression S evaluates to true.   Clearly if S is simple direct comparison expression E3 op E4, a comparison can be created if op is one of LE, LT, GE, GT, or EQ.      If S has the form A && B, comparisons can be created for A and B.

Could you add a TODO for handling logical negation the (`!`) operator?

dtarditi:
We also need to conservative in the set of comparisons that we collect with respect to pointers and calls.   
- We can't include comparisons whose expressions involve calls.  The call could produce a different value.
- Similarly a comparison shouldn't include a reference to a volatile variable.   
- If an expression in a comparison contains a pointer dereference, we need to kill the comparison at any potential pointer assignment expression (assignment via a pointer or a call).

lib/Sema/SemaBounds.cpp
```diff
@@ -3285,6 +3288,318 @@ namespace {
...
+  // Then, it records them as elements of type Comparison in `ISet`.
+  // Each Comparison is a pair in form of (Expr1, Expr2) where both
+  // Expr1 and Expr2 are of type Expr*.
+  // Each pair corresponds to a comparison expression of one of the following forms:
+  // - Expr1 < Expr2
+  // - Expr1 <= Expr2
+  // - Expr2 > Expr1
+  // - Expr2 >= Expr1
+  void ExtractComparisons(const Stmt *St, ComparisonSet &ISet) {
```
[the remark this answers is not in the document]
ppashakhanloo: Sure.

include/clang/Driver/Options.td
```diff
@@ -2625,6 +2625,9 @@ def : Flag<["-"], "integrated-as">, Alias<fintegrated_as>, Flags<[DriverOption]>
 def : Flag<["-"], "no-integrated-as">, Alias<fno_integrated_as>,
       Flags<[CC1Option, DriverOption]>;
 
+def fdump_extracted_comparison_facts : Flag<["-"], "fdump_extracted_comparison_facts">, Group<f_Group>, Flags<[CC1Option]>,
```
mgrang: The actual flag should be dash (-) separated, not underscore separated.
`"fdump_extracted_comparison_facts" ==> "fdump-extracted-comparison-facts"`

include/clang/Sema/AvailableFactsAnalysis.h
```diff
@@ -22,6 +22,7 @@
 #include "clang/Analysis/Analyses/PostOrderCFGView.h"
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "llvm/ADT/SmallPtrSet.h"
+#include "clang/Sema/Sema.h"
```
mgrang: Sort includes alphabetically.

lib/Sema/SemaBounds.cpp
ppashakhanloo: Done.

lib/Sema/SemaBounds.cpp
ppashakhanloo: Done.

test/CheckedC/dump-dataflow-facts.c
ppashakhanloo: Done. The new flag is -fdump_extracted_comparison_facts.

lib/Sema/SemaBounds.cpp
```diff
@@ -3378,6 +3381,335 @@ namespace {
...
+  typedef std::set<Comparison> ComparisonSet;
+
+  class ElevatedCFGBlock {
+  private:
+    const CFGBlock *Block;
+    ComparisonSet In, OutThen, OutElse;
+    ComparisonSet Kill, GenThen, GenElse;
+
+  public:
+    ElevatedCFGBlock(const CFGBlock *Block) : Block(Block) {}
```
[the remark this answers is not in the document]
ppashakhanloo: @sunnychatterjee I followed a similar naming convention that appears in the existing constructors in this file. Should I change it?

lib/Sema/SemaBounds.cpp
```diff
@@ -3378,6 +3381,335 @@ namespace {
...
+namespace {
+class AvailableFactsAnalysis {
+private:
+  typedef llvm::SmallPtrSet<const Expr *, 16> ExprSet;
+  typedef std::pair<Expr *, Expr *> Comparison;
+  typedef std::set<Comparison> ComparisonSet;
+
+  class ElevatedCFGBlock {
+  private:
+    const CFGBlock *Block;
```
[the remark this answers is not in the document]
ppashakhanloo: @sunnychatterjee ElevatedCFGBlock has one constructor with `const CFGBlock *Block` as input. And there is no default constructor for it. I am not sure why we need to add default initializations for the data members here.

include/clang/Sema/AvailableFactsAnalysis.h
```diff
@@ -0,0 +1,82 @@
+//===-------- AvailableFactsAnalysis.h - collect comparison facts --------===//
```
mgrang: Align the ends of lines like these.

include/clang/Sema/AvailableFactsAnalysis.h
mgrang: > Sort includes alphabetically.

Uppercase includes come before lowercase ones.

include/clang/Sema/AvailableFactsAnalysis.h
```diff
@@ -0,0 +1,82 @@
...
+
+#include "clang/Analysis/CFG.h"
+#include "clang/Analysis/Analyses/PostOrderCFGView.h"
+#include "clang/AST/RecursiveASTVisitor.h"
+#include "llvm/ADT/SmallPtrSet.h"
+#include "clang/Sema/Sema.h"
+#include <queue>
+
+namespace clang {
+  typedef std::set<const Expr *> ExprSet;
```
mgrang: C++ convention is to use `using` instead of `typedef`.

include/clang/Sema/AvailableFactsAnalysis.h
```diff
@@ -0,0 +1,82 @@
...
+  typedef std::pair<Expr *, Expr *> Comparison;
```
mgrang: You should use llvm containers like SmallPtrSet, SmallSet, etc instead of std ones.

lib/Sema/SemaBounds.cpp
```diff
@@ -3378,6 +3381,335 @@ namespace {
...
+  void GetFacts(std::set<std::pair<Expr *, Expr *>>& ComparisonFacts) {
+    ComparisonFacts = Blocks[CurrentIndex]->In;
+  }
+
+private:
+  ElevatedCFGBlock* GetByCFGBlock(const CFGBlock *B) {
+    return Blocks[BlockIDs[B->getBlockID()]];
+  }
+
+  // Given two sets S1 and S2, the return value is S1 \ S2.
```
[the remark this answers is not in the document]
ppashakhanloo: @sunnychatterjee I will change this comment to `S1 - S2` for more clarity. I meant `\` which is the set difference operator.

lib/Sema/SemaBounds.cpp
```diff
@@ -33,11 +33,13 @@
 #include "clang/Analysis/CFG.h"
 #include "clang/Analysis/Analyses/PostOrderCFGView.h"
 #include "clang/AST/CanonBounds.h"
+#include "clang/Sema/AvailableFactsAnalysis.h"
```
mgrang: Sort includes.

test/CheckedC/dump-dataflow-facts.c
```diff
@@ -0,0 +1,256 @@
+// Tests for dumping of datafow analysis for collecting facts
+//
+// RUN: %clang_cc1 -fcheckedc-extension -fdump_extracted_comparison_facts %s 2>1 | FileCheck %s
```
mgrang: -fcheckedc-extension needs to be removed from the entire codebase as it does not do anything. Please remove it from here.

test/CheckedC/dump-dataflow-facts.c
```diff
@@ -0,0 +1,256 @@
...
+
+void fn_1(void) {
+  int a, b, c;
+  if (b < c)
+    if (f((a=5)+3))
+      b = c;
+}
+
+// CHECK: Block #4: {
+// CHECK: In:
```
mgrang: Can you use CHECK-NEXT instead of CHECK here? Also inside a new function it's good to check for the function name, like:

`CHECK-LABEL: fn_1`

test/CheckedC/dump-dataflow-facts.c
```diff
@@ -0,0 +1,256 @@
...
+// CHECK: Block #1: {
+// CHECK: In:
+// CHECK: Kill:
+// CHECK: }
+// CHECK: Block #0: {
+// CHECK: In:
+// CHECK: Kill:
+// CHECK: }
+
+
```
mgrang: Remove extra newline.

lib/Sema/SemaBounds.cpp
```diff
@@ -3378,6 +3381,335 @@ namespace {
...
+      }
+  }
+
+  void CollectExpressions(const Stmt *St, ExprSet &AllExprs) {
+    if (!St)
+      return;
+    if (const Expr *E = dyn_cast<Expr>(St))
+      AllExprs.insert(E);
+    for (auto I = St->child_begin(); I != St->child_end(); ++I)
+      CollectExpressions(*I, AllExprs);
```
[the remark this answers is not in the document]
ppashakhanloo: Done!

lib/Sema/SemaBounds.cpp
```diff
@@ -3378,6 +3381,335 @@ namespace {
...
+            }
+          }
+        }
+      }
+
+    for (auto I = St->child_begin(); I != St->child_end(); ++I)
+      CollectDefinedVars(*I, DefinedVars);
+  }
+
+//#if DEBUG_DATAFLOW
```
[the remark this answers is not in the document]
ppashakhanloo: Yes, I removed this comment and added a new compiler flag which is only used when testing the functionality of this part of the code.

lib/Sema/SemaBounds.cpp
```diff
@@ -3394,8 +3726,11 @@ void Sema::CheckFunctionBodyBoundsDecls(FunctionDecl *FD, Stmt *Body) {
   ComputeBoundsDependencies(Tracker, FD, Body);
   std::unique_ptr<CFG> Cfg = CFG::buildCFG(nullptr, Body, &getASTContext(), CFG::BuildOptions());
   CheckBoundsDeclarations Checker(*this, Body, Cfg.get(), FD->getBoundsExpr());
-  if (Cfg != nullptr)
+  AvailableFactsAnalysis Collector(*this, Cfg.get());
```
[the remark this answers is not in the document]
ppashakhanloo: Done.

include/clang/Driver/Options.td
ppashakhanloo: Done.

test/CheckedC/dump-dataflow-facts.c
ppashakhanloo: Done

test/CheckedC/dump-dataflow-facts.c
ppashakhanloo: @mgrang Why should I check for the function name?

test/CheckedC/dump-dataflow-facts.c
ppashakhanloo: Done.

lib/Sema/SemaBounds.cpp
```diff
@@ -3378,6 +3381,335 @@ namespace {
...
+      CurrentBlock->OutThen = Difference(Union(CurrentBlock->In, CurrentBlock->GenThen), CurrentBlock->Kill);
+      CurrentBlock->OutElse = Difference(Union(CurrentBlock->In, CurrentBlock->GenElse), CurrentBlock->Kill);
+
+      // Recompute the Affected Blocks
+      if (Differ(OldOutThen, CurrentBlock->OutThen) || Differ(OldOutElse, CurrentBlock->OutElse))
+        for (CFGBlock::const_succ_iterator I = CurrentBlock->Block->succ_begin(), E = CurrentBlock->Block->succ_end(); I != E; ++I)
+          WorkList.push(GetByCFGBlock(*I));
+    }
+
+  //if(DEBUG_DATAFLOW)
```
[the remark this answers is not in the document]
ppashakhanloo: Yes, I added a compiler flag for debugging purposes and removed this comment.

test/CheckedC/dump-dataflow-facts.c
mgrang: > @mgrang Why should I check for the function name?

To make sure you are checking output from the correct function. Consider this:

```
// CHECK: Block #1 
// CHECK: In

void foo() {}
void bar() {}
```

Here it is not clear whose output you are checking. The output of foo may conflict with the output of bar and can confuse lit. So it's good to always check for `CHECK-LABEL` first. See [https://llvm.org/docs/CommandGuide/FileCheck.html#the-check-label-directive](url)

lib/Sema/SemaBounds.cpp
```diff
@@ -3285,6 +3288,318 @@ namespace {
   };
 }
 
+namespace {
```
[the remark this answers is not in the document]
ppashakhanloo: Done

lib/Sema/SemaBounds.cpp
```diff
@@ -3378,6 +3381,335 @@ namespace {
...
+  class ElevatedCFGBlock {
+  private:
+    const CFGBlock *Block;
+    ComparisonSet In, OutThen, OutElse;
+    ComparisonSet Kill, GenThen, GenElse;
+
+  public:
+    ElevatedCFGBlock(const CFGBlock *Block) : Block(Block) {}
+
+    friend class AvailableFactsAnalysis;
```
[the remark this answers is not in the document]
ppashakhanloo: @sunnychatterjee I used `friend` since `AvailableFactsAnalysis` should be the only class that manipulates an object of `ElevatedCFGBlock`. (I thought of it as an Iterator class.)
If I make public methods to access the data members of it, I think I share more than necessary with the world outside `AvailableFactsAnalysis`. Is there a better way to achieve this goal?

include/clang/Sema/AvailableFactsAnalysis.h
ppashakhanloo: Fixed.

lib/Sema/SemaBounds.cpp
```diff
@@ -3378,6 +3381,335 @@ namespace {
...
+
+    PostOrderCFGView POView = PostOrderCFGView(Cfg);
+    unsigned int MaxBlockID = 0;
+    for (const CFGBlock *Block : POView)
+      if (Block->getBlockID() > MaxBlockID)
+        MaxBlockID = Block->getBlockID();
+
+    BlockIDs.resize(MaxBlockID + 1, 0);
+    for (const CFGBlock *Block : POView) {
+      auto NewBlock = new ElevatedCFGBlock(Block);
```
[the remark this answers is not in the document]
ppashakhanloo: Dataflow information for every block is stored in an ElevatedCFGBlock. We need them even after the analysis is done, i.e. it goes out of this scope. That's why I used `new`.
By the way, thanks for pointing this out. I had totally forgotten to `delete` them.

lib/Sema/SemaBounds.cpp
```diff
@@ -3378,6 +3381,335 @@ namespace {
...
+    assert(Cfg && "expected CFG to exist");
+    ComparisonSet AllComparisons;
+
+    PostOrderCFGView POView = PostOrderCFGView(Cfg);
+    unsigned int MaxBlockID = 0;
+    for (const CFGBlock *Block : POView)
+      if (Block->getBlockID() > MaxBlockID)
+        MaxBlockID = Block->getBlockID();
+
+    BlockIDs.resize(MaxBlockID + 1, 0);
```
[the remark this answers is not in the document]
ppashakhanloo: Fixed.

lib/Sema/SemaBounds.cpp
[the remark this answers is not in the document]
sunnychatterjee: @ppashakhanloo You don't have to change this if it appears in the existing convention. Personally, I don't prefer this style since you have to qualify member access with a "this" in such cases.

lib/Sema/SemaBounds.cpp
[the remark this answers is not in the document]
sunnychatterjee: @ppashakhanloo For a single data member like in your example, it's fine. I usually follow the rule to always initialize data members of the object: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-always.
This way, we avoid errors when constructors sometimes don't initialize all the data members of a class.

lib/Sema/SemaBounds.cpp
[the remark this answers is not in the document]
sunnychatterjee: @ppashakhanloo This was a clarifying question. Feel free keep the set difference operator.

ppashakhanloo:
@dtarditi I have addressed your comments. The code is ready for review.

test/CheckedC/dump-dataflow-facts.c
ppashakhanloo: @mgrang Do I need to dump the function name as well to be able to use CHECK-LABEL?
I followed the doc on this directive but it cannot match the function name inside the function.

lib/Frontend/CompilerInvocation.cpp
```diff
@@ -2568,6 +2568,9 @@ static void ParseLangArgs(LangOptions &Opts, ArgList &Args, InputKind IK,
   if (Args.hasArg(OPT_fdump_inferred_bounds))
     Opts.DumpInferredBounds = true;
 
+  if (Args.hasArg(OPT_fdump_extracted_comparison_facts))
+    Opts.DumpExtractedComparisonFacts = true;
```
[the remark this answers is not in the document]
ppashakhanloo: `DumpExtractedComparisonFacts` is used in clang/lib/Sema/SeamBounds.cpp Line 3402.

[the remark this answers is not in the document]
mgrang: Is it part of another patch?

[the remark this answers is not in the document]
ppashakhanloo: No, I use this flag for dumping the information that is collected by dataflow analysis.
`SeamBounds.cpp` above was a type. I edited the comment with the correct filename.

test/CheckedC/dump-dataflow-facts.c
mgrang: I see you are not dumping the IR. Then you do not need CHECK-LABEL, I guess. However, would it be more useful to the user to also dump the function name as part of your new flag?

lib/Frontend/CompilerInvocation.cpp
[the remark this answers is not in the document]
mgrang: I see you have declared `DumpExtractedComparisonFacts` as `LangOpts` as well as `CodeGenOpts` but you are really only using it from `LangOpts`. I guess you could remove it from `CodeGenOpts`?

test/CheckedC/dump-dataflow-facts.c
ppashakhanloo: I think it makes the future debugging easier if the function name is dumped.

mgrang: I think so too. So once you add it to your flag you can add back the CHECK-LABELs.

ppashakhanloo: Good idea, thanks!

lib/Frontend/CompilerInvocation.cpp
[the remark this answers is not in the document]
ppashakhanloo: Oh, I see. I will remove it from `CodeGenOpts` then.

lib/Sema/SemaBounds.cpp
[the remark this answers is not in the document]
sunnychatterjee: @ppashakhanloo Thanks for the explanation. I try to avoid "friend" wherever possible since it tends to break encapsulation. However, there are exceptions to this like iterators as you mention. Given that the two classes are so close to each other (nested), let's leave them as is.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,458 @@
...
+          return true;
+  return false;
+}
+
+// This function returns true only if an expression in Comparison `I` contains
+// a pointer deref. A pointer deref can appear as one of the following form:
+// 1. With Deref operator `*`
+// 2. `->` for accessing a member
+// 3. Array subscript `[]`
+bool AvailableFactsAnalysis::ContainsPointerDeref(Comparison& I) {
```
dtarditi: This isn't the right test for whether an expression is or contains a pointer dereference.  The expression should be a cast expression whose operator is an LValueToRValue cast and whose child expression produces an lvalue from a ponter.  For the child expression, the function f(e) must be true, where f(e) is defined as:

1. If e is one of the expressions you list (*, -> or subscript expr), return true.
2.  If e has the form (e1), return f(e1).
3.  If e is an  lvalue cast or no-op cast of e1, return f(e1).  An lvalue is a cast expression whose operator is LValueBitCast.  The no-op cast operator is NoOp (see clang\include\clang\AST\OperationKinds.def).
4.  Otherwise, return false.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,458 @@
...
+       ComparisonContainsDeref.push_back(true);
+     else
+       ComparisonContainsDeref.push_back(false);
+   }
+
+   // Which blocks contain potential pointer assignments? (assignment via a pointer or a call)
+   std::vector<bool> PointerAssignmentInBlocks(Blocks.size(), false);
+   for (unsigned int Index = 0; Index < Blocks.size(); Index++) {
+     std::set<const VarDecl *> DefinedVars;
+     for (CFGElement Elem : *(Blocks[Index]->Block))
```
dtarditi: This isn't the right test for a pointer assignment.   You should be checking for assignments to lvalues produced by a pointer-dereferencing expressions (this is the function f(e) described in my comment for checking for a pointer-dereference).   You should be checking for:

1. An assignment operator whose left-hand side is an expression e where f(e) is true.
2. An increment/decrement operator whose child expression  e is an expression for which f(e) is true.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,458 @@
...
+          DefinedVars.insert(V);
+    }
+  } else if (const UnaryOperator *UO = dyn_cast<const UnaryOperator>(St)) {
+    if (UO->isIncrementDecrementOp()) {
+      Expr *LHS = UO->getSubExpr()->ignoreParenBaseCasts()->IgnoreImpCasts();
+      if (const DeclRefExpr *D = dyn_cast<const DeclRefExpr>(LHS))
+        if (const VarDecl *V = dyn_cast<const VarDecl>(D->getDecl()))
+          DefinedVars.insert(V);
+    }
+  } else if (const DeclRefExpr *DRE = dyn_cast<const DeclRefExpr>(St)) {
```
dtarditi: This case is not necessary: calls don't define their parameters.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,458 @@
...
+// We assume a variable is defined
+// 1. if it appears in the left hand side of an assignment (a = ..., a++, etc.)
+// 2. if it is a pointer which is passed as an argument of a function call
+void AvailableFactsAnalysis::CollectDefinedVars(const Stmt *St, std::set<const VarDecl *> &DefinedVars) {
+  if (!St)
+    return;
+
+  if (const BinaryOperator *BO = dyn_cast<const BinaryOperator>(St)) {
+    if (BO->isAssignmentOp()) {
+      Expr *LHS = BO->getLHS()->ignoreParenBaseCasts()->IgnoreImpCasts();
```
dtarditi: You should only ignore parenthesized expressions, no-op casts, and lvaue bit casts.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,458 @@
...
+      if (const DeclRefExpr *D = dyn_cast<const DeclRefExpr>(LHS))
+        if (const VarDecl *V = dyn_cast<const VarDecl>(D->getDecl()))
+          DefinedVars.insert(V);
+    }
+  } else if (const UnaryOperator *UO = dyn_cast<const UnaryOperator>(St)) {
+    if (UO->isIncrementDecrementOp()) {
+      Expr *LHS = UO->getSubExpr()->ignoreParenBaseCasts()->IgnoreImpCasts();
```
dtarditi: Same here.

ppashakhanloo:
@dtarditi I addressed your comments and added more tests.
Then, I updated CompareUpperOffsets/CompareLowerOffsets to use Facts of the current block that is passed by reference to them. Instead of the function `EqualExtended` that was used before, I added `LessThanOrEqualExtended` which is similar to EqualExtended except that it also looks up facts when comparing two offsets. The corresponding test is in `test/CheckedC/static-checking/bounds-decl-checking.c `.

The code is ready for review.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,499 @@
...
+      if (IsPointerDerefLValue(BO->getLHS()))
+        return true;
+  }
+  if (const UnaryOperator *UO = dyn_cast<UnaryOperator>(E)) {
+    if (UO->isIncrementDecrementOp())
+      if (IsPointerDerefLValue(UO->getSubExpr()))
+        return true;
+  }
+  if (const CallExpr *CE = dyn_cast<CallExpr>(E)) {
+    for (int ArgIndex = 0; ArgIndex < CE->getNumArgs(); ArgIndex++) {
```
dtarditi: Any call expression must be considered as possibly doing a pointer assignment.   We should not consider the types of the arguments passed to the function call.  It's possible that a pointer could be accessible to a function via a global variable or through a structure type.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,499 @@
...
+  if (const ParenExpr *PE = dyn_cast<ParenExpr>(E))
+    return IsPointerDerefLValue(PE->getSubExpr());
+  if (const CastExpr *CE = dyn_cast<CastExpr>(E))
+    if (CE->getCastKind() == CastKind::CK_LValueBitCast ||
+        CE->getCastKind() == CastKind::CK_NoOp)
+      return IsPointerDerefLValue(CE->getSubExpr());
+  return false;
+}
+
+// This function return true if an expression is volatile, and false otherwise.
```
dtarditi: Typo: return->returns

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,499 @@
...
+  return false;
+}
+
+// This function returns true only if expression `E` is a pointer deref.
+// A pointer deref is defined as follows:
+// - `E` should be a cast of type `LValueToRValue, and
+// - If the child of `E` is `e`, then `f(e)` must be true. `f(e)` is defined in
+//   a separate function `IsPointerDerefLValue`.
+// Otherwise, the function returns false;
+bool AvailableFactsAnalysis::IsPointerDeref(const Expr *E) {
```
dtarditi: You need to check whether E contains a pointer dereference.  That means you need to check the children recursively also.  I think you should rename this function.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,499 @@
...
+  if (const CastExpr *CE = dyn_cast<CastExpr>(E)) {
+    if (CE->getCastKind() != CastKind::CK_LValueToRValue)
+      return false;
+    return IsPointerDerefLValue(CE->getSubExpr());
+  }
+  return false;
+}
+
+bool AvailableFactsAnalysis::IsPointerAssignment(const Expr *E) {
```
dtarditi: You need to check whether E contains a pointer assignment.   You need to recursively check the children too.  I think you should rename this function.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,499 @@
...
+  return Result;
+}
+
+// This function returns true if variable `V` is used in the comparison `I`.
+bool AvailableFactsAnalysis::ContainsVariable(Comparison& I, const VarDecl *V) {
+  std::set<const Expr *> Exprs;
+  CollectExpressions(I.first, Exprs);
+  CollectExpressions(I.second, Exprs);
+  for (auto InnerExpr : Exprs)
+    if (const DeclRefExpr *DRE = dyn_cast<DeclRefExpr>(InnerExpr))
```
dtarditi: A variable v produces an lvalue.   For the variable to be used, the lvalue must be the used by an LValueToRValue cast expression.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,499 @@
...
+// - If `E` has the form `A && B`, comparisons can be created for A and B.
+// Note that we do not include comparisons whose expressions involve
+// function calls or references to volatile variables.
+//
+// Some examples:
+// - `E` has the form A < B: add (A, B) to `ISet`.
+// - `E` has the form `E1 && E2`: ExtractComparisons in E1 and E2 and add them to `ISet`.
+// TODO: handle the case where logical negation operator (!) is used.
+void AvailableFactsAnalysis::ExtractComparisons(const Expr *E, ComparisonSet &ISet) {
+  if (IsVolatile(E) || ContainsCallExpr(E))
```
dtarditi: You can use the function Sema::CheckIsNonModifying instead.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,499 @@
...
+// - If `E` has the form `A || B`, negated comparisons can be created for A and B.
+// Note that we do not include comparisons whose expressions involve
+// function calls or references to volatile variables.
+//
+// Some examples:
+// - `E` has the form A < B: add (B, A) to `ISet`.
+// - `E` has the form `E1 || E2`: ExtractNegatedComparisons in E1 and E2 and add
+//   them to `ISet`.
+void AvailableFactsAnalysis::ExtractNegatedComparisons(const Expr *E, ComparisonSet &ISet) {
+  if (IsVolatile(E) || ContainsCallExpr(E))
```
dtarditi: You can use the function Sema::CheckIsNonModifying instead.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,499 @@
...
+  if (const CastExpr *CE = dyn_cast<CastExpr>(E))
+    if (CE->getCastKind() == CastKind::CK_LValueBitCast ||
+        CE->getCastKind() == CastKind::CK_NoOp)
+      return IsPointerDerefLValue(CE->getSubExpr());
+  return false;
+}
+
+// This function return true if an expression is volatile, and false otherwise.
+// An expression is volatile if there is at least one volatile variable in it.
+bool AvailableFactsAnalysis::IsVolatile(const Expr *E) {
```
dtarditi: This is overly conservative: it would flag an expression that used &v, where v is a volatile variable, as being volatile.   I suggest that you use the existing function CheckIsNonModifying in the functions that are calling this instead.

lib/Sema/AvailableFactsAnalysis.cpp
```diff
@@ -0,0 +1,499 @@
...
+    AllExprs.insert(E);
+  for (auto I = St->child_begin(); I != St->child_end(); ++I)
+    CollectExpressions(*I, AllExprs);
+}
+
+// This function collects the defined variables in statement `St`.
+// We assume a variable is defined if it appears in the lhs of an assignment:
+// 1. increment or decrement operator (a++)
+// 2. assignment operator (a += 1, a = 2)
+void AvailableFactsAnalysis::CollectDefinedVars(const Stmt *St, std::set<const VarDecl *> &DefinedVars) {
```
dtarditi: This function could miss some definitions of variables.  In the unary operator case, the expression being incremented/decremented could be parenthesized.    In both cases, the LValueBitCast could also be parenthesized.  I think you need a loop that ignores parentheses and lvalue bit casts until nothing changes.  This should be factored into a separate function.

lib/Sema/SemaBounds.cpp
```diff
@@ -2151,6 +2156,62 @@ namespace {
...
+
+      if (!CreatedStdForm1 || !CreatedStdForm2)
+        return false;
+      if (IsOpSigned1 != IsOpSigned2)
+        return false;
+      if (ConstantPart1 != ConstantPart2)
+        return false;
+
+      bool EqualWithoutFacts = EqualValue(Ctx, VariablePart1, VariablePart2, EquivExprs);
+      bool LEWithFacts = FactExists(Ctx, VariablePart1, VariablePart2, EquivExprs, Facts);
```
dtarditi: The logic would be clearer to return true if EqualWithoutFact is true, and similarly return true if LEWtihFacts is true.   Then return false otherwise.

lib/Sema/AvailableFactsAnalysis.cpp
ppashakhanloo: Fixed.

lib/Sema/AvailableFactsAnalysis.cpp
ppashakhanloo: Fixed.

lib/Sema/AvailableFactsAnalysis.cpp
ppashakhanloo: Fixed.

lib/Sema/AvailableFactsAnalysis.cpp
ppashakhanloo: Thanks for explaining this. Fixed.

ppashakhanloo:
@dtarditi Thanks for your comments. I addressed all of them.
The code is ready for review.

lib/Sema/SemaBounds.cpp
```diff
@@ -2151,6 +2156,62 @@ namespace {
       return true;
     }
 
+    // This function is an extension of EqualExtended. It looks into the provided `Facts`
+    // in order to prove `Base1 + Offset1 <= Offset2 + Offset2`.
```
dtarditi: There's typo here: Offset2 is used twice.

Could you add a TODO comment: we are ignoring the possibility of overflow in the addition.

PR #659, opened 2019-08-07
Support running LNT tests locally

mgrang:


mgrang:
> Great to see this. However, I believe the changes as-is will break our existing LNT automation.

Yes, this will break our existing automation. That's why I did not want to merge this or put it on the wiki. This is only meant for internal usage for now.
I am working on a cleaner and more general solution which we can share outside.

mgrang:
I have generalized the scripts so that our automation does not break.

mgrang:
> Looks good, thanks! Could you add a comment to the top of the run-local script explaining how to use it and the need to install LNT?

Thanks. I will add a comment on the top about how to run it. I plan to add a more detailed how-to (including installation of necessary components) in the README.

mgrang:
Added usage guide and comments here. Added detailed setup and run instructions in https://github.com/microsoft/checkedc-llvm-test-suite/pull/95.

mgrang:
> Do the LNT tests have to be run from the Linux machines? Can we run these tests from Windows machine as well?

Yes, they can be run from Windows too. This README documents this: https://github.com/Microsoft/checkedc-llvm-test-suite/blob/master/README.md.

However, this patch only adds script to run LNT locally from Linux.

Issue #637, opened 2019-07-19
Generic Structs for Checked C


abeln:
This is a tracking issue for adding generic structs to Checked C.

abeln:
- Feature PR: https://github.com/microsoft/checkedc-clang/pull/638
- Tests PR: https://github.com/microsoft/checkedc/pull/370

abeln:
Support for itypes.
* feature: https://github.com/microsoft/checkedc-clang/pull/660
* spec tests: https://github.com/microsoft/checkedc/pull/375

abeln:
> 
> 
>     * Feature PR: #638
> 
>     * Tests PR: [microsoft/checkedc#370](https://github.com/microsoft/checkedc/pull/370)

abeln:
After the itypes PR above is merged, there's one remaining TODO. See https://github.com/microsoft/checkedc-clang/pull/660#discussion_r317836958

abeln:
Here's another TODO for this one David and I discussed:

Right now the representation of type applications is somewhat hacky. We use `RecordDecls` to represent both generic struct definitions _and_ their applications.

e.g. `struct Foo _For_any(T)` is a RecordDecl(isGeneric = true, typeParameters = [T])

and a subsequent application `struct Foo<int>` is a RecordDecl(isApplication = true, typeArguments=[T])

What we really want here is a dedicated type for type applications, such that `struct Foo<int>` is its own type node `TypeApp(pointer_to_foo_type, [list_of_type_arguments])`

Related to this is the idea that we need to properly canonicalize type applications. Right now each type application is its own canonical type, which leads to problems when caching them. What we want is the type application `TypeApp(Foo, [Arg])` to be canonicalized to `TypeApp(canon(Foo), [canon(Arg)])`. canon(Foo) will probably end up as `Foo`, but canon('Arg') might be != `Arg`: e.g. `canon(TypeApp(Foo, T)) = TypeApp(Foo, int)` if `T` has been typedef'ed to `int`.

abeln:
Assigning back to @dtarditi for future work.

PR #660, opened 2019-08-09
Support itypes for generic structs

abeln:
This PR adds support for having generic structs that contain interface types (itypes).

For example, if we have an existing `List` struct that's non-generic, we can annotate it
with `itypes` so that it has a generic interface.

```C
struct List _Itype_for_any(T) {
  void *head : itype(_Ptr<T>);
  struct List *next : itype(_Ptr<struct List<T>>);
};
```

In checked contexts, the compiler will force us to specify a type argument:
```C
checked {
  struct List<int> *li;
  li->head; // has type _Ptr<int>
  li->next; // has type _Ptr<struct List<int>>
}
```

In unchecked contexts, specify type arguments is optional. If no arguments are specified,
then we'll fill in `void` for all arguments.

```C
struct List<int> *li; // ok in unchecked context: type arguments are optional
struct List *lv; // equivalent to struct List<void>
lv->head; // has type void*
lv->next; // has type struct List<void>*
```
The changes to the compiler involve parsing itype_for_any type parameters, storing
them in the corresponding RecordDecl (just like _For_any), doing the type substitution in
itypes associated to fields, and involving field itypes in the expanding cycles check.


abeln:
The spec tests are in https://github.com/microsoft/checkedc/pull/375

lib/Parse/ParseDeclCXX.cpp
```diff
@@ -1655,6 +1655,14 @@ void Parser::ParseClassSpecifier(tok::TokenKind TagTokKind,
     ParseForanySpecifier(DS);
   }
 
+  if (Tok.is(tok::kw__Itype_for_any)) {
+    // TODO(abeln): add error handling here
```
dtarditi: Please remove your GitHub name from the comment.

lib/Sema/CheckedCSubst.cpp
```diff
@@ -140,6 +140,16 @@ void Sema::CompleteTypeAppFields(RecordDecl *Incomplete) {
...
     FieldDecl *NewField = FieldDecl::Create(Field->getASTContext(), Incomplete, SourceLocation(), SourceLocation(),
       Field->getIdentifier(), InstType, Field->getTypeSourceInfo(), Field->getBitWidth(), Field->isMutable(), Field->getInClassInitStyle());
+
+    // Substitute in the bounds-safe interface type (itype).
+    if (auto IType = Field->getInteropTypeExpr()) {
+      auto InstType = SubstituteTypeArgs(IType->getType(), Incomplete->typeArgs());
+      InteropTypeExpr *NewIType = new (Context) InteropTypeExpr(InstType, SourceLocation(), SourceLocation(), IType->getTypeInfoAsWritten());
+      NewField->setInteropTypeExpr(Context, NewIType);
+    }
+    // No substitution is needed for the bounds expression, because no types appear here.
```
dtarditi: Actually, the programmer could write code that uses the type variable in a bounds expression.

lib/Parse/ParseDeclCXX.cpp
```diff
@@ -1947,6 +1966,24 @@ void Parser::ParseClassSpecifier(tok::TokenKind TagTokKind,
         // We're parsing a reference to a generic struct, so we need to parse
         // the type arguments before we can instantiate.
         TagOrTempResult = ParseRecordTypeApplication(Decl);
+      } else if (Decl->isItypeGeneric()) {
+        if (Actions.IsCheckedScope() || Tok.is(tok::less)) {
+          // For an itype generic in a checked scope, we require type arguments as well.
+          // If the scope is unchecked but the user provides arguments, we allow that too.
+          TagOrTempResult = ParseRecordTypeApplication(Decl);
```
dtarditi: This is correct, but the programmer could get confused by the resulting error messages if they forget the type application.  The same problem could happen with previous clause for IsGeneric().   The message would be something like "expected '<'", which won't be meaningful.

I think you need to modify ParseRecordTypeApplication to issue a diagnostic if the initial '<' is missing.  The error message should say that we expected a list of type arguments.  (the error message for missing type arguments for generic functions is `err_expected_type_argument_list_for_generic_call`, defined in include\clang\Basic\DiagnosticSemaKinds.td).

lib/Parse/ParseDeclCXX.cpp
abeln: Done

lib/Sema/CheckedCSubst.cpp
abeln: @dtarditi 
Good point. Can types appear in bound expressions only inside `sizeof()`? If so, should we bother with doing the substitution?
e.g.

Suppose we have

```C
struct Foo _For_any(T) {
  _Array_ptr<int> p: count(sizeof(T*)); // non-sensical as a bound, but let's use it just for the sake of having a type in a bounds expression
};

struct Foo<int> f;
```

If (as things stand right now) we don't do the substitution T->int, then the bound on `f` will be `count(sizeof(T*))`, which is erased to `count(size(void*))`. Ultimately, this is the right value, because sizeof(T*) == sizeof(void*) for all T.

So not doing the substitution means we leave a `T` floating around when it shouldn't be, but the result seems to be correct.

The difficulty in doing the substitution is that we'd need to write an expression traversal function for substitution. Right now the substitution function we have operates on types.

Are you ok with leaving the type variable unchanged?

abeln: I've replaced the comment to be more accurate. Let me know if you want to do the substitution.

abeln:
@dtarditi Updated the comments and TODOs. Substitution in bounds expressions tracked in https://github.com/microsoft/checkedc-clang/issues/637

PR #667, opened 2019-08-19
Implemented the "*" operator for _MMSafe_ptr.

jzhou76:
When a mmsafe_ptr is dereferenced by the "*" operator and after the
null pointer checking is done, the compiler inserts a
DynamicStructIDCheck() function to check if the ID of the mmsafe_ptr
matches the ID of the pointee (a struct).

The algorithm is straightforward:

1. Extract the ID from the mmsafe_ptr.
2. Extract the ID from the struct (always the first field)
3. Insert an integer comparison instruction (ICmpInst) to compare the
   two IDs.
4. Continue executing if the IDs match; otherwise abort the program.

We also implemented a customized memory allocator and a deallocator,
and added tests for this commit. The customized allocator/deallocator
and the tests are in the checkedc repo:
https://github.com/microsoft/checkedc/commit/e4e3036386b82d848f82091bebddbdaa5b46d8b7

### DIFF
--- a/lib/CodeGen/Address.h
+++ b/lib/CodeGen/Address.h
@@ -25,11 +25,41 @@ namespace CodeGen {
 class Address {
   llvm::Value *Pointer;
   CharUnits Alignment;
+
+private:
+  bool _containMMSafePtr;
+  llvm::Type *originalPointerTy;
+  llvm::Type *rawPointerTy;  // The inner pointer type of a _MMSafe_ptr.
+
 public:
   Address(llvm::Value *pointer, CharUnits alignment)
       : Pointer(pointer), Alignment(alignment) {
     assert((!alignment.isZero() || pointer == nullptr) &&
            "creating valid address with invalid alignment");
+    if (pointer) {
+      // Backup the original _MMSafe_ptr type.
+      originalPointerTy = pointer->getType();
+
+      // Checked C: for _MMSafe_ptr, reset the poitner type.
+      if (pointer->getType()->isMMSafePointerTy()) {
+        _containMMSafePtr = true;
+        rawPointerTy = pointer->getType()->getInnerPtrFromMMSafePtr();
+        pointer->mutateType(rawPointerTy);
+      }
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


lib/CodeGen/Address.h
```diff
@@ -25,11 +25,41 @@ namespace CodeGen {
 class Address {
   llvm::Value *Pointer;
   CharUnits Alignment;
+
+private:
+  bool _containMMSafePtr;
```
sunnychatterjee: This data member will be uninitialized since not all paths through the constructor body initialize it. Please consider initializing it to prevent uninitialized memory bugs.

lib/CodeGen/Address.h
```diff
@@ -25,11 +25,41 @@ namespace CodeGen {
...
+      }
+    }
+  }
+
+  // Return true if this Address contains a _MMSafe_ptr.
+  bool containMMSafePtr() const { return _containMMSafePtr; }
+
+  //  Set the pointer type to be the inner pointer type of a _MMSafe_ptr.
+  void mutatePointerType() {
+    if (containMMSafePtr()) Pointer->mutateType(rawPointerTy);
```
sunnychatterjee: Can Pointer be null? If not, consider adding an assertion.

lib/CodeGen/Address.h
sunnychatterjee: Similar observation for the data members originalPointerTy and rawPointerTy. Consider initializing them to nullptr.

jzhou76: Thanks for the remind. I was aware of the uninitialization. I didn't bother to initialize them because until now the functions that use them are actually not called by any other functions. But you're right that it's good programming practice to always initialize the variables. I'll fix them soon.

lib/CodeGen/Address.h
jzhou76: @sunnychatterjee Pointer here cannot be null  because when we need call this `mutatePointerType()` function, we've already known that we're using a non-void pointer in the program and that means the `Pointer` field has already been initialized to be non-null.

PR #672, opened 2019-08-21
Do not verify alignment of load/store in unit tests

mgrang:
We do not verify the alignment of any loads/stores in the IR.

Example: "%1 = load i32*, i32** @gp1, align 8"

We discard the "align 8" at the end. This is because alignment is type-size
dependent, which in turn is platform dependent. On a 64-bit system the
alignment is 8 bytes but on 32 bit systems it could be 4 bytes.  It is
better to elide these checks, because we really care about the order of
loads, checks and stores, in these tests, not their specific alignment.

### DIFF
--- a/test/CheckedC/dynamic-checks/ptr-arith-null-check-code-gen.c
+++ b/test/CheckedC/dynamic-checks/ptr-arith-null-check-code-gen.c
@@ -9,7 +9,7 @@ array_ptr<char> q : count(3) = "abc";
 
 void f1() {
 // CHECK-LABEL: f1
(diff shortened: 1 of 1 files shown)


mgrang:
This fixes the failure with the 32 bit compiler seen in https://dev.azure.com/msresearch/CheckedC/_build/results?buildId=329231

mgrang:
Fix verified on X86 Windows toolchain. See https://dev.azure.com/msresearch/CheckedC/_build/results?buildId=329666.

PR #681, opened 2019-09-06
Update setup and build instructions for Visual Studio

dtarditi:
Recent versions of CMake have changed how you generate a build system for Visual Studio.   Update the Checked C clang setup and build instructions to reflect this, as well describe way of doing this with older versions of CMake.

### DIFF
--- a/docs/checkedc/Setup-and-Build.md
+++ b/docs/checkedc/Setup-and-Build.md
@@ -3,8 +3,8 @@
 ## Setting up your machine
 
 See the clang [Getting started guide](http://clang.llvm.org/get_started.html) for information
-on how to set up your machine. If you will be developing on Windows, you should install CMake 3.8
-or later on your machine.
+on how to set up your machine.  For Linux, you should install 3.8 or later.
+If you will be developing on Windows, you should install CMake 3.14 or later on your machine.
 
 ### Developing on Windows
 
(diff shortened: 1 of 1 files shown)


docs/checkedc/Setup-and-Build.md
```diff
@@ -141,13 +141,42 @@ git clone https://github.com/Microsoft/checkedc
...
 where `{llvm-path}` is the path to the root of your LLVM repo.
 
+The directions for generating a build system for Visual Studio depend on which
+version of CMake you are using.  You must use CMake 3.14 or higher to
+generate a build system for Visual Studio 2019.
+
+### Visual Studio with CMake 3.14 or higher
+
+If you are using CMake 3.14 or higher, you use the -G option to specify
+the geneerator (the target build system) and the -A option to specify
```
mgrang: Typo: geneerator --> generator

docs/checkedc/Setup-and-Build.md
```diff
@@ -156,9 +185,11 @@ where `{llvm-path}` is the path to the root of your LLVM repo.
...
 ```
 
-   On Windows, when using Visual Studio, CMake by default produces a build system for x86.  This means that
-   the clang tests will run in 32-bit compatiblity mode, even on a 64-bit version of Windows.  To build and run
-   tests on x64, specify a different generator using the `-G` option.  For Visual Studio 2017, you can use:
+   On Windows, when using Visual Studio, CMake versions earlier than 3.14
+   by default produce a build system for x86.  This means that the clang tests
+   will run in 32-bit compatiblity mode, even on a 64-bit version of Windows.
+   To build and run tests on x64, specify a different generator using the `-G`
+   option.  For    Visual Studio 2017, you can use:
```
mgrang: Extraneous space between "For    Visual".

PR #685, opened 2019-09-20
Fix release build failures for existential types.

dtarditi:
The release build of the Checked C clang compiler fails to compile the new tests for existential types.  The compiler crashes.  I investigated the crashes and found that the iterator functions for PackExpr were incorrect.  They need to return the address of the member, not the address of a local variable.

Also add missing classof method for PackExpr.

Testing:
- Checked C tests passed with a release with debug info compiler built on a local machine.


abeln:
@dtarditi Sorry I forgot to run the release tests. Thanks for fixing this!

Issue #669, opened 2019-08-20
Improve unit tests for null ptr arithmetic


mgrang:
In https://github.com/microsoft/checkedc-clang/pull/663 we added the flag `-fno-checkedc-null-ptr-arith` to several unit tests because these tests perform ptr arithmetic but do not expect null checks. So they fail. We need to fix the tests to check for null ptr before doing ptr arithmetic.

PR #687, opened 2019-09-24
Remove -fno-checkedc-null-ptr-arith flag from null ptr unit tests

kkjeer:
Modify dynamic checks tests to expect null pointer checks before doing null pointer arithmetic and remove the `-fno-checkedc-null-ptr-arith` flag (see issue #669).


mgrang:
> We also want to remove this flag from the clang code base.

On second thought, I think we can let this flag remain in the code base for measuring impact of the null ptr checks.

mgrang:
LGTM. You can merge once the ADO test run comes back clean.

Issue #691, opened 2019-10-02
IsValuePreserving treats Dynamic_bounds_cast and Assume_bounds_cast as value preserving


kkjeer:
For a function that involves bounds cast expressions, such as:
```
void f(_Array_ptr<char> a : bounds(
             _Assume_bounds_cast<_Array_ptr<int>>(a, count(7)),
             _Assume_bounds_cast<_Array_ptr<int>>(a, count(7)) + 4)
);

void f(_Array_ptr<char> a : bounds(
             _Dynamic_bounds_cast<_Array_ptr<int>>(a, count(7)), 
             _Assume_bounds_cast<_Array_ptr<int>>(a, count(7)) + 4)
) {}
```
Since `_Dynamic_bounds_cast` and `_Array_bounds_cast` are marked as value preserving, `_Assume_bounds_cast<_Array_ptr<int>>(a, count(7))` and `_Dynamic_bounds_cast<_Array_ptr<int>>(a, count(7))` are treated as lexicographically equal, and the above example compiles when it should not.

Issue #331, opened 2017-06-22
StmtProfile.cpp missing some bounds information


dtarditi:
- VisitRangeBoundsExpr needs to examine relative alignment information.
- VisitBoundsCastExpr needs to examine target bounds.

kkjeer:
The issue with bounds cast expressions is described in #691.

PR #692, opened 2019-10-03
Add relative alignment to StmtProfiler::VisitRangeBoundsExpr

kkjeer:
Visit the relative bounds clause (if any) for a RangeBoundsExpr in StmtProfiler::VisitRangeBoundsExpr. This means that examples such as the following now do not compile:

```
void f(_Array_ptr<int> p : bounds(p, p + 1) rel_align(char);
void f(_Array_ptr<int> p : bounds(p, p + 1) rel_align(double) {}
```

This addresses the first part of issue #331 (VisitRangeBoundsExpr needs to examine relative alignment information). Issue #691 tracks work related to VisitBoundsCastExpr and will be addressed in a separate PR.

Tests:
* Add bounds-rel-align.c to test function declarations involving parameters with rel_align and rel_align_value clauses
* Existing tests pass

### DIFF
--- a/lib/AST/StmtProfile.cpp
+++ b/lib/AST/StmtProfile.cpp
@@ -1298,6 +1298,19 @@ void StmtProfiler::VisitNullaryBoundsExpr(const NullaryBoundsExpr *S) {
 void StmtProfiler::VisitRangeBoundsExpr(const RangeBoundsExpr *S) {
   VisitExpr(S);
   ID.AddInteger(S->getKind());
+
+  if (S->hasRelativeBoundsClause()) {
+    RelativeBoundsClause *R = S->getRelativeBoundsClause();
+    switch (R->getClauseKind()) {
+    case RelativeBoundsClause::Kind::Type:
+      VisitType(cast<RelativeTypeBoundsClause>(R)->getType());
+      break;
+    case RelativeBoundsClause::Kind::Const:
+      VisitExpr(cast<RelativeConstExprBoundsClause>(R)->getConstExpr());
+      break;
+    }
+  }
+
 }
 
 void StmtProfiler::VisitInteropTypeExpr(
--- a/lib/AST/StmtProfile.cpp
+++ b/lib/AST/StmtProfile.cpp
@@ -1308,9 +1308,12 @@ void StmtProfiler::VisitRangeBoundsExpr(const RangeBoundsExpr *S) {
     case RelativeBoundsClause::Kind::Const:
       VisitExpr(cast<RelativeConstExprBoundsClause>(R)->getConstExpr());
       break;
+    default: {
+	}
     }
+  } else {
+	ID.AddPointer(nullptr);
   }
-
 }
 
 void StmtProfiler::VisitInteropTypeExpr(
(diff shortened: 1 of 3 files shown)


test/CheckedC/static-checking/bounds-rel-align.c
```diff
@@ -0,0 +1,142 @@
...
+             : bounds(p, p + 1) rel_align(short)); // expected-note {{previous bounds declaration is here}}
+
+void fail4(_Array_ptr<int> p
+             : bounds(p, p + 1) rel_align(struct S1)) {} // expected-error {{function redeclaration has conflicting parameter bounds}}
+
+//
+// Test declaration and definition of functions with a range bounds parameter with a rel_align_value clause.
+//
+
+const int val1 = 6;
```
sunnychatterjee: Have you tried with more complicated compile time expressions like constexpr int val3 = foo(); where "foo" returns '5+1' as a constexpr?

test/CheckedC/static-checking/bounds-rel-align.c
```diff
@@ -0,0 +1,142 @@
+// Tests for checking that function declarations involving range bounds
+// account for relative alignment information.
+//
+// RUN: %clang -cc1 -fcheckedc-extension -Wcheck-bounds-decls -verify %s
```
dtarditi: You don't need the -fcheckedc-extension flag.

test/CheckedC/static-checking/bounds-rel-align.c
dtarditi: This test is for C, not C++, so we don't have constexpr available.

lib/AST/StmtProfile.cpp
```diff
@@ -1298,6 +1298,22 @@ void StmtProfiler::VisitNullaryBoundsExpr(const NullaryBoundsExpr *S) {
...
+  if (S->hasRelativeBoundsClause()) {
+    RelativeBoundsClause *R = S->getRelativeBoundsClause();
+    switch (R->getClauseKind()) {
+    case RelativeBoundsClause::Kind::Type:
+      VisitType(cast<RelativeTypeBoundsClause>(R)->getType());
+      break;
+    case RelativeBoundsClause::Kind::Const:
+      VisitExpr(cast<RelativeConstExprBoundsClause>(R)->getConstExpr());
+      break;
+    default: {
```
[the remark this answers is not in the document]
dtarditi: I agree with this suggestion.

lib/AST/StmtProfile.cpp
```diff
@@ -1298,6 +1298,20 @@ void StmtProfiler::VisitNullaryBoundsExpr(const NullaryBoundsExpr *S) {
 void StmtProfiler::VisitRangeBoundsExpr(const RangeBoundsExpr *S) {
   VisitExpr(S);
   ID.AddInteger(S->getKind());
+
+  if (S->hasRelativeBoundsClause()) {
+    RelativeBoundsClause *R = S->getRelativeBoundsClause();
+	RelativeBoundsClause::Kind CK = R->getClauseKind();
+	if (CK == RelativeBoundsClause::Kind::Type) {
```
mgrang: You can leave out braces in single statement if's.

PR #694, opened 2019-10-11
Introduce temporary bindings to BoundsCastExprs

kkjeer:
This PR adds a temporary binding expression when building a BoundsCastExpr. This helps enable dynamic and assume bounds casts expressions to be treated as non-value-preserving in CanonBounds::CompareExpr (see issue #691).

### Changes
* Introduce temporary binding expression to bounds casts.
* Adjust bounds inference and visiting cast expressions to reflect the temporary bindings.
* Account for BoundsCastExprs when getting the temporary binding of an expression.
* Account for whether the expression is in a checked scope or not in `PruneTemporaryBindings`. (A follow-up issue #698 is opened to use the `CheckedScopeSpecifier` enum rather than a boolean here.)
* ~~Consider BoundsCastExprs to be value-preserving in CheckBoundsDeclarations::EqualValue. Note: this assumes that the expressions E1 and E2 have been successfully evaluated.~~ This change has been removed. This affects the Checked C static_check_bounds_cast.c test (modified in microsoft/checkedc#385 - opened #695 to track remaining work).
* Treat bounds cast expressions as non-value-preserving.
* Consider an expression e to be lexicographically equal to the usage of a temporary binding of e.
* Remove the llvm_unreachable call if a dynamic cast fails in CanonBounds::Compare. Since bounds casts are no longer considered value-preserving, it is possible to compare, e.g. a BoundsCastExpr and an ImplicitCastExpr. An ImplicitCastExpr cannot be dynamically cast to a BoundsCastExpr.
* Account for the cast kind when profiling a BoundsCastExpr.

### Tests
* Remove the function call test from bounds-decl-challenges. With the introduction of temporaries to bounds cast expressions, this test no longer produces an error.
* Add AST tests to verify that bounds cast expressions include a temporary binding expression.
* Other tests are added to checked-c (microsoft/checkedc#385).

### DIFF
--- a/lib/AST/CanonBounds.cpp
+++ b/lib/AST/CanonBounds.cpp
@@ -322,27 +322,23 @@ Result Lexicographic::CompareExpr(const Expr *Arg1, const Expr *Arg2) {
   // The use of an expression temporary is equal to the
   // value of the binding expression.
   if (BoundsValueExpr *BV1 = dyn_cast<BoundsValueExpr>(E1)) {
-	  CHKCBindTemporaryExpr *Binding = BV1->getTemporaryBinding();
+    CHKCBindTemporaryExpr *Binding = BV1->getTemporaryBinding();
     if (Binding == E2)
       return Result::Equal;
-	  if (Binding) {
-	    Expr *SubExpr = Binding->getSubExpr();
-	    if (CompareExpr(SubExpr, E2) == Result::Equal) {
-	      return Result::Equal;
-	    }
-	  }
-  }
 
+    if (Binding)
+      if (CompareExpr(Binding->getSubExpr(), E2) == Result::Equal)
+        return Result::Equal;
+  }
+  
   if (BoundsValueExpr *BV2 = dyn_cast<BoundsValueExpr>(E2)) {
-	  CHKCBindTemporaryExpr *Binding = BV2->getTemporaryBinding();
+    CHKCBindTemporaryExpr *Binding = BV2->getTemporaryBinding();
     if (Binding == E1)
       return Result::Equal;
-	  if (Binding) {
-	    Expr *SubExpr = Binding->getSubExpr();
-	    if (CompareExpr(SubExpr, E1) == Result::Equal) {
-	      return Result::Equal;
-	    }
-	  }
+
+    if (Binding)
+      if (CompareExpr(Binding->getSubExpr(), E1) == Result::Equal)
+        return Result::Equal;
   }
 
    // Compare expressions structurally, recursively invoking
@@ -460,9 +456,9 @@ Result Lexicographic::CompareExpr(const Expr *Arg1, const Expr *Arg2) {
        // - Pointer arithmetic where the pointer referent types are the same
        //   size, checkedness is the same, and the integer types are the
        //    same size/signedness.
-
-	     if (!isa<BoundsExpr>(E1ChildExpr))
-		     Cmp = CompareTypeIgnoreCheckedness(E1ChildExpr->getType(), E2ChildExpr->getType());
+       
+       if (!isa<BoundsExpr>(E1ChildExpr))
+         Cmp = CompareTypeIgnoreCheckedness(E1ChildExpr->getType(), E2ChildExpr->getType());
 
        if (Cmp != Result::Equal)
          return CheckEquivExprs(Cmp, E1, E2);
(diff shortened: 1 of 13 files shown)


lib/AST/CanonBounds.cpp
```diff
@@ -323,13 +321,31 @@ Result Lexicographic::CompareExpr(const Expr *Arg1, const Expr *Arg2) {
 
   // The use of an expression temporary is equal to the
   // value of the binding expression.
-  if (BoundsValueExpr *BV1 = dyn_cast<BoundsValueExpr>(E1))
-    if (BV1->getTemporaryBinding() == E2)
+  if (BoundsValueExpr *BV1 = dyn_cast<BoundsValueExpr>(E1)) {
+	CHKCBindTemporaryExpr *Binding = BV1->getTemporaryBinding();
```
dtarditi: Stylisitc issues: I believe the indentation of the declaration of Binding is off by one-space.

Clang/LLVM developers don't put single statements in blocks.  So:
```
if (Binding == E2) {
 return Result::Equal;
}
```
should be
```
if (Binding == E2) 
  return Result::Equal

File: lib/AST/CanonBounds.cpp [hunk 2]
```diff
@@ -323,13 +321,31 @@ Result Lexicographic::CompareExpr(const Expr *Arg1, const Expr *Arg2) {
...
+  }
 
-  if (BoundsValueExpr *BV2 = dyn_cast<BoundsValueExpr>(E2))
-    if (BV2->getTemporaryBinding() == E1)
+  if (BoundsValueExpr *BV2 = dyn_cast<BoundsValueExpr>(E2)) {
+	CHKCBindTemporaryExpr *Binding = BV2->getTemporaryBinding();
+    if (Binding == E1) {
       return Result::Equal;
+	}
+	if (Binding) {
```
dtarditi: The indentation seems off here.

lib/Sema/SemaBounds.cpp
```diff
@@ -2471,6 +2470,12 @@ namespace {
     }
 
     CHKCBindTemporaryExpr *GetTempBinding(Expr *E) {
+	  // Bounds casts should always have a temporary binding.
+	  if (BoundsCastExpr *BCE = dyn_cast<BoundsCastExpr>(E)) {
+		CHKCBindTemporaryExpr *Result = dyn_cast<CHKCBindTemporaryExpr>(BCE->getSubExpr());
```
dtarditi: The indentation should be 2 characters.

lib/Sema/SemaBounds.cpp
```diff
@@ -2234,7 +2222,18 @@ namespace {
     }
 
     static bool EqualValue(ASTContext &Ctx, Expr *E1, Expr *E2, EquivExprSets *EquivExprs) {
-      Lexicographic::Result R = Lexicographic(Ctx, EquivExprs).CompareExpr(E1, E2);
+	  // It is assumed that E1 and E2 have been successfully evaluated, so that dynamic and assume bounds casts can be treated as value preserving here.
```
dtarditi: I think you need to push this check into Lexicographic expression comparison.  There should be a special mode that answer the questions: if e1 and e2 evaluate to values, are the values guaranteed to be equal?  We can assume there that any evaluations of dynamic_bounds_cast in subexpressions of e1 (or 32) will succeed and any bounds checks that depend on bounds changed by *_cast operation will also succeed.

lib/AST/CanonBounds.cpp
```diff
@@ -446,8 +456,10 @@ Result Lexicographic::CompareExpr(const Expr *Arg1, const Expr *Arg2) {
        // - Pointer arithmetic where the pointer referent types are the same
        //   size, checkedness is the same, and the integer types are the
        //    same size/signedness.
-       Cmp = CompareTypeIgnoreCheckedness(E1ChildExpr->getType(),
-                                          E2ChildExpr->getType());
+       
+       if (!isa<BoundsExpr>(E1ChildExpr))
```
dtarditi: It'd be helpful to add a comment "Bounds expressions don't have types".

kkjeer:
> Looks great - nice to see code be deleted and become simpler! I have one suggested comment update. Please run automated testing before committing this,.

The automated testing has passed.

dtarditi:
If we were to add support for statement expressions in GCC (https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html) , an expression might not always be in the same checked scope.

Issue #697, opened 2019-10-18
Inference on empty bounds range


saeednj:
In the following code an array with bounds `R1: (a, a + 10)` is declared, but passed to a function `foo` with a negative element count which generates bounds `R2: (a, a - 5)`. This code compiles without any errors or warnings. The injected dynamic check is triggered at runtime though.
```c
#include <stdchecked.h>

int foo(array_ptr<int> a : count(n), int n)
{
    return a[n / 2];
}

int main()
{
    int i;
    int a checked[10];
    for( i=0; i<10; i++ ) a[i] = i;

    return foo(a, -5);
}
```
In the bounds inference logic, `R2` should be a sub-range of `R1` to be acceptable. Although the offsets are constant values, the compiler does not catch that `R2` is empty and therefore any access to `a` is a violation of bounds. 

We can handle this case by adding this snippet to the `BaseRange::InRange` function in `SemaBounds.cpp`:
```c
          if ( R.IsEmpty() ) {
            Cause = CombineFailures(Cause, ProofFailure::Empty);
            return ProofResult::False;
          }
```
The case of source bounds (`R1`) being empty seems to be handled by the code already, but should be tested.

PR #702, opened 2019-10-30
Handling empty ranges in bounds inference

saeednj:
When assigning dest <-- src, this property should hold: dest_range \subseteq src_range (subset or equal). Empty destination ranges pass this test as demonstrated in issue  #697.

Here we disallow empty destination ranges for pointers of type `_Array_ptr`, and disallowing "strictly" empty ranges for pointers of type `_Nt_array_ptr` (we allow the case of lowerBound==upperBound for null-terminated array types).


lib/Sema/SemaBounds.cpp
```diff
@@ -1827,6 +1828,16 @@ namespace {
       // Is R a subrange of this range?
       ProofResult InRange(BaseRange &R, ProofFailure &Cause, EquivExprSets *EquivExprs,
                           std::pair<ComparisonSet, ComparisonSet>& Facts) {
+
+        // In assignments dest <-- src, this relation should hold: range_dest \subseteq range_src
+        // range_dest == {} always passes this test, however we are disallowing empty ranges
+        // for _Array_ptr (upperBound <= lowerBound). With the exception that
+        // upperBound == lowerBound is allowed for _Nt_array_ptr.
+        if ( R.IsEmpty() ) {
```
mgrang: You can remove the spaces inside the if clause.

``
if ( R.IsEmpty() ) --> if (R.IsEmpty())
``

lib/Sema/SemaBounds.cpp
```diff
@@ -1932,12 +1943,18 @@ namespace {
...
-      // `UpperOffsetConstant <= LowerOffsetConstant`.
+      // `UpperOffsetConstant <= LowerOffsetConstant`,
+      // except for _Nt_Array_ptr types
+      // that `UpperOffsetConstant == LowerOffsetConstant` is allowed.
       // Currently, it returns false when the range is not ConstantSized.
       // However, this should be generalized in the future.
       bool IsEmpty() {
-        if (IsConstantSizedRange())
+        if (IsConstantSizedRange()) {
+          if ( Base->IgnoreParenCasts()->getType()->isCheckedPointerNtArrayType() )
```
mgrang: Same comment here.

PR #707, opened 2019-10-31
Handling empty ranges when inferring bounds

saeednj:
When assigning dest <-- src, this property should hold: dest_range \subseteq src_range (subset or equal). Empty destination ranges pass this test as demonstrated in issue  #697.

Here we disallow empty destination ranges for pointers of type `_Array_ptr`, and disallowing "strictly" empty ranges for pointers of type `_Nt_array_ptr` (we allow the case of lowerBound==upperBound for null-terminated array types).


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1827,6 +1828,16 @@ namespace {
       // Is R a subrange of this range?
       ProofResult InRange(BaseRange &R, ProofFailure &Cause, EquivExprSets *EquivExprs,
                           std::pair<ComparisonSet, ComparisonSet>& Facts) {
+
```
dtarditi: We talked about this in person.   We agreed that we want to allow empty bounds declarations: they might represent a zero-element array or an empty buffer.   We want to disallow memory access to empty ranges.  We may want to statically disallow ranges that are invalid (where the upper bound < lower bound).    There are some algorithms like quicksort where pointers intentionally cross - I'm wondering if that leads to "invalid ranges" in some formulations of quicksort.

Given this, I think it makes sense to make sure there is a check for an empty range in CheckBoundsAtMemoryAccess.  It doesn't matter what the actually access expression is.   It would also make sense at bounds declarations, calls, and static casts to check for invalid ranges.    I believe that should be done as a separate check on destination bounds for calls and bounds casts.   For variables, I believe you can add the check to VisitVarDecl (declared around line 3228).    At these locations, it will also be easiere to distinguish between ounds for nt_array_ptr and array_ptr.

saeednj: Previously we were disallowing LB == UB cases, which was problematic for nt_arrays. But now that we are allowing this case, is there any difference in the handling of nt_array_ptr vs. array_ptr?

saeednj:
The latest update implements:

- Declaration of _empty_ and _invalid_ ranges results in Warning.
- Memory access to empty and invalid bounds results in Error.
- All statically deducible out-of-bounds memory accesses were elevated from Warning to Error.
- Bounds of reading and writing to null-terminated arrays were matched to detect both reads and writes that are out-of-bound. Writing to the last element (only writing 0 is allowed according to specification), is checked at runtime.
- Tests were updated according to the new implementation.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -42,6 +42,7 @@
 #include <queue>
 
 // #define TRACE_CFG 1
+//#define DEBUG_BOUNDS 1
```
mgrang: Space between `//` and `#`.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -435,33 +436,33 @@ namespace {
...
-  Expr *PruneTemporaryBindings(Sema &SemaRef, Expr *E, CheckedScopeSpecifier CSS) {	
-    // Account for checked scope information when transforming the expression.
-    Sema::CheckedScopeRAII CheckedScope(SemaRef, CSS);
-
-    Sema::ExprSubstitutionScope Scope(SemaRef); // suppress diagnostics	
-    ExprResult R = PruneTemporaryHelper(SemaRef).TransformExpr(E);	
-    assert(!R.isInvalid());	
-    return R.get();	
-  }	
+// Convert all temporary bindings in an expression to uses of the values
```
mgrang: Nothing seems to have changed in this part of code?

clang/test/CheckedC/static-checking/bounds-decl-checking.c
```diff
@@ -310,7 +310,6 @@ _Array_ptr<int> f37(unsigned num) : count(num) {
   return p;
 }
 
-
```
mgrang: Please revert the deleted line.

clang/lib/Sema/SemaBounds.cpp
saeednj: It seems that the line ending of these lines in the master version is different than the other lines (0d0a instead of 0a).
Should I keep the master version, or mine?

mgrang: Let's keep the master version.

saeednj: Reverted back to CRLF line ending.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2450,6 +2484,16 @@ namespace {
       llvm::outs() << "Valid range:\n";
       ValidRange.Dump(llvm::outs());
 #endif
+#ifdef DEBUG_BOUNDS
```
dtarditi: This looks redundant with the prior #ifdef block.  Can you delete this?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2694,6 +2740,11 @@ namespace {
       ProofFailure Cause;
       ProofResult Result;
       ProofStmtKind ProofKind;
+      #ifdef DEBUG_BOUNDS
```
dtarditi: Can you use TRACE_RANGE instead?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2711,7 +2762,12 @@ namespace {
       }
 
       if (Result == ProofResult::False) {
-        unsigned DiagId = diag::warn_out_of_bounds_access;
+        #ifdef DEBUG_BOUNDS
```
dtarditi: Can you use TRACE_RANGE?

saeednj:
Thanks for the comments!

For the bullet points please see the latest commit.

Yes, I have updated the tests as well (mostly in memory-access-checking.c), so all CheckedC tests pass now.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1827,6 +1851,16 @@ namespace {
       // Is R a subrange of this range?
       ProofResult InRange(BaseRange &R, ProofFailure &Cause, EquivExprSets *EquivExprs,
                           std::pair<ComparisonSet, ComparisonSet>& Facts) {
+
+        // We will warn on declaration of Invalid ranges (upperBound < lowerBound).
+        // The following cases are handled by the callers of this function:
+        // - Warning on declaration of Empty ranges (upperBound == lowerBound).
```
dtarditi: I don't think we should warn on declaration of empty ranges.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2363,7 +2408,9 @@ namespace {
           return R;
         if (R == ProofResult::False || R == ProofResult::Maybe) {
           if (SrcRange.IsEmpty())
-            Cause = CombineFailures(Cause, ProofFailure::Empty);
+            Cause = CombineFailures(Cause, ProofFailure::SrcEmpty);
```
dtarditi: For this specific case, I believe we should not report a failure for this when R == ProofResult::Maybe.

clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -9950,12 +9950,21 @@ def err_bounds_type_annotation_lost_checking : Error<
   def error_static_cast_bounds_invalid : Error<
     "cast source bounds are too narrow for %0">;
 
+  def error_out_of_bounds_access : Error<
+    "out-of-bounds %select{||memory access|base value}0">;
+
```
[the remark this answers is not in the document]
saeednj: It is not used. I can remove it if there is no possible use case for it in the future.

dtarditi:
Yes, it would be good to remove it.

________________________________

clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -9950,12 +9950,17 @@ def err_bounds_type_annotation_lost_checking : Error<
...
 
-  def warn_out_of_bounds_access : Warning<
-    "out-of-bounds %select{||memory access|base value}0">,
-    InGroup<CheckMemoryAccesses>;
+  def error_out_of_bounds_access : Error<
+    "out-of-bounds %select{||memory access|base value}0">;
 
   def note_source_bounds_empty : Note<"source bounds are an empty range">;
 
+  def note_source_bounds_invalid : Note<"source bounds are an inavlid range">;
```
dtarditi: There's a typo in the error message: invalid is misspelled.

Issue #717, opened 2019-11-06
Disallow assignments of values that use free type variables to variables with static storage


kkjeer:
(Follow-up from #684)

Example of how things can go wrong:
```
_For_any(T) void f1(_Ptr<T> val) {
  static _Ptr<int> static_ptr;
  // should not be allowed - val uses free type variable T
  static_ptr = (_Ptr<int>)val; 
  printf(" *static_ptr = %d\n", *static_ptr);
}

_For_any(T) void f2(_Ptr<T> val) {
  static struct S<int> static_struct = { 0 };
  // should not be allowed - val uses free type variable T
  static_struct.member = (_Ptr<int>)val; 
  printf(" *static_struct.member: %d\n", *static_struct.member);
}

int main() {
  short a = 0;
  f1<short>(&a);
  double b = 3.14159;
  f1<double>(&b); // corrupt memory at static_ptr

  int c = 0;
  f2<int>(&c);
  float d = 2.71828;
  f2<float>(&d); // corrupt memory at static_struct.member
}
```
Output when compiled with the Checked C clang compiler:
```
 *static_ptr = 0
 *static_ptr = -266631570
 *static_struct.member: 0
 *static_struct.member: 1076754509
```

dtarditi:
This is complete.

Issue #684, opened 2019-09-19
Disallow use of type variables for variables with static storage.


dtarditi:
We cannot allow static variables to have types with free type variables.   The storage for static variables in generic functions persists and could be used with different generic type instantiations. This creates a soundness  issue.  

Here is an example of how things can go wrong:
```
#include <stdbool.h>
#include <stdio.h>

_For_any(T) _Ptr<T> get_set(int  isSetter, _Ptr<T> val) {
	static _Ptr<T> capture;
	if (isSetter) {
		capture = val;
		return 0;
	}
	else
		return capture;
}

int main(int argc, _Array_ptr<_Nt_array_ptr<char>> argv : count(argc)) {
	short j = 0;
	short i = 1;
	get_set<short>(true, &i);
	_Ptr<double> p = get_set<double>(false, 0);
	*p = 3.141516; // corrupt memory
	printf(" j = %d", j);
}
```
When compiled by the Checked C clang compiler on Windows x64,, the store through p overwrites j.  Here is the output.
```
D:\checkedc1\llvm\tools\clang\test\CheckedC>a.exe
 j = 9214
```

PR #718, opened 2019-11-06
Prevent static variables declarations from using free type variables

kkjeer:
(See #684) 
Emit an error for each usage of a free type variable in the declaration of a static variable. The following: 
```
struct S _For_any(T, U) { };
_For_any(T, U) void f(void) {
  static struct S<T, U> s;
}
```
will emit the following:
```
error: static variable 's' has a type that uses a type variable bound in an enclosing scope (type is 'struct S<T, U>' and type variable is 'T')
note: type variable 'T' declared here
error: static variable 's' has a type that uses a type variable bound in an enclosing scope (type is 'struct S<T, U>' and type variable is 'U')
note: type variable 'U' declared here
```
These errors are emitted at the location of the static variable. For example:
static-free-type-variable-error-message-1

Future work: disallow free type variables in assignments to static variables (see #717)

Testing:
* Added typechecking tests to Checked C to verify that static variable declarations cannot use free type variables (microsoft/checkedc#392)

### DIFF
--- a/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -246,7 +246,8 @@ def warn_vector_long_decl_spec_combination : Warning<
   "Use of 'long' with '__vector' is deprecated">, InGroup<Deprecated>;
 
 def err_static_decl_uses_free_type_variable : Error<
-  "type '%0' for static variable '%1' cannot use type variable '%2' that is bound by an enclosing scope">;
+  "static variable '%0' has a type that uses a type variable bound in an enclosing scope "
+  "(type is '%1' and type variable is '%2')">;
 def note_free_type_variable_declared : Note<"type variable '%0' declared here">;
 def err_redeclaration_different_type : Error<
   "redeclaration of %0 with a different type%diff{: $ vs $|}1,2">;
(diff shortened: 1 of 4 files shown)


clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -245,6 +245,9 @@ def err_invalid_vector_long_double_decl_spec : Error<
 def warn_vector_long_decl_spec_combination : Warning<
   "Use of 'long' with '__vector' is deprecated">, InGroup<Deprecated>;
 
+def err_static_decl_uses_free_type_variable : Error<
```
dtarditi: I think the error messages needs to be changed to be more understandable by C programmers.  Most of them won't know what a free type variable.   How about something like the following?  "static variable cannot have a type variable bound by an enclosing scope"

clang/lib/Sema/CheckedCSubst.cpp
```diff
@@ -515,6 +517,13 @@ class FreeVariablesFinder : public TreeTransform<FreeVariablesFinder> {
     return std::vector<const TypeVariableType *>(FreeVars.begin(), FreeVars.end());
   }
 
+  /// Returns the list of free typedef declarations referenced in the given type.
+  /// Typedef declarations enable more readable diagnostics than type variable types.
+  std::vector<const TypedefNameDecl *> findTypedefDecls(QualType Tpe) {
```
dtarditi: Instead of Type, I'd suggest Ty as the variable name.

clang/include/clang/Basic/DiagnosticSemaKinds.td
dtarditi: I think you need to include the type and indicate in the message that the problem is with type.  Something akin to:

"type for static variable '%0' cannot use a type variable '%1' that is bound by an enclosing scope"

or 

"static variable '%0' has a type '%1' that uses a type variable bound by an enclosing scope'

PR #724, opened 2019-11-11
For _Nt_array_ptr's, set bounds as the inferred bounds

mgrang:
For `_Nt_array_ptr`'s, set the bounds as those inferred from the initialization.
For example,

` _Nt_array_ptr<char> p = "abc";`
`Here, DeclaredBounds(p) = [0, 0] and InferredBounds(p) = [0, 3).`
`So, bounds(p) = [0, 3).`

Refer issue https://github.com/microsoft/checkedc-clang/issues/725

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -46,6 +46,57 @@
 using namespace clang;
 using namespace sema;
 
+// Common helper functions used by various classes in this file.
+namespace {
+  // Determine if the mathematical value of I (an unsigned integer) fits within
+  // the range of Ty, a signed integer type.  APInt requires that bitsizes
+  // match exactly, so if I does fit, return an APInt via Result with exactly
+  // the bitsize of Ty.
+  static bool Fits(clang::ASTContext &Ctx, QualType Ty,
+                   const llvm::APInt &I, llvm::APInt &Result) {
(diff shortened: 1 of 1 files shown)


mgrang:
This patch breaks these four existing tests:
``
    checkedc :: checkedc/tests/dynamic_checking/bounds/deref_arith_call_expr_opt.c
    checkedc :: checkedc/tests/dynamic_checking/bounds/nullterm_pointers.c
    checkedc :: checkedc/tests/dynamic_checking/bounds/subscript_call_expr_opt.c
    checkedc :: checkedc/tests/static_checking/bounds_decl_checking.c
``
Working on fixing these and adding more tests.

mgrang:
The next step would be to handle bounds in case of assignments. For example,

`_Nt_array_ptr<char> s = "ab";`
`_Nt_array_ptr<char> t = "a";`
`Here, bounds(s) = [0, 2) and bounds(t) = [0, 1) ==> bounds(t) < bounds(s).
`s = t;`

Now, the `bounds(s)` should be narrowed to `bounds(t)`. This would require a flow-sensitive analysis. I will handle this in a separate patch.

mgrang:
For now, we do not want to infer the bounds for nt_array_ptr automatically. We would rather rely on the user specifying the bounds explicitly. We would then need to do a dataflow analysis to widen the bounds. So this PR is no longer needed.

PR #728, opened 2019-11-13
Bounds widening for _Nt_array_ptr's

mgrang:
`_Nt_array_ptr`'s are terminated by a null terminator. If we determine that the dereference of a
`_Nt_array_ptr` is non-null then we can widen the bounds of the pointer by 1 because there will always be at least the null terminator in the string.

Formally we can state the problem as:
`∀ p | p ∈ _Nt_array_ptr & bounds(p) = [l, u)`
`*p != 0 => bounds(p) = [l, u + 1)`

Example:
`_Nt_array_ptr<char> p : count(0) = "abc"; // bounds = [0, 0]`
`if (*p) { // bounds = [0, 1) }`


mgrang:
Dataflow equations:

`Let p be a _Nt_array_ptr with declared bounds [l, u)`
`For each basic block B:`
`If branch_condition(B) is of the form "if (*p)", then`
`Gen(B) = { bounds(p) = [l, u + 1) }`
`Kill(B) = { bounds(p) in In(B) }`
`In(B) = { ∩ Out(B') } where B' ∈ preds(B).`
`Out(B) = { ((In(B) - Kill(B)) ∪ Gen(B) }. }`

mgrang:
Please review this PR as WIP. I am currently testing it out. Will add unit tests soon.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -0,0 +1,163 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "clang/Sema/BoundsAnalysis.h"
+
+using namespace llvm;
+
+namespace clang {
+
+void BoundsAnalysis::WidenBounds() {
+  llvm::dbgs() << "### Debug bounds analysis\n";
```
mgrang: Will remove the debug messages soon.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -0,0 +1,220 @@
...
+bool BoundsAnalysis::ContainsPointerDeref(const Expr *E) const {
+  if (const auto *CE = dyn_cast<CastExpr>(E)) {
+    if (CE->getCastKind() != CastKind::CK_LValueToRValue)
+      return false;
+    return IsPointerDerefLValue(CE->getSubExpr());
+  }
+  return false;
+}
+
+BoundsMap BoundsAnalysis::Intersect(BoundsMap &A, BoundsMap &B) {
```
saeednj: Maybe add a comment that "Intersect" mutates the first argument. Or make the second argument of "intersect", "union", "differ" to be const?

mgrang:
> Is it useful to have a base class for "DataFlowAnalysis" and implement case specific uses?
> There is a very similar implementation with a different usage in AvailableFactAnalysis.

Yes, the goal is ultimately to have a common base. We can do that in a separate clean-up PR.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -0,0 +1,289 @@
...
+      Intersections = PredEB->Out;
+      ItersectionEmpty = false;
+    } else
+      Intersections = Intersect(Intersections, PredEB->Out);
+  }
+  EB->In = Intersections;
+}
+
+BoundsMapTy BoundsAnalysis::UpdateOutMap(ElevatedCFGBlock *EB) {
+  // Out(B) = max(Gen(B) u In(B)).
```
dtarditi: If you reformulate Gen to be "an expression of the form *e" known to be non-zero", then you could walk over all the bounds in In and widen them if possible.  Specifically, for any variable v in In, if the bounds of v have the form (low, high + i) and e has the form (high + i), then the bounds can widened by 1.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -0,0 +1,289 @@
...
+        // Worklist is a SetVector. So it will only insert a key if it does not
+        // already exist.
+        Worklist.insert(BlockMap[succ]);
+    }
+  }
+
+  CollectWidenedBounds(BlockMap);
+}
+
+void BoundsAnalysis::UpdateGenMap(ElevatedCFGBlock *EB, BlockMapTy BlockMap) {
```
dtarditi: Normally, a gen set in a dataflow analysis is something that is created by that basic  block.  For example, in constant propagation, a  basic block may contain one or more assignments of variables to constants.  That in turn would cause the basic block to generate a map whose domain is variables definitely known to be constant at the end of the basic block and whose range is the constants values.

That doesn't appear to the intent of your gen set here.   It is unusual that that gen set depends on predecessor gen sets.  I'm also not following the counting that you are doing for each variable declaration in a gen set.  The counts don't appear to correspond to offsets; they appear to be counts of the number if times that a variable was tested.  For example,
```
if (*p) {
  if (*p) {
     // EB->Gen(p) would be 2 here
```
I think what you want to compute for a gen set is whether each predecessor block ends in the same `if (*e)` and that the current block is the true branch if that if.    To widen bounds at the entry to a block, you would walk over all the known bounds, and see if the upper bound of one of them is *e.  If so, then that bound could be widened by 1.

clang/test/CheckedC/inferred-bounds/widened-bounds.c
```diff
@@ -0,0 +1,154 @@
...
+  if (*p) {}
+
+// CHECK: In function: f1
+// CHECK:  [B1]
+// CHECK: upper_bound(p) = 1
+}
+
+void f2() {
+  int a;
+  if (*p) {
```
dtarditi: This isn't correct, the bounds aren't widened by subsequent tests of (*p).    You would want the code to have the form
```
if (*p)   // bounds are now count(1)
 if (*(p + 1))   // bounds are now count(2)
   if (*(p + 2)) // bounds are now count(3)

File: clang/lib/Sema/BoundsAnalysis.cpp [hunk 6]
```diff
@@ -0,0 +1,436 @@
...
+          ++WidenedBounds;
+
+        EB->WidenedBounds.insert(std::make_pair(V, WidenedBounds));
+      }
+    }
+  }
+}
+
+void BoundsAnalysis::ComputeGenSets(BlockMapTy BlockMap) {
+  // If there is an edge B1->B2 and the edge condition is of the form
```
mgrang: I need to update some of these comments.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -0,0 +1,440 @@
...
+      Intersections = Intersect(Intersections, PredEB->Out[EB->Block]);
+  }
+
+  EB->In = Intersections;
+}
+
+void BoundsAnalysis::ComputeOutSets(ElevatedCFGBlock *EB,
+                                    BlockMapTy BlockMap,
+                                    WorkListTy &WorkList) {
+  // Out[B1->B2] = In[B1] u Gen[B1->B2].
```
dtarditi: I think you should do the widening of bounds here, not with a separate function as you are doing.  The widening should happen along the true branch of an if statement that checks the value stored at the (potentially widened) upper bound of an nt_array_ptr variable.   I recommend something like the following computation:
1. Let Tmp  = In[B1].   Remove (kill) any entries in Tmp for any variable v with bounds(e1, e2),  where v or a variable in e1 or e2 is assigned to in B1.
2. For successors s of B1, let Out[B1->S] = Tmp.
3. Widen: If  B1 ends in the form "if (*(v + k))"  (where k could = 0, i.e. if (*v)). and v is an nt_array_ptr,
- Compute the upper bound u of v.  If v has no entry in  Tmp, and the declared upper bound of v is u', let u = u'.   If v has an entry in Tmp, and the declared bounds are u', where u', has the form (p + j), let u = p + (j + Tmp[v]).
- If u has the form v + k, widen the bounds to v + (k+1), i.e. set Tmp[v] = k+1.
- Let B2 be the true branch of "if (*(v + k))".  Let Out[B1->B2] = Tmp.

mgrang: The analysis is edge-based. I need to handle sequential access of the `_Nt_array_ptr`'s.
For example,

`if (*(p)) //bounds: 1`
` if (*(p+1)) //bounds: 2`
 `if (*(p+2)) //bounds: 3`

`if (*(p)) //bounds: 1`
 `if (*(p+2)) //bounds: 1`
` if (*(p+3)) //bounds: 1`

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -0,0 +1,491 @@
...
+  // Example:
+  // _Nt_array_ptr<char> p : bounds(p + i, p + j);
+  // LowerExpr: p + i.
+  // UpperExpr: p + j.
+  if (const auto *RBE = dyn_cast<RangeBoundsExpr>(E)) {
+    CollectBoundsVars(RBE->getLowerExpr(), BoundsVars);
+    CollectBoundsVars(RBE->getUpperExpr(), BoundsVars);
+  }
+
+  // Collect bounds vars for the LHS and RHS of bounds exprs.
```
dtarditi: Did you mean LHS and RHS of a binary expression?

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -0,0 +1,491 @@
...
+    if (!Exp)
+      return;
+
+    // TODO: Handle accesses of the form:
+    // "if (*(p + i) && *(p + j) && *(p + k))"
+
+    // For conditions of the form "if (*p)".
+    if (const auto *D = dyn_cast<DeclRefExpr>(Exp)) {
+      if (const auto *V = dyn_cast<VarDecl>(D->getDecl())) {
+        if (V->getType()->isCheckedPointerNtArrayType()) {
```
dtarditi: There's some confusion about the meaning of the gen entries.  I think you want gen to be a dereference expression, not a constant offset.  When you widen bounds, you need to check that the dereference is at exactly the upper bound of the variable.

Currently gen entries are  constant offsets.  The question is what is an entry a  constant offset from.   If it is from a variable, then we can't handle cases where the upper bound constant has the from v + i, where v is a variable and i is a variable too.    If it is from the upper bound of v, that creates a problem, because the upper bound can change if the bounds of v are widened.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -0,0 +1,491 @@
...
+    // run into cases like "if (*(p + 100))". In this case, we cannot
+    // initialize the bounds of p. So additionally we check if Gen[p] == 0.
+
+    // 2. Else if the bounds of p in (In - Kill) == Gen[V] then widen the
+    // bounds of p by 1.
+    // Consider this example:
+    // B1: if (*p) { // In[B1] = {}, Gen[Entry->B1] = {} ==> bounds(p) = 1.
+    // B2:   if (*(p + 1)) { // In[B2] = {p:1}, Gen[B1->B2] = {p:1} ==> bounds(p) = 2.
+    // B3:     if (*(p + 2)) { // In[B2] = {p:2}, Gen[B1->B2] = {p:2} ==> bounds(p) = 3.
+
```
dtarditi: At this point, you need to check for each variable with bounds whether the dereference recorded in Gen[succ] widens bounds.  

There are going to be casees to handle:
1. A case where the variable has been widened, i.e. has an entry in Diff.
2. A case the variable has not been widened, i.e. does not have an entry diff.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -0,0 +1,491 @@
...
+    EB->Out[succ] = Union(Diff, EB->Gen[succ]);
```
dtarditi: For each variable v with an entry in EB->Gen[succ], Union needs to widen bounds by checking that the access  in gen is at the upper bound specified in Diff.

mgrang:
Thanks for your review, David. If we change Gen to be an expression instead of a constant offset how would we compute the intersection of sets? One option is to assign a number to each expression in Gen. Numbers would increase from entry to exit for a particular V. Then to compute intersection we choose the expr with the smaller number. This would represent the LUB of the sets.

clang/test/CheckedC/inferred-bounds/widened-bounds.c
```diff
@@ -0,0 +1,249 @@
...
+    p = "xy";
+    if (*(p + 1))
+      a = 1;
+  }
+
+// CHECK: [B8]
+// CHECK:   1: p = "xy"
+// CHECK:   2: *(p + 1)
+// CHECK: upper_bound(p) = 1
```
dtarditi: This is wrong.  You need to widen against the upper bound of p, which is declared to p + j.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -0,0 +1,491 @@
...
+    ComputeInSets(EB, BlockMap);
+    ComputeOutSets(EB, BlockMap, WorkList);
+  }
+
+  CollectWidenedBounds(BlockMap);
+}
+
+void BoundsAnalysis::ComputeGenSets(BlockMapTy BlockMap) {
+  // If there is an edge B1->B2 and the edge condition is of the form
+  // "if (*(p + i))" then Gen[B1] = {B2, p:i} .
```
dtarditi: Could you clarify whether i is a constant or a variable?   I know that i is a constant from context, but it could confuse people later on.

Gen sets should be constant offsets relative to the declared upper bound of a pointer.     If p has declared upper bounds of e, and this is a dereference of *(e + i), then Gen[B1] = i.   You'll need to handle expressions that are equivalent to `e + i`.  Note that we already have methods for checking the equivalence of expressions.

clang/lib/Sema/BoundsAnalysis.cpp
dtarditi: I agree that gen can remain a constant offset.  However, it should be a constant offset relative to the declared upper bound of a pointer variable.

mgrang:
> @mgrang, you are correct that changing gen sets to be expressions would complicate the algorithm. The gen set entries should be constant offsets relative to the declared upper bound of an nt_array_ptr variable. Given
> 
> ```
> nt_array_ptr<char> p : bounds(e1, e2)
> ```
> 
> it is not correct to widen p given `*p`. `p` should be widened given a dereference of the form `*(e2 + k)`, where `k` is a constant integer.
> 
> Note that we have methods for determining whether some expression e3 is equivalent to the expression e2. See AST\CanonBounds.

Thanks David. I think we still have a complication. Let's say we have a case like:
`nt_array_ptr<char> p : bounds(p, p + 1)`

and then we encounter `*(p + 2)` which is essentially *((p + 1) + 1). How do we separate `k` from *(e) here? What about accesses like `*(i + j + e + k + l)`. This can get complicated quickly!

I think a simpler option is to keep the implementation we have now as is, meaning the bounds widened by this algorithm are not relative to the declared upper bound. Instead we "adjust" the widened bounds at the call site of this algorithm. Let's say the widened bounds for `p` returned by this algorithm is `k` and the declared bounds of `p` are `(l, u)`.

If `(l + k) <= u`, we do not need to widen.
If `(l + k) > u`, we can widen the bounds by `u - (l + k)`.

For example:
`nt_array_ptr<char> p : bounds(p, p + 2)`

`if (*p) // calc_bounds: 1, adjusted_bounds: p + 1 <= p + 2 => no widening done`
&nbsp;&nbsp;  `if (*(p + 1)) // calc_bounds: 2, adjusted_bounds: p + 2 <= p + 2 => no widening done`
 &nbsp;&nbsp;&nbsp;&nbsp;   `if (*(p + 2)) // calc_bounds: 3, adjusted_bounds: p + 3 > p + 2 => bounds widened by (p + 3) - (p + 2) = 1.`
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;     `if (*(p + 3)) // calc_bounds: 4, adjusted_bounds: p + 4 > p + 2 => bounds widened by (p + 4) - (p + 2) = 2`

dtarditi:
I agree it could get complicated quickly. The way we will handle this is by putting expressions in a standard form, such that the constant is on the outside.  We already face similar problems when checking bounds declarations.

 The problem with the approach that you propose is that we cannot handle simple cases like widening bounds of a variable v with bounds(v, v + len).  

I think it would be appropriate  to address additional cases and using standard forms in separate PRs, and finish this PR.  How about you add a check that the upper bound is the variable v?  This would bring it into agreement with what I am proposing, and let us finish the current PR.

mgrang:
> I agree it could get complicated quickly. The way we will handle this is by putting expressions in a standard form, such that the constant is on the outside. We already face similar problems when checking bounds declarations.
> 
> The problem with the approach that you propose is that we cannot handle simple cases like widening bounds of a variable v with bounds(v, v + len).
> 
> I think it would be appropriate to address additional cases and using standard forms in separate PRs, and finish this PR. How about you add a check that the upper bound is the variable v? This would bring it into agreement with what I am proposing, and let us finish the current PR.

> How about you add a check that the upper bound is the variable v?
Could you please clarify this? Should we only allow bounds widening when `nt_array_ptr<T> v : bounds(v, v)`? Or when the dereference is at the upper bound?

dtarditi:
I am suggesting that in this PR, we only widen bounds for nt_array_ptr variables whose bounds are bounds(v, v) or count(0).   I am suggesting that in a future PR, we generalize this to widen bounds for dereferences involving constant offsets from the declared upper bound of a variable.

mgrang:
Done. Latest PR widens bounds only when the declared bounds are count(0) or bounds(v, v). Also removed the more generic tests and replaced them with only tests specific to count(0) bounds.

mgrang:
ADO Validations Passed:
https://dev.azure.com/msresearch/CheckedC/_build/results?buildId=395620

Issue #731, opened 2019-11-21
Fix merging of function declarations involving bounds-safe interfaces.


saeednj:
@dtarditi 's note on merging function redeclarations:

> A function declaration with a bounds-safe interface for a parameter or return
> value is compatible with a declaration that omits the bounds-safe interface.
> When we merge a function redeclaration with a prior declaration of the
> function, the merged function type reflects any prior bounds-safe interfaces
> or parameters or return values. However, in the merged function declaration,
> the parameter declarations and the return value do not reflect the merged
> information.
> 

This problem was observed when reading the bounds on a pointer dereference, where the pointer was declared in a function argument in an unchecked function. Because of the merging problem discussed above, bounds-safe information was not available.

Also, in the context of symbolic execution of the clang static analyzer, for each function declaration, there is a list of symbolic values representing arguments. Reading the bounds-safe information off of canonical declaration leaves us with two copies of symbolic variables referring to variables in different declarations. Which makes the symbolic values of arguments of one declaration and bounds expression of another declaration incomparable.

PR #732, opened 2019-11-21
Fix merging of function declarations involving bounds-safe interfaces

saeednj:
This fixes function merging to propagate the information to parameter declarations in the function redeclaration. This in turn allows us to check that function definitions respect their bounds-safe interface.

See the related issue #731 for more details.


saeednj:
Thanks! Added the test.

The master fails at this test, and the code in this branch passes the test.

PR #739, opened 2019-12-10
Unary operator: combine bounds inference and checking

kkjeer:
#### Overview
This pull request is the first step in refactoring the CheckBoundsDeclarations class to combine bounds inference and bounds checking in the same methods. The first step adds a SideEffects argument to suppress any side effects that occur as a result of bounds checking (to prevent duplicate side effects that can otherwise occur from calling, e.g. LValueBounds on an expression). It also replaces the former VisitUnaryOperator method (which performed bounds checking on a unary operator) with the CheckUnaryOperator method (which performs both bounds inference and, if enabled, bounds checking on a unary operator).

#### Preventing unwanted errors, asserts, and unreachables
This change allows bounds inference methods such as LValueBounds and PruneTemporaryBindings to be called on a wider variety of expressions than before. Since bounds inference is performed at the same time as bounds checking, bounds inference is now done on all unary operators that are visited by TraverseStmt. These means that some asserts, unreachables, and error messages are now removed or suppressed to avoid unwanted side effects and test failures (see comments below for details).

#### Testing:

#### Future work:
* Convert remaining Visit* methods to Check* methods that combine bounds inference and checking
* Modify TraverseStmt to only check substatements if the called Check* method did not check subexpressions
* Remove RValueBounds method (replace with calls to TraverseStmt)

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -1891,7 +1891,7 @@ namespace {
 
       switch (S->getStmtClass()) {
         case Expr::UnaryOperatorClass:
-          VisitUnaryOperator(cast<UnaryOperator>(S), CSS);
+          CheckUnaryOperator(cast<UnaryOperator>(S), CSS, SideEffects::Enabled);
           break;
         case Expr::CallExprClass:
           VisitCallExpr(cast<CallExpr>(S), CSS, Facts);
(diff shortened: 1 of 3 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2209,17 +2218,67 @@ namespace {
...
+          if (NeedsBoundsCheck && DumpBounds)
+            DumpExpression(llvm::outs(), E);
+        }
+      }
+
+      // `*e` is not an rvalue.
+      // TraverseStmt (and CheckUnaryOperator) may be called on non-rvalues,
+      // so this is not unexpected behavior.
+      if (Op == UnaryOperatorKind::UO_Deref)
+        return CreateBoundsInferenceError();
```
kkjeer: Since CheckUnaryOperator combines bounds inference and checking, and is called on every unary operator visited by TraverseStmt, is no longer guaranteed or expected that this (rvalue) bounds inference will only be done on rvalue unary operators. Adding in the previous unreachable call causes many test failures in Checked C, including static_checking/bounds_decl_checking.c

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2680,19 +2739,17 @@ namespace {
       case Expr::UnaryOperatorClass: {
         UnaryOperator *UO = cast<UnaryOperator>(E);
         if (UO->getOpcode() == UnaryOperatorKind::UO_Deref)
-          return RValueBounds(UO->getSubExpr(), CSS);
-        else {
-          llvm_unreachable("unexpected lvalue unary operator");
+          return RValueBounds(UO->getSubExpr(), CSS, SideEffects::Disabled);
+        else
```
kkjeer: Since CheckUnaryOperator can be called on any kind of unary operator visited by TraverseStmt, it is no longer guaranteed that any calls to LValueBounds made by CheckUnaryOperator will only be on deref operator lvalues. Adding in this unreachable call causes the Clang Sema/complex-imag.c test to fail.

clang/lib/Sema/SemaChecking.cpp
```diff
@@ -13945,7 +13945,8 @@ void Sema::CheckArgumentWithTypeTag(const ArgumentWithTypeTagAttr *Attr,
 
 void Sema::AddPotentialMisalignedMembers(Expr *E, RecordDecl *RD, ValueDecl *MD,
                                          CharUnits Alignment) {
-  MisalignedMembers.emplace_back(E, RD, MD, Alignment);
+  if (!DisableSubstitionDiagnostics)
+    MisalignedMembers.emplace_back(E, RD, MD, Alignment);
```
kkjeer: When this method is called from PruneTemporaryBindings (as a result of calls to CheckUnaryOperator), there should be no (duplicate) warning messages emitted about taking an address of a packed member. Removing this guard causes the Clang test Sema/address-unaligned.c to fail.

clang/lib/Sema/SemaExpr.cpp
```diff
@@ -14177,8 +14177,14 @@ QualType Sema::ValidateBoundsExprArgument(Expr *Arg) {
   // we do see them.
   Qualifiers ArgTypeQuals = ArgTypePointee.getQualifiers();
   ArgTypeQuals.removeCVRQualifiers();
-  assert(!ArgTypeQuals.hasQualifiers() &&
+
+  if (ArgTypeQuals.hasQualifiers()) {
+    if (!DisableSubstitionDiagnostics)
```
kkjeer: When this method is called from PruneTemporaryBindings (as a result of calls to CheckUnaryOperator), there should be no assertion. Removing this change causes the Clang test Sema/address-unaligned.c to fail.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2209,17 +2218,67 @@ namespace {
         DumpExpression(llvm::outs(), E);
     }
 
-    void VisitUnaryOperator(UnaryOperator *E, CheckedScopeSpecifier CSS) {
-      if (E->getOpcode() == UO_AddrOf)
-        S.CheckAddressTakenMembers(E);
+    BoundsExpr *CheckUnaryOperator(UnaryOperator *E, CheckedScopeSpecifier CSS, SideEffects SE) {
```
dtarditi: I think it would be helpful to add a comment of the form: "If e is an rvalue, CheckUnaryOperator returns the bounds for the value produced by e.  If e is an lvalue expression, it returns BoundsUnknown."

PR #740, opened 2019-12-12
Binary operator: combine bounds inference and checking

kkjeer:
#### Overview
This pull request converts the VisitBinaryOperator method (which only performed bounds checking) to the CheckBinaryOperator method (which performs bounds checking and returns rvalue bounds).

#### Preventing an unwanted assert
Since CheckBinaryOperator is called on all binary operators that are visited by TraverseStmt (not just binary operators for which bounds need to be inferred), this pull request removes an assert in CallExprBounds to prevent test failures.

#### Adding the Facts argument
In order for RValueBounds to call CheckBinaryOperator, RValueBounds (and its callers) now take a Facts argument. This will also be necessary for future Check* methods (CheckCastExpr, CheckCallExpr, CheckVarDecl, etc.).

#### Testing


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3224,10 +3262,12 @@ namespace {
       }
       else {
         // Get the function prototype, where the abstract function return
-        // bounds are kept. The callee is always a function pointer.
+        // bounds are kept.  The callee (if it exists) 
+        // is always a function pointer.
         const PointerType *PtrTy =
           CE->getCallee()->getType()->getAs<PointerType>();
-        assert(PtrTy != nullptr);
+        if (PtrTy == nullptr)
```
kkjeer: Leaving this assert in causes the Clang Parser/block-pointer-decl.c test to fail.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1957,58 +1961,166 @@ namespace {
...
+                   LHSType->isIntegerType() ||
+                   IsBoundsSafeInterfaceAssignment(LHSType, RHS)) {
+            // Check that the value being assigned has bounds if the
+            // target of the LHS lvalue has bounds.
+            LHSTargetBounds = InferLValueTargetBounds(LHS, CSS);
+            if (!LHSTargetBounds->isUnknown()) {
+              if (E->isCompoundAssignmentOp())
+                RightBounds = InferRValueBounds(E, CSS, Facts);
+              else
+                RightBounds = InferRValueBounds(RHS, CSS, Facts);
```
dtarditi: InferRValueBounds calls RValueBounds and then does an additional check.    Instead of calling InferRValueBounds, could you use the previously computed RValueBounds of RHS and do the check here?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1957,58 +1961,166 @@ namespace {
...
+          }
+          else if (LHSType->isCheckedPointerType() ||
+                   LHSType->isIntegerType() ||
+                   IsBoundsSafeInterfaceAssignment(LHSType, RHS)) {
+            // Check that the value being assigned has bounds if the
+            // target of the LHS lvalue has bounds.
+            LHSTargetBounds = InferLValueTargetBounds(LHS, CSS);
+            if (!LHSTargetBounds->isUnknown()) {
+              if (E->isCompoundAssignmentOp())
+                RightBounds = InferRValueBounds(E, CSS, Facts);
```
dtarditi: I am concerned that how we avoid infinitely looping is not obvious here.  InferRValueBounds(E, …) calls RValueBounds(E, …),    For this case, RValueBounds in turn directly calls CheckBinaryOperator(E, …), with the side-effects flag set to false.  I suggest you short-circuit the whole thing, and make the recursive call directly here, adding the additional checking done by InferRValueBounds.

PR #744, opened 2019-12-20
Handle bounds widening for array subscripts

mgrang:


mgrang:
Changes in the latest patch set:
- `IgnoreCasts` now invokes `IgnoreValuePreservingOperations`.
- `IgnoreCasts` also explicitly checks for `LValueToRValue` casts which signify pointer derefs.
- Got rid of `ContainsPointerDeref`, `ContainsArraySubscript` and `IsPointerDeref` functions making the logic simpler.
- Replaced the use of the `auto` keyword with explicit type names in non-obvious contexts to make code more readable.
- Added tests for `_Nt_checked` arrays (both as parameters and local variables).
- Added tests for parenthesized pointer derefs and array subscripts.

clang/test/CheckedC/inferred-bounds/widened-bounds.c
```diff
@@ -52,3 +52,21 @@ void f3() {
 // CHECK: [B1]
 // CHECK-NOT: upper_bound(p) = 1
 }
+
```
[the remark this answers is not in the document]
dtarditi: Could you add tests of widening by more than 1 element?   
```
if (p[0]) {
  if (p[1]) {
    ... // bounds should be widened to p + 2 here.
```
I know that currently the compiler will issue an error for p[2] because it'll think the access is out-of-bounds.   You can use the `-verify` command-line option to clang to handle this.  It'll let you check that the dataflow analysis produces the right result, even if the compiler still produces an error later.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -136,128 +133,174 @@ bool BoundsAnalysis::AreDeclaredBoundsZero(const Expr *E, const Expr *V) {
...
+
+  // Handle if conditions of the form "if (*e1 && *e2)".
+  if (const auto *BO = dyn_cast<const BinaryOperator>(E)) {
+    if (BO->getOpcode() == BO_LAnd) {
+      FillGenSet(BO->getLHS(), EB, SuccEB);
+      FillGenSet(BO->getRHS(), EB, SuccEB);
+    }
+  }
+
+  E = IgnoreCasts(E);
```
dtarditi: I think there should be an explicit check for an lvalue-to-rvalue cast here, instead of doing it as part of IgnoreCasts.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -136,128 +133,174 @@ bool BoundsAnalysis::AreDeclaredBoundsZero(const Expr *E, const Expr *V) {
...
+                                          ElevatedCFGBlock *EB,
+                                          ElevatedCFGBlock *SuccEB) {
+
+  // An array access can be written A[4] or 4[A] (both are equivalent).
+  // getBase() and getIdx() always present the normalized view: A[4].
+  // In this case getBase() returns "A" and getIdx() returns "4".
+  const Expr *Base = IgnoreCasts(AE->getBase());
+  const Expr *Index = IgnoreCasts(AE->getIdx());
+
+  if (const auto *D = dyn_cast<DeclRefExpr>(Base)) {
```
dtarditi: I think you should check for an lvalue-to-rvalue cast of a decl ref.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -136,128 +133,174 @@ bool BoundsAnalysis::AreDeclaredBoundsZero(const Expr *E, const Expr *V) {
...
 
-  // Check if the edge condition contains a pointer deref.
-  if (!ContainsPointerDeref(E))
-    return;
+  // TODO: Handle accesses of the form:
+  // "if (*(p + i) && *(p + j) && *(p + k))"
 
-  E = IgnoreCasts(E);
+  // For conditions of the form "if (*p)".
+  if (const auto *D = dyn_cast<DeclRefExpr>(E)) {
```
dtarditi: I think you need to check for the pattern:
```
   implicit lvalue-to-rvalue cast 
      declref
```
instead of relying on IgnoreCasts to strip out the lvalue-to-rvalue cast.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -136,128 +133,174 @@ bool BoundsAnalysis::AreDeclaredBoundsZero(const Expr *E, const Expr *V) {
...
-    if (const auto *D = dyn_cast<DeclRefExpr>(Exp)) {
-      // TODO: Remove this check. Currently, for the first version of this
-      // algorithm, we are enabling bounds widening only when the declared
-      // bounds are bounds(p, p) or count(0). We need to generalize this to
-      // widen bounds for dereferences involving constant offsets from the
-      // declared upper bound of a variable.
-      if (!AreDeclaredBoundsZero(UO->getBoundsExpr(), D))
-        return;
+    // Handle *(p + i).
+    if (isa<DeclRefExpr>(LHS) && isa<IntegerLiteral>(RHS)) {
```
dtarditi: I think you should check for an lvalue-to-rvalue cast of a DeclRefExpr.

mgrang:
PR updated.
1. Fixed unit tests
2. Renamed method IsPtrDerefOrArraySubscript to IsDeclOperand.
3. Renamed method GetDeclRefExpr to GetDeclOperand.

mgrang:
ADO LNT Testing Passed. https://dev.azure.com/msresearch/CheckedC/_build/results?buildId=401520

PR #746, opened 2020-01-01
Handle non-zero length nt_array_ptrs

mgrang:
We can do bounds-widening for nt_array_ptrs variables whose upper bound is not just the variables itself (that is, non-zero length array_ptrs).
Consider,
nt_array_ptr<T> p : bounds(e1, e2);
where e1 and e2 are expressions and e2 is of the form (e + k)
To widen the bounds, we look for derefs of the form "*(e2 + i)".

Example:
`nt_array_ptr<T> p : bounds(p, p + i);`

```
if (*(p + i)) // widen by 1
  if (*(p + i + 1)) // widen by 2
    if (*(p + i + 2)) // widen by 3
```

```
if (*(p + i)) // widen by 1
  if (*(p + i + 2)) // no widening
```

```
if (*p) // no widening
  if (*(p+ 1)) // no widening
    if (*(p + 2)) // no widening
```


mgrang:
This is an initial WIP implementation. This needs to be cleaned up to handle more cases. I am pushing this PR in case someone wants to review it early on in the implementation phase.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -119,33 +116,259 @@ void BoundsAnalysis::CollectBoundsVars(const Expr *E, DeclSetTy &BoundsVars) {
...
+  // Case 3: LHS and RHS are both DeclRefExpr's. This means there is no
+  // IntegerLiteral in the expr. In this case, we return the incoming
+  // BinaryOperator expr with a nullptr for the RHS.
+  // p + q ==> return (p + q, nullptr)
+  if (IsDeclOperand(LHS) && IsDeclOperand(RHS))
+    return std::make_pair(BO, nullptr);
+
+  if (!isa<BinaryOperator>(LHS) &&
+      !isa<BinaryOperator>(RHS))
+    return std::make_pair(nullptr, nullptr);
```
dtarditi: Can you just return the binary operator with a nullptr for length?   This would let you get rid of the prior statement testing just for DeclOperands.

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -119,33 +116,259 @@ void BoundsAnalysis::CollectBoundsVars(const Expr *E, DeclSetTy &BoundsVars) {
...
+  const llvm::APInt Zero(Ctx.getTypeSize(Ctx.IntTy), 0);
+
+  const llvm::APInt DerefOffsetVal = !DerefOffset ? Zero :
+    dyn_cast<IntegerLiteral>(DerefOffset)->getValue();
+
+  const llvm::APInt UpperOffsetVal = !UpperOffset ? Zero :
+    dyn_cast<IntegerLiteral>(UpperOffset)->getValue();
+
+  // We cannot widen the bounds if the offset in the deref expr is less than
+  // the offset in the upper bounds expr. For example:
```
dtarditi: Should we only widen the bounds only exactly when the offset for the deref expr equals the offset in the upper bounds expr?

clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -119,33 +116,259 @@ void BoundsAnalysis::CollectBoundsVars(const Expr *E, DeclSetTy &BoundsVars) {
...
+      return std::make_pair(BE, RHS);
+
+    // Expr is Case 5. ie: The BinaryOperator expr has an IntegerLiteral on
+    // the RHS.
+    // (p + j) + i ==> return (p, j + i)
+    Expr::EvalResult R1, R2;
+    RHS->EvaluateAsInt(R1, Ctx);
+    BinOpRHS->EvaluateAsInt(R2, Ctx);
+    auto *I = new (Ctx)
+                IntegerLiteral(Ctx, R1.Val.getInt() + R2.Val.getInt(),
```
dtarditi: You need to check for integer overflow or wraparound here.   I think it is fine to open an issue tracking this and place a TODO in the code.

mgrang:
Will add unit tests soon.

clang/lib/Sema/BoundsAnalysis.cpp
mgrang: Will try this tomorrow.

mgrang:
I spent a lot of time trying to merge master on this branch but it led to numerous merge conflicts. Even after resolving those I see GitHub complains about conflicting files. At this time, the simplest approach would be to push my changes to a fresh branch. I have done so here: https://github.com/microsoft/checkedc-clang/pull/750.

@dtarditi Please let me know if this works. Otherwise, I can go back and spend more time resolving conflicts.

mgrang:
Spoke to David and he is fine with closing this PR in favor of https://github.com/microsoft/checkedc-clang/pull/750.

PR #747, opened 2020-01-06
Cast expression: combine bounds inference and checking

kkjeer:
Convert the VisitCastExpr method (which only performed bounds checking) to the CheckCastExpr method (which performs both bounds inference and checking).

This pull request also modifies the Sema ExprSubstitutionScope RAII class to be able to optionally disable diagnostics. This scope is used in Check* methods (e.g. CheckCastExpr) to suppress diagnostics if side effects are disabled. Some methods such as CallExprBounds can emit errors while inferring bounds. Since bounds may be inferred more than once for a given expression (since TraverseStmt is not fully refactored yet and still checks all substatements of an expression), these helper methods for inferring bounds may emit errors twice for an expression. Using the ExprSubstitutionScope class to suppress diagnostics helps eliminate unwanted duplicate error messages.

Testing:
* No changes to existing tests and no new tests. 
* Passed automated testing on Windows/Linux, with no apparent increase in test runtimes.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -2237,24 +2237,23 @@ namespace {
       if (SE == SideEffects::Enabled)
         CheckDisallowedFunctionPtrCasts(E);
 
-      // If rvalue bounds for e cannot be determined,
+      // If the rvalue bounds for e cannot be determined,
       // e may be an lvalue (or may have unknown rvalue bounds).
-      BoundsExpr *ResultBounds = CreateBoundsAlwaysUnknown();
+      BoundsExpr *ResultBounds = CreateBoundsUnknown();
 
-      // Recursively compute rvalue bounds for the subexpression,
-      // without performing bounds checking.  Since TraverseStmt still checks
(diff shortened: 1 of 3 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2250,64 +2293,77 @@ namespace {
...
-        E->setSubExprBoundsExpr(SubExprBounds);
-        if (DumpBounds)
-          DumpBoundsCastBounds(llvm::outs(), E, DeclaredBounds,
-                               NormalizedBounds, SubExprBounds);
+      // during code generation.
+      if (CK == CK_DynamicPtrBounds || CK == CK_AssumePtrBounds) {
+        CHKCBindTemporaryExpr *TempExpr = dyn_cast<CHKCBindTemporaryExpr>(SubExpr);
+        assert(TempExpr);
+
+        // These bounds will be computed and tested at runtime.  Don't
```
dtarditi: I think this should say "These bounds may be computed and tested at runtime."   Tthe original comment in the code was flawed.  The flaw is more obvious here.

PR #748, opened 2020-01-07
Call expression: combine bounds inference and checking

kkjeer:
Convert the VisitCallExpr method (which only performed bounds checking) to the CheckCallExpr method (which performs both bounds inference and checking).

Note: unlike previously implemented Check* methods (CheckUnaryOperator, CheckBinaryOperator, etc.), CheckCallExpr is not called from RValueBounds. CheckCallExpr suppresses diagnostics from CallExprBounds to prevent unwanted duplicate error messages. However, those diagnostics from CallExprBounds may be wanted when called from RValueBounds. Once TraverseStmt has finished being refactored, it will replace RValueBounds. In order to maintain the current compiler behavior, CheckCallExpr is not called from RValueBounds.

Testing:
* Passed automated testing on Windows/Linux with no apparent increase in testing time

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -3241,6 +3241,12 @@ namespace {
         case Expr::CompoundAssignOperatorClass:
           return CheckBinaryOperator(cast<BinaryOperator>(E), CSS, Facts, SE);
         case Expr::CallExprClass: {
+          // Do not call CheckCallExpr here.  Since CheckCallExpr suppresses
+          // diagnostics emitted as part of CallExprBounds (to reduce unwanted
+          // duplicate diagnostics), calling CheckCallExpr here can result in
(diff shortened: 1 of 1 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2113,119 +2114,137 @@ namespace {
...
-          S.ConcretizeFromFunctionTypeWithArgs(
-            const_cast<BoundsExpr *>(ParamBounds),
-            ArgExprs,
-            Sema::NonModifyingContext::NMC_Function_Parameter);
+      if (SE == SideEffects::Disabled)
+        return ResultBounds;
 
-        if (!SubstParamBounds)
-          continue;
+      // Perform bounds checking, if enabled.
```
dtarditi: I think this should say "Perform checking of bounds declarations, if enabled."

PR #751, opened 2020-01-08
Update directory location in automation for code signing.

dtarditi:
When we moved to a monorepo, we adjusted the location of automation scripts to be clang\automation instead of automation.  We missed updating a path for sync'ing a repo that contain a scripts for code signing.  This worked on existing ADO pipelines because the old copy of the repo was still there.  The sync'ing did not have the intended effect, but the operations did not fail.  When I cloned an ADO pipeline for some experimentation, the sync'ing failed because there is no old copy of the repo.

### DIFF
--- a/clang/automation/Windows/setup-files.bat
+++ b/clang/automation/Windows/setup-files.bat
@@ -66,7 +66,7 @@ if ERRORLEVEL 1 (goto cmdfailed)
 
 rem Set up sources for scripts for signing installer
 if "%SIGN_INSTALLER%" NEQ "No" (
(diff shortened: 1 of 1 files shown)


mgrang:
Did this pass the Windows ADO build?

dtarditi:
A Windows ADO build is in progress and has passed the point of failure.  It will take over an hour to complete it, though.

mgrang:
> A Windows ADO build is in progress and has passed the point of failure. It will take over an hour to complete it, though.

Alright. LGTM!

PR #755, opened 2020-01-13
Traverse children from CheckCastExpr

kkjeer:
Traverse the subexpression in CheckCastExpr rather than TraverseStmt to help prevent duplicate RValueBounds calls from TraverseStmt. Clean up diagnostic suppression in case where side effects are removed by moving the ExprSubstitutionScope to TraverseStmt and RValueBounds rather than individual Check* methods.

Testing:

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -1953,16 +1953,19 @@ namespace {
           ResultBounds = CheckReturnStmt(RS, CSS, SE);
           break;
         }
+        // Since TraverseStmt still checks all children of temporary binding
+        // expressions, this case should perform bounds inference only,
+        // with no side effects.  In future refactoring stages, there will be
+        // a CheckTemporaryBinding method that performs bounds inference with
+        // side effects in a bottom-up manner.
         case Stmt::CHKCBindTemporaryExprClass: {
-          // Suppress diagnostics.  Any intended diagnostics from bounds
-          // inference will be emitted from other calls to RValueBounds.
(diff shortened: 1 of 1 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2323,29 +2319,46 @@ namespace {
...
           E->getType()->getPointeeOrArrayElementType()->isNtCheckedArrayType();
+      bool PreviousIncludeNullTerminator = IncludeNullTerminator;
+      IncludeNullTerminator = IncludeNullTerm;
 
-      CastKind CK = E->getCastKind();
-      Expr *SubExpr = E->getSubExpr();
+      // If the lvalue target bounds and lvalue bounds for the
+      // subexpression are needed, they must be computed before
+      // performing potential side effects on the subexpression.
+      bool IsResultCastBounds = (E->getStmtClass() == Stmt::ImplicitCastExprClass ||
```
dtarditi: Apparently I didn't save my comment when writing my pull request feedback, and it got lost..  I believe LValueToRvalue casts and ArrayToPointerDecay casts are always implicit casts.  That means that you can structure the code to do an if-test and then move the two if-statements below within it.
```
BoundsExpr *SubExprTarget = nullptr;
BoundsExpr *SubExprLValueBounds = nullptr;
if (E->getStmtClass() == Stmt::ImplicitCastExprClass &&
   !E->getType()->isCheckedPointerPtrType()) {
  if (CK == CK_LValueToRvalue) {
       …
  }
  if (CK == CK_ArrayToPointerDecay) {
   …
  }
}
```

PR #758, opened 2020-01-14
Traverse children in CheckCallExpr

kkjeer:
Traverse the children (callee and arguments) from CheckCallExpr rather than TraverseStmt.

Testing:

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -1910,10 +1910,13 @@ namespace {
                                                   CSS, Facts, SE);
           return AdjustRValueBounds(S, Bounds);
         }
-        case Expr::CallExprClass:
-          ResultBounds = CheckCallExpr(cast<CallExpr>(S),
-                                       CSS, Facts, SE);
-          break;
+        // CheckCallExpr traverses its callee and arguments,
+        // so there is no need to traverse its children below.
+        case Expr::CallExprClass: {
+          BoundsExpr *Bounds = CheckCallExpr(cast<CallExpr>(S),
+                                             CSS, Facts, SE);
+          return AdjustRValueBounds(S, Bounds);
+        }
         case Expr::MemberExprClass:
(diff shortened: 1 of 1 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2222,17 +2224,41 @@ namespace {
...
+        TraverseChildren(E, CSS, Facts, SE);
         return ResultBounds;
-      if (!FuncProtoTy->hasParamAnnots())
+      }
+      if (!FuncProtoTy->hasParamAnnots()) {
+        TraverseChildren(E, CSS, Facts, SE);
+        return ResultBounds;
+      }
+      if (SE == SideEffects::Disabled) {
+        TraverseChildren(E, CSS, Facts, SE);
```
dtarditi: If there are no side-effects, do you still need to call TraverseChildren here?   What does it do?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2222,17 +2224,41 @@ namespace {
...
+
+      // Recursively traverse the callee.  The arguments will be
+      // traversed below.  This prevents TraverseStmt from
+      // needing to traverse the children of call expressions.
+      TraverseStmt(E->getCallee(), CSS, Facts, SE);
 
       unsigned NumParams = FuncProtoTy->getNumParams();
       unsigned NumArgs = E->getNumArgs();
       unsigned Count = (NumParams < NumArgs) ? NumParams : NumArgs;
       ArrayRef<Expr *> ArgExprs = llvm::makeArrayRef(const_cast<Expr**>(E->getArgs()), E->getNumArgs());
```
dtarditi: You need to handle the case where the number of arguments > the number of parameters at the function call.  The additional arguments still need to be traversed.   This can happen at varargs calls, for example.   The additional arguments could require runtime bounds checks within them.  They could also have casts that need to be statically checked (for example, to _Ptr type).

PR #759, opened 2020-01-14
Traverse children in CheckVarDecl

kkjeer:
If a variable declaration has an initializer expression, traverse it in CheckVarDecl rather than traversing the children of DeclStmts in TraverseStmt.

Testing:
* No new tests or change to existing tests
* Passed manual testing on Windows
* Passed automated testing on Windows/Linux

PR #763, opened 2020-01-16
LValueBounds: remove duplicate TraverseStmt calls

kkjeer:
This PR removes duplicate calls made to TraverseStmt that can result from calling LValueBounds. LValueBounds can call TraverseStmt to get the rvalue bounds for the subexpression `e` of a deference `*e`, or the base of an array subscript `e1[e2]`. The callers of LValueBounds would then call TraverseStmt on their subexpressions, which could result in duplicate TraverseStmt calls. LValueBounds and CheckUnaryOperator now use out parameters to avoid recomputing rvalue bounds for expressions.

Future work: after this PR and #759 are merged, the next step is to remove the SideEffects parameter that is currently used to prevent duplicate side effects, since there should no longer be any duplicate calls to infer/check the bounds for an expression.

Testing:


kkjeer:
These changes are included in #770

PR #765, opened 2020-01-16
Replace ExprPairTy to ExprIntPairTy [NFC]

mgrang:
We use ExprPairTy as a return type when we split an Expr into a base and an
offset. We return the offset as an Expr and then extract the integer constant
from it. But while computing the offset we already have the integer constant
available. We can make use of this instead of re-computing. In case, the offset
is not an integer constant we return a zero integer.


mgrang:
This PR will simplify my next PR for overflow checking.

PR #766, opened 2020-01-16
Add compile-time integer overflow checking for bounds widening

mgrang:
In order to make an expression e uniform we split the expr into a "base" and an "offset". The base contains all DeclRefExprs of e and the Offset contains all IntegerLiterals of e. In order to do the split, we need to re-associate the integers.

For example: We convert an expression like (p + i) + j into p + (i + j), where p is  DeclRefExpr and i and j are integers. In this case, (i + j) may potentially overflow. We implement compile time checking to check for integer overflow.


mgrang:
This is a follow up to https://github.com/microsoft/checkedc-clang/pull/765

PR #769, opened 2020-01-17
Use llvm::APSInt::compareValues instead of directly comparing integer constants

mgrang:
llvm::APSInt::compareValues correctly account for signedness/unsignedness of integer when comparing them. So it is good to use this method instead of directly comparing integer constants.


mgrang:
Directly comparing integer constant caused an LNT breakage in `MultiSource/Benchmarks/Ptrdist/yacr2/vcg.c` where BinOpRHS was an APInt whereas Zero was an APSInt. llvm::APSINT::compareValues correctly handles the signed/unsigned integers while comparing them.

mgrang:
Thanks @dtarditi. Will merge once the Linux X64 LNT tests pass.

mgrang:
Linux X64 LNT testing passed : https://dev.azure.com/msresearch/CheckedC/_build/results?buildId=409782

PR #775, opened 2020-01-21
Pass FunctionDecl as a parameter to BoundsAnalysis functions [2/n]

mgrang:
Currently, FunctionDecl (FD) is a member of BoundsAnalysis class. As a result,
we need to pass FD while instantiating a BoundsAnalysis object. We plan to add
a BoundsAnalysis object to CheckBoundsDeclaration class but we do not have a
FunctionDecl in the CheckBoundsDeclaration constructor. So we would not be able
to add the BoundsAnalysis object as long as FD is a member of BoundsAnalysis
class. So we need to decouple FD from BoundsAnalysis. We would pass FD only to
methods which need FD.

This is the second PR in a series of PRs to integrate the results of bounds-widening with bounds inference/checking.

### DIFF
--- a/clang/include/clang/Sema/BoundsAnalysis.h
+++ b/clang/include/clang/Sema/BoundsAnalysis.h
@@ -95,7 +95,6 @@ namespace clang {
     Sema &S;
     CFG *Cfg;
     ASTContext &Ctx;
-    FunctionDecl *FD;
     Lexicographic Lex;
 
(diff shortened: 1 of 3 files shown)


mgrang:
Linux X64 LNT Testing Passed: https://dev.azure.com/msresearch/CheckedC/_build/results?buildId=412047

PR #777, opened 2020-01-22
Add BoundsAnalysis object to CheckBoundsDeclaration class [3/n]

mgrang:
Adding a BoundsAnalysis object to ChecBoundsDeclaration class would allow us to
easily invoke BoundsAnalysis methods without having to pass around the
BoundsAnalysis object.

This is the third PR in a series of PRs to integrate the results of
bounds-widening with bounds inference/checking.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -487,6 +487,12 @@ namespace {
                               // function, if any.
     ASTContext &Context;
     std::pair<ComparisonSet, ComparisonSet> &Facts;
+
+    // Having a BoundsAnalysis object here allows us to easily invoke methods
+    // for bounds-widening and get back the bounds-widening info needed for
+    // bounds inference/checking.
+    BoundsAnalysis WidenBounds;
+
     // When this flag is set to true, include the null terminator in the
     // bounds of a null-terminated array.  This is used when calculating
     // physical sizes during casts to pointers to null-terminated arrays.
@@ -1727,6 +1733,7 @@ namespace {
       ReturnBounds(ReturnBounds),
       Context(SemaRef.Context),
       Facts(Facts),
+      WidenBounds(BoundsAnalysis(SemaRef, Cfg)),
       IncludeNullTerminator(false) {}
 
     CheckBoundsDeclarations(Sema &SemaRef, std::pair<ComparisonSet, ComparisonSet> &Facts) : S(SemaRef),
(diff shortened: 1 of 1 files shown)


mgrang:
On second thoughts, I think we don't really require a BoundsAnalysis object _inside_ CheckBoundsDeclaration. As long as we have access to SemaRef and Cfg objects we should be able to instantiate BoundsAnalysis. We just need to have a <Stmt, <Variable, WidenedBounds>> map inside CheckBoundsDeclaration in order to avoid passing the map around. Then given a Stmt we should be able to get the widened bounds for all nt_array_ptrs used in the Stmt.

Thoughts?

mgrang:
Closing this PR so you won't need to spend time reviewing it.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3915,10 +3925,10 @@ void Sema::CheckFunctionBodyBoundsDecls(FunctionDecl *FD, Stmt *Body) {
   }
 
   if (Cfg != nullptr) {
-    BoundsAnalysis Collector(*this, Cfg.get());
-    Collector.WidenBounds(FD);
+    BoundsAnalysis WB = Checker.getBoundsAnalysis();
+    WB.WidenBounds(FD);
```
kkjeer: Can this be done within Collector.TraverseCFG?

mgrang:
Had a discussion with @kkjeer and we came to the conclusion that in fact it makes more sense to have a BoundsAnalysis object inside CheckBoundsDeclaration class. This would make querying the BoundsAnalysis methods easier and would avoid having to pass around the widened bounds map and the kill set for the block.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -487,6 +487,12 @@ namespace {
                               // function, if any.
     ASTContext &Context;
     std::pair<ComparisonSet, ComparisonSet> &Facts;
+
+    // Having a BoundsAnalysis object here allows us to easily invoke methods
+    // for bounds-widening and get back the bounds-widening info needed for
+    // bounds inference/checking.
+    BoundsAnalysis WidenBounds;
```
kkjeer: Nit: I'd suggest a noun name for this like `BoundsAnalyzer` or `BoundsWidener`.

clang/lib/Sema/SemaBounds.cpp
mgrang: I guess I would have to pass FD to TraverseCFG. In any case, we would eventually get rid of this piece of code. WidenBounds would be invoked only where we actually use the widen bounds info. So I would keep this here for now.

kkjeer: I thought WidenBounds would be invoked once per function declaration, either before Collector.TraverseCFG or at the beginning of Collector.TraverseCFG. I think it needs to be called before iterating over the CFG blocks and calling TraverseStmt on each block so that the bounds map is populated before CheckBoundsDeclarations needs the widened bounds information.

PR #778, opened 2020-01-22
Change Kill set to be a map from a Stmt to a set of vars whose bounds are killed by the Stmt [4/n]

mgrang:
Currently, the Kill set is a set of variables killed in a block. But we have no
info on "where" exactly a variable is killed inside a block. As a result, for a
given Stmt we cannot determine if the bounds of a variable are valid after the
Stmt or not. So we change the Kill set to be a mapping between the Stmt and the
variables killed by that Stmt inside a particular block.

This info would be used by bounds inference/checking to check the widened
bounds after a given Stmt.

### DIFF
--- a/clang/include/clang/Sema/BoundsAnalysis.h
+++ b/clang/include/clang/Sema/BoundsAnalysis.h
@@ -70,9 +70,7 @@ namespace clang {
   // For each edge B1->B2, EdgeBoundsTy denotes the Gen and Out sets.
   using EdgeBoundsTy = llvm::DenseMap<const CFGBlock *, BoundsMapTy>;
 
-  // For each block B, DeclSetTy denotes the Kill set. A VarDecl V is killed if:
-  // 1. V is assigned to in the block, or
-  // 2. any variable used in the bounds expr of V is assigned to in the block.
+  // DeclSetTy denotes a set of VarDecls.
   using DeclSetTy = llvm::DenseSet<const VarDecl *>;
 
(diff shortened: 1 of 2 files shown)


clang/include/clang/Sema/BoundsAnalysis.h
```diff
@@ -266,6 +263,17 @@ namespace clang {
     // @param[in] FD is the current function.
     void CollectNtPtrsInScope(FunctionDecl *FD);
 
+    // Get the Kill set for the current block. The Kill set is a mapping of
+    // Stmts to variables whose bounds killed by each Stmt in the block.
+    // @param[in] B is the current CFGBlock.
+    // return A mapping of Stmts to variables whose bounds killed by the Stmt.
+    StmtDeclSetTy GetKillSet(const clang::CFGBlock *B);
```
kkjeer: Is GetKillSet intended to be called from CheckBoundsDeclarations? It looks like it's a private method.

clang/include/clang/Sema/BoundsAnalysis.h
```diff
@@ -266,6 +263,17 @@ namespace clang {
     // @param[in] FD is the current function.
     void CollectNtPtrsInScope(FunctionDecl *FD);
 
+    // Get the Kill set for the current block. The Kill set is a mapping of
+    // Stmts to variables whose bounds killed by each Stmt in the block.
```
kkjeer: "bounds killed" -> "bounds are killed"

PR #782, opened 2020-01-23
Create a baseline for release_90 sources from community

mgrang:
This is the baseline branch for release_90 from community. The custom Checked C changes would be applied on top of this branch.


mgrang:
We have created the baseline branch baseline-llvm-90-rc1. We do not necessarily need to have a PR for this. So we can close this PR.

PR #783, opened 2020-01-25
Check lvalues and rvalues separately

kkjeer:
This PR separates the bounds inference and checking of lvalue expressions and rvalue expressions into separate methods.

The Check method behaves as follows:
* For an lvalue expression: check it (and its children) by calling CheckLValue, and return unknown bounds.
* For an rvalue expression: check its children, perform any side effects such as bounds declaration checking, and return its rvalue bounds.

The CheckLValue method behaves as follows (for an lvalue expression):
* Check its children by calling Check
* Infer its target bounds via an out parameter
* Perform any side effects such as adding a bounds check
* Return its lvalue bounds

These two Check and CheckLValue methods work together for the invariant that each expression in a CFG is traversed exactly once.

Testing:


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3799,7 +3784,8 @@ void Sema::CheckFunctionBodyBoundsDecls(FunctionDecl *FD, Stmt *Body) {
     // __finally or may encounter a malformed AST.  Fall back on to non-flow 
     // based analysis.  The CSS parameter is ignored because the checked
     // scope information is obtained from Body, which is a compound statement.
-    Checker.TraverseStmt(Body, CheckedScopeSpecifier::CSS_Unchecked);
+    Checker.SetCheckedScopeSpecifier(CheckedScopeSpecifier::CSS_Unchecked);
```
dtarditi: I don't believe this is the proper use of the RAII ("resource allocation is acquisition") idiom.  I think you want to create an object that on construction stores the existing CSS value and sets the new CSS value, and on destruction restores the old CSS value.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1875,55 +1873,93 @@ namespace {
...
+  // 2. As the operand to a postfix or prefix incrementation operators (which
+  //    implicitly do assignment).
+  // 3. As the operand of the address-of (&) operator.
+  // 4. If a member access operation e1.f denotes an lvalue, e1 denotes an
+  //    lvalue.
+  // 5. In clang IR, as an operand to an LValueToRValue cast operation.
+  // Otherwise an expression denotes an rvalue.
+
+  public:
+    // If e is an rvalue, Check returns the bounds for the value produced by e.
```
dtarditi: I think this comment still needs to be updated.  Checked checks e and its children, even if it is an rvalue.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1937,47 +1973,127 @@ namespace {
...
     // Traverse a top-level variable declaration.  If there is an
     // initializer, it will be traversed in CheckVarDecl.
     void TraverseTopLevelVarDecl(VarDecl *VD, CheckedScopeSpecifier CSS) {
+      SetCheckedScopeSpecifier(CSS);
       ResetFacts();
-      CheckVarDecl(VD, CSS);
+      CheckVarDecl(VD);
+    }
+
+    void SetCheckedScopeSpecifier(CheckedScopeSpecifier Scope) {
```
dtarditi: I think you need a wrapper object type to properly implement the RAII pattern.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2600,29 +2646,230 @@ namespace {
...
+    // If e is an lvalue, CheckUnaryLValue returns the
+    // lvalue and target bounds of e.
+    // If e is an rvalue, CheckUnaryOperator should be called instead.
+    BoundsExpr *CheckUnaryLValue(UnaryOperator *E,
+                                 BoundsExpr *&OutTargetBounds) {
+      BoundsExpr *SubExprBounds = Check(E->getSubExpr());
+
+      if (E->getOpcode() == UnaryOperatorKind::UO_Deref) {
+        // Currently, we don't know the target bounds of a pointer stored in a
+        // pointer dereference, unless it is a _Ptr type (handled
```
dtarditi: It isn't clear to me that _Ptr is handled earlier.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2600,29 +2646,230 @@ namespace {
...
+
+      OutTargetBounds = CreateBoundsInferenceError();
+      return CreateBoundsInferenceError();
+    }
+
+    // CheckArraySubscriptExpr returns the lvalue and target bounds of e.
+    // e is an lvalue.
+    BoundsExpr *CheckArraySubscriptExpr(ArraySubscriptExpr *E,
+                                        BoundsExpr *&OutTargetBounds) {
+      // Currently, we don't know the target bounds of a pointer returned
```
dtarditi: It isn't clear that _Ptr is handled earlier.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2600,29 +2646,230 @@ namespace {
...
+      if (NeedsBoundsCheck && DumpBounds)
+        DumpExpression(llvm::outs(), E);
+      return Bounds;
+    }
+
+    // If e is an lvalue, CheckCastLValue returns the
+    // lvalue and target bounds of e.
+    // If e is an rvalue, CheckCastExpr should be called instead.
+    BoundsExpr *CheckCastLValue(CastExpr *E, BoundsExpr *&OutTargetBounds) {
+      // An LValueBitCast adjusts the type of the lvalue, but
```
dtarditi: I think this should be reworded as:
```
// An LValueBitCast adjusts the type of the lvalue.  The bounds are not changed, except
// that their relative alignment may change (the bounds may only cover a partial object). 
// TODO: When we add relative alignment support to the compiler, adjust the relative alignment.

File: clang/lib/Sema/SemaBounds.cpp (hunk 4 above)
Comment (kkjeer) [reply clang/lib/Sema/SemaBounds.cpp]: This was left over from the former top-down approach, where _Ptr would be handled earlier (e.g. by an LValueToRValue cast in VisitCastExpr). I've updated the comment and behavior here to return type-based bounds for both _Ptr and _Nt_array_ptr types.

File: clang/lib/Sema/SemaBounds.cpp (hunk 5 above)
Comment (kkjeer) [reply clang/lib/Sema/SemaBounds.cpp]: This was also left over from the top-down approach and has been updated.

File: clang/lib/Sema/SemaBounds.cpp [hunk 7]
```diff
@@ -1858,7 +1873,7 @@ namespace {
            if (NestedElements.find(S) != NestedElements.end())
              continue;
 
-           CheckedScopeSpecifier CSS = CheckedScopeSpecifier::CSS_Unchecked;
+           CSS = CheckedScopeSpecifier::CSS_Unchecked;
```
dtarditi: I think it is confusing to directly set the CSS member here.  I'd recommend using the RAII pattern that is now being used elsewhere in the file.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1875,55 +1890,95 @@ namespace {
             S->dump(llvm::outs());
             llvm::outs().flush();
 #endif
-            TraverseStmt(S, CSS);
+            Check(S);
```
dtarditi: I think you could introduce an RAII object for setting the CSS member here.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1875,55 +1890,95 @@ namespace {
...
+          break;
         case Expr::BinaryOperatorClass:
         case Expr::CompoundAssignOperatorClass:
-          ResultBounds = CheckBinaryOperator(cast<BinaryOperator>(S), CSS);
-          return AdjustRValueBounds(S, ResultBounds);
+          ResultBounds = CheckBinaryOperator(cast<BinaryOperator>(S));
+          break;
         case Stmt::CompoundStmtClass: {
           CompoundStmt *CS = cast<CompoundStmt>(S);
           CSS = CS->getCheckedSpecifier();
```
dtarditi: I think you need to the RAII pattern here too.

clang/lib/Sema/SemaBounds.cpp
kkjeer: Based on our discussion, I made CSS a parameter passed through CheckBoundsDeclarations methods again.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1875,55 +1875,95 @@ namespace {
...
+        case Expr::BoundsCastExprClass:
+          ResultBounds = CheckCastExpr(cast<CastExpr>(S), CSS);
+          break;
         case Expr::BinaryOperatorClass:
         case Expr::CompoundAssignOperatorClass:
           ResultBounds = CheckBinaryOperator(cast<BinaryOperator>(S), CSS);
-          return AdjustRValueBounds(S, ResultBounds);
+          break;
         case Stmt::CompoundStmtClass: {
           CompoundStmt *CS = cast<CompoundStmt>(S);
```
kkjeer: Following up from our discussion about checking the children of CompoundStmts:

Constructing the CFG "flattens" children of compound statements. For example, given this test code:
```
void compound_test(array_ptr<int> a : count(3), nt_array_ptr<char> c : count(4)) {
  checked {
    array_ptr<int> arr : count(3) = a;
    unchecked {
      arr = b; // error: declared bounds for arr are invalid after assignment
    }
  }
  unchecked {
    nt_array_ptr<char> buf : count(4) = c;
    buf = "ab"; // error: declared bounds for buf are invalid after assignment
  }
}
```
The following AST will be constructed as the body of the CFG:
```
CompoundStmt 0x1fac9ea5408
|-CompoundStmt 0x1fac9ea5110 _Checked checking-state bounds-and-types
| |-DeclStmt 0x1fac9ea5038
| | `-VarDecl 0x1fac9ea4f20  used arr '_Array_ptr<int>' cinit
| |   |-CountBoundsExpr 0x1fac9ea4fb0 'NULL TYPE' Element
| |   | `-IntegerLiteral 0x1fac9ea4f88 'int' 3
| |   `-ImplicitCastExpr 0x1fac9ea5008 '_Array_ptr<int>' <LValueToRValue>
| |     `-DeclRefExpr 0x1fac9ea4fe0 '_Array_ptr<int>' lvalue ParmVar 0x1fac9ea4ab8 'a' '_Array_ptr<int>'
| `-CompoundStmt 0x1fac9ea50f0 _Unchecked
|   `-BinaryOperator 0x1fac9ea50d0 '_Array_ptr<int>' '='
|     |-DeclRefExpr 0x1fac9ea5050 '_Array_ptr<int>' lvalue Var 0x1fac9ea4f20 'arr' '_Array_ptr<int>'
|     `-ImplicitCastExpr 0x1fac9ea50a0 '_Array_ptr<int>' <LValueToRValue>
|       `-DeclRefExpr 0x1fac9ea5078 '_Array_ptr<int>' lvalue ParmVar 0x1fac9ea4b58 'b' '_Array_ptr<int>'
`-CompoundStmt 0x1fac9ea5380 _Unchecked
  |-DeclStmt 0x1fac9ea5280
  | `-VarDecl 0x1fac9ea5168  used buf '_Nt_array_ptr<char>' cinit
  |   |-CountBoundsExpr 0x1fac9ea51f8 'NULL TYPE' Element
  |   | `-IntegerLiteral 0x1fac9ea51d0 'int' 4
  |   `-ImplicitCastExpr 0x1fac9ea5250 '_Nt_array_ptr<char>' <LValueToRValue>
  |     `-DeclRefExpr 0x1fac9ea5228 '_Nt_array_ptr<char>' lvalue ParmVar 0x1fac9ea4bf0 'c' '_Nt_array_ptr<char>'
  `-BinaryOperator 0x1fac9ea5360 '_Nt_array_ptr<char>' '='
    |-DeclRefExpr 0x1fac9ea5298 '_Nt_array_ptr<char>' lvalue Var 0x1fac9ea5168 'buf' '_Nt_array_ptr<char>'
    `-ImplicitCastExpr 0x1fac9ea5330 '_Nt_array_ptr<char>' <ArrayToPointerDecay> BoundsSafeInterface
      `-CHKCBindTemporaryExpr 0x1fac9ea5318 'char [3]' lvalue
        `-StringLiteral 0x1fac9ea52f8 'char [3]' lvalue "ab"
```
The following CFG will be constructed:
```
 [B2 (ENTRY)]
   Succs (1): B1

 [B1]
   1: _Array_ptr<int> arr : count(3) = a;
   2: arr = b
   3: _Nt_array_ptr<char> buf : count(4) = c;
   4: buf = "ab"
   Preds (1): B2
   Succs (1): B0

 [B0 (EXIT)]
   Preds (1): B1
```
Statements 1, 2, 3, and 4 from CFG block B1 will be traversed in `TraverseCFG`, so `Check` will be called on each of these four statements. `Check` will not be called on a CompoundStmt.

PR #789, opened 2020-02-14
Remove automation from checkedc-clang

mgrang:
The automation dir has been moved to its own repo:
https://github.com/microsoft/checkedc-automation.git


dtarditi:
Nice work!  I think some further work is still needed on the moved scripts.  See the comments for https://github.com/microsoft/checkedc-automation/pull/5.  

In addition, I think you'll need to make some updates to documentation:
- We should document where the automation scripts are because otherwise they will be hard to find. 
- We need to update the setup, build, and test instructions in checkedc-clang/clang/docs/checkedc.

mgrang:
We need to have automation in the clang directory for now because we need to run SDL analysis which only runs using msbuild. Once we have figured out a way to run analysis using ninja/cmake then we can remove automation from clang. For now I will close this PR and delete this branch.

PR #792, opened 2020-02-18
Equivalent expressions: non-assignments [2/n]

kkjeer:
This PR updates the sets UEQ and G of equivalent expressions while checking rvalue expressions that do not involve assignments. Specifically, the following methods update UEQ and G:
* CheckBinaryOperator (non-assignments and non-logical operators)
* CheckCallExpr (not adjusting UEQ to account for uses of any variables passed as arguments)
* CheckCastExpr
* CheckUnaryOperator (non-increment/decrement operators)
* CheckTemporaryBinding

Testing:
* Added tests for more kinds of rvalue expressions (not involving assignments) to equiv-exprs.c).


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2211,14 +2226,21 @@ namespace {
                                     CheckingState &State) {
       Expr *LHS = E->getLHS();
       Expr *RHS = E->getRHS();
+      ExprEqualMapTy SubExprGs;
 
-      // Infer the lvalue or rvalue bounds of the LHS.
+      // Infer the lvalue or rvalue bounds of the LHS,
```
dtarditi: For multi-line comments, please be sure to use most of the 80 character line (say 70 characters).   The clang/LLVM coding convention encourages this.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2306,6 +2328,17 @@ namespace {
...
       }
 
+      if (E->isAssignmentOp()) {
+        // TODO: update State for assignments `e1 = e2` and `e1 @= e2`.
+      } else if (BinaryOperator::isLogicalOp(Op)) {
+        // TODO: update State for logical operators `e1 && e2` and `e1 || e2`.
+      } else if (Op != BinaryOperatorKind::BO_Comma) {
+        // For comma operators `e1, e2`, State.G remains the same
+        // as the set State.G produced by checking e2.  For all other
+        // binary operators, use the sets for e1 and e2 to update State.G.
```
dtarditi: Where is the comma operator case handled?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3162,6 +3233,136 @@ namespace {
...
+      switch (E->getStmtClass()) {
+        case Expr::UnaryOperatorClass: {
+          UnaryOperator *UO = cast<UnaryOperator>(E);
+          // *e reads memory via a pointer.
+          return UO->getOpcode() == UnaryOperatorKind::UO_Deref;
+        }
+        // e1[e2] is a synonym for *(e1 + e2), which reads memory via a pointer.
+        case Expr::ArraySubscriptExprClass:
+          return true;
+        // e1.e2 and e1->e2 read memory via a pointer.
```
dtarditi: I think the proper syntax is e1.f or e1->f.   In the first case, e1.f reads memory via a pointer only of e1 reads memory via a pointer.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2512,13 +2548,38 @@ namespace {
...
       InferBounds(SubExpr, CSS, SubExprTargetBounds,
                   SubExprLValueBounds, SubExprBounds, State);
 
       IncludeNullTerminator = PreviousIncludeNullTerminator;
 
+      // Update the set State.G of expressions that produce the
+      // same value as e.
+      if (CK == CastKind::CK_ArrayToPointerDecay)
+        // State.G = { e } for lvalues with array type.
+        State.G = { E };
```
kkjeer: This doesn't account for adding the use of a temporary to State.G. For example, for the expression `ArrayToPointerDecay(Temp("abc"))`, State.G will only contain `ArrayToPointerDecay(Temp("abc"))`. It will not include `BoundsValue(Temp("abc"))`. The spec doesn't mention adding the use of any temporaries for ArrayToPointerDecay or LValueToRValue casts (or any lvalues). Should uses of temporaries be added to G in these cases?

dtarditi: No, you don't need to do this.   There are two reasons.  First, when we lexicographically compare expression, we'll treat ArrayToPointerDecay(Temp("abc")) as equivalent to ArrayToPointerDeclary(BoundsValue(Temp("abc")).  See CompareExpr around line 222 in AST\CanonBounds.cpp.  Second, I don't believe you'll ever end up with BoundsValue by expression.  It'll still need to be wrapped in the ArrayToPointerDecay operator.

PR #796, opened 2020-02-23
Sort blocks on block IDs instead of using PostOrderCFGView

mgrang:
Currently, we iterate blocks in the topologically sorted order. As a result,
there is no relative ordering between blocks at the same level. For example,
consider the blocks: B1->B3, B2->B3. A topologically sorted order would result
in the order [B1, B3, B2]. This can make the analysis difficult to understand
and debug in case of loops. So we are simplifying the ordering of blocks by
simply sorting them on block IDs which decrease from entry to exit.

### DIFF
--- a/clang/lib/Sema/BoundsAnalysis.cpp
+++ b/clang/lib/Sema/BoundsAnalysis.cpp
@@ -19,36 +19,23 @@ namespace clang {
 void BoundsAnalysis::WidenBounds(FunctionDecl *FD) {
   assert(Cfg && "expected CFG to exist");
 
-  WorkListTy WorkList;
-
-  // Add each block to WorkList and create a mapping from Block to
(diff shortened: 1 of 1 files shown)


dtarditi:
We choose a topologically sorted order because that makes the dataflow analysis more efficient.  It can be the difference between converging in O(N) vs. O(N^2) time.  What happens with this ordering?

mgrang:
> We choose a topologically sorted order because that makes the dataflow analysis more efficient. It can be the difference between converging in O(N) vs. O(N^2) time. What happens with this ordering?

This ordering invokes std::sort, which in turn, uses merge sort which has an O(n log n) runtime.

mgrang:
We decided that we do not want to introduce artificial ordering based on block numbers. A topological sort is more efficient. So I would go ahead and close this PR.

PR #798, opened 2020-02-24
Move bounds expression assertions from TreeTransform to setBoundsExpr

kkjeer:
TreeTransform previously asserted that expressions did not have a bounds expression set to prevent setting a bounds expression twice for an expression. However, in some cases it is expected behavior to transform an expression that has already had a bounds expression set, in particular during the bottom-up analysis in SemaBounds.

This PR removes the asserts from TreeTransform and instead asserts that the bounds expressions are null during setBoundsExpr, since setting the bounds expression twice on any given expression is still unwanted behavior.

Testing:


mgrang:
LGTM.

PR #799, opened 2020-02-24
Support bounds widening for conditionals with while loops

mgrang:
Added support for nt_array_ptr dereferences with while loops. For example, in
"while (*p) {}" we would widen the bounds of p upon entry to the loop.


mgrang:
This is superseded by https://github.com/microsoft/checkedc-clang/pull/803.

PR #802, opened 2020-03-04
Equivalent expressions: assignments [3/n]

kkjeer:
This PR updates the sets UEQ and G of equivalent expressions while checking expressions that involve assignments. This is done with the following changes:

* Update UEQ and G from CheckVarDecl (for declarations with initializers), CheckBinaryOperator (for assignments and compound assignments), and CheckUnaryOperator (for increment/decrement operators).
* Add a GetIncomingBlockState method that gets the state that is true at the beginning of a CFG block by getting the intersection of the UEQ sets of the block's predecessors.
* Add an UpdateAfterAssignment method that updates UEQ and G after an assignment to a variable V.
* Add a GetOriginalValue method that gets the original value of a variable in an expression. Helper methods IsInvertible and Inverse are used to get the original value.
* Add helpers PruneVariableReferences and VariableOccurrenceCount to help with getting the inverse of an expression.
* Add utility classes VariableUtil and ExprCreatorUtil to help with detecting and comparing variables and creating certain kinds of expressions.
* Modify asserts and diagnostic behavior in CanonBounds, TreeTransform, and SemaExpr to reflect the greater variety of expressions that can be lexicographically compared and transformed.

Testing:
* Changed the behavior of the -fdump-checking-state flag to only dump the checking state after checking a top-level CFG statement, rather than after every expression. The equiv-exprs.c test has been modified to reflect this change.
* Added a test equiv-expr-sets.c to test that the set UEQ of sets of equivalent expressions is properly updated while checking assignments. This file also tests that the incoming UEQ set is properly determined for CFG blocks in functions with `if` statements.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -3879,16 +3879,9 @@ namespace {
           EqualExprTy G2 = *I2;
           EqualExprTy IntersectedG = IntersectG(G1, G2);
           if (IntersectedG.size() > 1) {
-            // If IntersectedG is a subset of some set in IntersectedUEQ,
-            // don't add IntersectedG to IntersectedUEQ.
-            bool ContainsIntersectedG = false;
-            for (auto OuterList = IntersectedUEQ.begin(); OuterList != IntersectedUEQ.end(); ++OuterList) {
-              if (IsEqualExprsSubset(IntersectedG, *OuterList, IntersectedUEQ)) {
-                ContainsIntersectedG = true;
-                break;
-              }
-            }
-            if (!ContainsIntersectedG)
+            // Only add IntersectedG if it is not a subset of some set
+            // in IntersectedUEQ.
+            if (!IsSubsetOfSet(IntersectedG, IntersectedUEQ))
               IntersectedUEQ.push_back(IntersectedG);
           }
         }
@@ -3901,14 +3894,8 @@ namespace {
       EqualExprTy IntersectedG;
       for (auto I = G1.begin(); I != G1.end(); ++I) {
         Expr *E1 = *I;
-        for (auto J = G2.begin(); J != G2.end(); ++J) {
-          Expr *E2 = *J;
-          // Don't add duplicate expressions to the intersection.
-          if (EqualExprsContainsExpr(IntersectedG, E2))
-            continue;
-          if (EqualValue(S.Context, E1, E2, nullptr))
-            IntersectedG.push_back(E1);
-        }
+        if (EqualExprsContainsExpr(G2, E1))
+          IntersectedG.push_back(E1);
       }
       return IntersectedG;
     }
@@ -3956,6 +3943,15 @@ namespace {
                                 EquivExprSets *EQ = nullptr) {
       for (auto I = G.begin(); I != G.end(); ++I) {
         if (EqualValue(S.Context, E, *I, EQ))
+          return true;
+      }
+      return false;
+    }
+
+    // IsSubsetOfSet returns true if G is a subset of some set in EQ.
+    bool IsSubsetOfSet(const EqualExprTy G, const EquivExprSets EQ) {
+      for (auto OuterList = EQ.begin(); OuterList != EQ.end(); ++OuterList) {
+        if (IsEqualExprsSubset(G, *OuterList, EQ))
           return true;
       }
       return false;
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -2922,11 +2922,31 @@ namespace {
                              CheckingState &State) {
       BoundsExpr *ResultBounds = CreateBoundsEmpty();
 
-      // If there is an initializer, check it.
       Expr *Init = D->getInit();
       BoundsExpr *InitBounds = nullptr;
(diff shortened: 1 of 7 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -112,6 +112,90 @@ class BoundsUtil {
 };
 }
 
+namespace {
+  class VariableUtil {
+    public:
+      // If E is a (possibly parenthesized or no-op cast) variable V,
+      // GetVariable returns V.  Otherwise, GetVariable returns nullptr.
+      static DeclRefExpr *GetVariable(Sema &SemaRef, Expr *E) {
```
dtarditi: I found this code hard to follow because variable are used in different ways for lvalue and rvalue expressions.  For an assignment, only a possibly-parenthesized DeclRef (an lvalue) occurs on the left-hand side of an operator.  When the value is read, there is an lvalue-to-rvalue cast.

 think you should have versions of GetVariable that are distinct for lvalue vs. rvalue expressions.  For an lvalue expression, the method would check for a possibly-parenthesized DeclRef expressions.  For an rvalue expression, it would check for a Ipossibly-parenthesized lvalue-to-rvalue cast of a possibly-parenthesized DeclRef expression.

IgnoreParenNoopCasts isn't the right method to use here.  LValue-to-rvalue casts actually read memory, and that isn't a no-op cast.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -112,6 +112,90 @@ class BoundsUtil {
...
+        if (!E)
+          return nullptr;
+        return dyn_cast<DeclRefExpr>(E->IgnoreParenNoopCasts(SemaRef.Context));
+      }
+
+      // SameVariable returns true if the expressions E1 and E2
+      // are the same (possibly parenthesized or no-op cast) variable.
+      static bool SameVariable(Sema &SemaRef, Expr *E1, Expr *E2) {
```
dtarditi: If you look at all uses of SameVariable, I believe the first expression is always a DeclRef expression.  I think it would be clearer to change the type of that argument.   You may need specialized versions for lvalue vs. rvalue expressions.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2327,8 +2518,50 @@ namespace {
         }
       }
 
+      // Update State.UEQ and State.G.
```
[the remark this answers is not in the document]
dtarditi: I didn't see this comment addressed in your most recent change.  It would have been helpful to have a clarifying response here about why it wasn't addressed.

I looked into this and see that UpdateG is screening out non-modifying expressions in the S_i.   I am pretty concerned about this.  Expression equivalence checking assumes that expressions are non-modifying expressions and may produce incorrect results if they are modifying expressions.   

For this reason, I think we should have an invariant that none of the subexpressions G_i containy modifying expression. This change is already large and I'd rather land it now, then have it continune onward.  @kkjeer could you file an issue about this, and follow-up in a subsequent check-in.

clang/test/CheckedC/inferred-bounds/equiv-expr-sets.c
```diff
@@ -0,0 +1,1228 @@
...
+  // CHECK-NEXT: ImplicitCastExpr {{.*}} <LValueToRValue>
+  // CHECK-NEXT:   DeclRefExpr {{.*}} 'i'
+  // CHECK-NEXT: }
+  // CHECK-NEXT: }
+}
+
+// BinaryOperator: compound assignment to a variable
+void f3(unsigned i) {
+  // Updated UEQ: { { (i - 2) + 2, i } }
+  i += 2;
```
dtarditi, line 67: This first expression in the UEQ set is a tautology.  When updating UEQ information, we should screen out equality facts implied by assignments to variables where the variable also appears on the right-hand side of the assignments.   It's fine to flag this as a follow-up issue.  This is already a large pull request.

clang/lib/Sema/SemaBounds.cpp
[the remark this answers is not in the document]
kkjeer: I created issue #815 to track this work.

clang/test/CheckedC/inferred-bounds/equiv-expr-sets.c
kkjeer: I created issue #814 to track this work.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2699,6 +2898,52 @@ namespace {
...
+          // expressions that produce the same value as the variable `e1`,
+          // and these expressions should not be added to UEQ.
+          State.G.clear();
+          Expr *OV = GetOriginalValue(V, RHS, State.UEQ);
+          UpdateAfterAssignment(V, Target, OV, CSS, State, State);
+        }
+
+        // Update the set G of expressions that produce the same value as e.
+        if (RHS) {
+          // For post-inc/dec operators, use the rvalue for `e1` to update G.
```
dtarditi, line 2934: This logic is reversed.   In the state after the expression ++x has executed, the non-modifying expression equivalent to the value produced by ++x is x.   For x++, the non-modifying expression is x - 1..

PR #803, opened 2020-03-06
Support bounds widening for conditionals with while loops

mgrang:
Added support for widenening bounds for nt_array_ptr dereferences in while
loops. For example, "while (*p) {}" would widen the bounds of p upon
entry to the loop.

To compute In[B] we compute the intersection of Out[B*->B], where B* are all
preds of B. When there is a back edge from block B' to B (for example in
loops), the Out set for block B' will be empty when we first enter B. As a
result, the intersection operation would always result in an empty In set for
B.

So to handle this, we consider the In and Out sets for all blocks to have a
default value of "Top" which indicates a set of all members of the Gen set. In
this way we ensure that the intersection does not result in an empty set even
if the Out set for a block is actually empty.

But we also need to handle the case where there is an unconditional jump into a
block (for example, as a result of a goto). In this case, we cannot widen the
bounds because we would not have checked for the ptr dereference. So in this
case we want the intersection to result in an empty set.

So we initialize the In and Out sets of all blocks, except the Entry block, as
"Top".

Top represents the union of the Gen sets of all edges. We have chosen the
offsets of ptr variables in Top to be the max unsigned int. The reason behind
this is that in order to compute the actual In sets for blocks we are going to
intersect the Out sets on all the incoming edges of the block. And in that case
we would always pick the ptr with the smaller offset. Chosing max unsigned int
also makes handling Top much easier as we do not need to explicitly store edge
info.

### DIFF
--- a/clang/include/clang/Sema/BoundsAnalysis.h
+++ b/clang/include/clang/Sema/BoundsAnalysis.h
@@ -116,6 +116,26 @@ namespace clang {
       // block.
       BoundsVarTy BoundsVars;
 
+      // To compute In[B] we compute the intersection of Out[B*->B], where B*
+      // are all preds of B. When there is a back edge from block B' to B (for
+      // example in loops), the Out set for block B' will be empty when we
+      // first enter B. As a result, the intersection operation would always
+      // result in an empty In set for B.
+
+      // So to handle this, we initialize the In and Out sets for all blocks as
+      // Top so that the intersection does not result in an empty In set.
+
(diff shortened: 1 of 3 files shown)


mgrang:
Here's a follow-up PR for "for" loops: https://github.com/microsoft/checkedc-clang/pull/804

mgrang:
@dtarditi My latest updated PR computes Top and initializes In and Out sets to Top. Please review.

PR #807, opened 2020-03-09
Bounds context [1/n]

kkjeer:
This PR adds a bounds context member to the CheckingState class and determines the incoming (declared) bounds context for a statement. After #802 is merged, the bounds context can be updated during bounds checking.

Testing:
* Added bounds-context.c to test dumping the bounds context.
* Automated testing in progress.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -2027,6 +2027,8 @@ namespace {
             Check(S, CSS, BlockState);
             // TODO: validate the updated context BlockState.UC against
             // the declared context DC.
+            if (DumpState)
+              DumpCheckingState(llvm::outs(), S, BlockState);
          }
        }
        AFA.Next();
@@ -2169,9 +2171,6 @@ namespace {
           break;
       }
 
-      if (DumpState)
-        DumpCheckingState(llvm::outs(), S, State);
-
       if (Expr *E = dyn_cast<Expr>(S)) {
         // Bounds expressions are not null ptrs.
         if (isa<BoundsExpr>(E))
@@ -2249,9 +2248,6 @@ namespace {
           CheckChildren(E, CSS, State);
           break;
       }
-
-      if (DumpState)
-        DumpCheckingState(llvm::outs(), E, State);
 
       // The type for inferring the target bounds cannot ever be an array
       // type, as these are dealt with by an array conversion, not an lvalue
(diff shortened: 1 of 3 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -614,6 +653,37 @@ namespace {
...
+        // The keys in an llvm::DenseMap are unordered.  Create a set of
+        // variable declarations in the context ordered first by name,
+        // then by location in order to guarantee a deterministic output
+        // so that printing the bounds context can be tested.
+        std::vector<DeclaratorDecl *> OrderedDecls;
+        for (auto Pair : UC)
+          OrderedDecls.push_back(Pair.first);
+        llvm::sort(OrderedDecls.begin(), OrderedDecls.end(),
+             [] (DeclaratorDecl *A, DeclaratorDecl *B) {
+               if (A->getNameAsString() == B->getNameAsString())
```
mgrang: You could write this comparator more succinctly as:
`
return std::tie(A->getNameAsString(), A->getLocation()) <
           std::tie(B->getNameAsString(), B->getLocation())
`

kkjeer: After I made this change, there were build warnings related to `A->getNameAsString()`, etc. not being lvalues so I've reverted it at least for now.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2867,7 +2958,10 @@ namespace {
       BoundsExpr *B = nullptr;
       InteropTypeExpr *IT = nullptr;
       if (VD) {
-        B = VD->getBoundsExpr();
+        if (State.UC[VD])
```
dtarditi: Should use the find method instead of the subscript operator?   This avoids two lookups in the DenseMap.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2958,8 +2961,9 @@ namespace {
       BoundsExpr *B = nullptr;
       InteropTypeExpr *IT = nullptr;
       if (VD) {
-        if (State.UC[VD])
-          B = State.UC[VD];
+        auto Declared = State.UC.find(VD);
```
mgrang: I would do it this way:
```
if (State.UC.count(VD))
  B = State.UC[VD];
```

kkjeer:
> Overall, looks good. I had one minor code suggestion.
> 
> Could you add comments explaining the general approach that is being used? As we discussed off-line, the goal of attaching contexts and propagating them around is that we can determine what variables with bounds declarations are in scope? Also, what is the effect on compile time of this change?

This change resulted in the following compile time deltas compared to the current master:
* Debug X64 Windows: 1.5% decrease in compile time
* Debug X86 Windows: 11% increase in compile time
* Debug X64 Linux: 1.7% decrease in compile time

PR #804, opened 2020-03-06
Support bounds widening for conditionals with for loops

mgrang:
Added support for widenening bounds for nt_array_ptr dereferences in for
loops. For example, "for (; *p; ) {}" would widen the bounds of p upon
entry to the loop.

PR #808, opened 2020-03-11
Use the widened bounds to update bounds in context

mgrang:
Use the widened bounds calculated by the analysis in BoundsAnalysis.cpp to
    widen the bounds of nt_array_ptr.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -2027,6 +2027,8 @@ namespace {
             Check(S, CSS, BlockState);
             // TODO: validate the updated context BlockState.UC against
             // the declared context DC.
+            if (DumpState)
+              DumpCheckingState(llvm::outs(), S, BlockState);
          }
        }
        AFA.Next();
@@ -2169,9 +2171,6 @@ namespace {
           break;
       }
 
-      if (DumpState)
-        DumpCheckingState(llvm::outs(), S, State);
-
       if (Expr *E = dyn_cast<Expr>(S)) {
         // Bounds expressions are not null ptrs.
         if (isa<BoundsExpr>(E))
@@ -2249,9 +2248,6 @@ namespace {
           CheckChildren(E, CSS, State);
           break;
       }
-
-      if (DumpState)
-        DumpCheckingState(llvm::outs(), E, State);
 
       // The type for inferring the target bounds cannot ever be an array
       // type, as these are dealt with by an array conversion, not an lvalue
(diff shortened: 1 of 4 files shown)


mgrang:
This PR depends on the bounds context PR (https://github.com/microsoft/checkedc-clang/pull/807). Once that is merged, I will add tests to this PR. Also in order to add tests for loops to this PR the PRS (https://github.com/microsoft/checkedc-clang/pull/803, https://github.com/microsoft/checkedc-clang/pull/804 and https://github.com/microsoft/checkedc-clang/pull/805) need to merge.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1896,15 +2004,53 @@ namespace {
...
+       // Also get the bounds killed (if any) by each statement in the current
+       // block.
+       StmtDeclSetTy KilledBounds = BA.GetKillSet(Block);
+
+       // Bounds killed by a statement in the current block remain killed
+       // through the end of the block. So we remove those bounds from the
+       // widened bounds for the current block.
+       for (const auto item : KilledBounds) {
+         for (const VarDecl *V : item.second)
+           WidenedBounds.erase(V);
```
kkjeer: Does this kill bounds widened in the current block even for statements that occur before the bounds are killed? E.g.
```
void f(nt_array_ptr p : count(len), int len) {
    if (*p) { // widen bounds of p
        p[1]; // bounds of p should still be widened here
        len = 0; // kill the widened bounds of p
    }
}
```
It seems like the widened bounds of `p` will be killed for the entire block, before the statement `len = 0;` kills them.

mgrang: Yes, my original intent was to remove from WidenedBounds only the bounds killed by the current statement. But the problem is that BlockState.UC would also need to be updated to set the bounds for the VarDecl to the declared bounds.
 For this GetDeclaredBounds would need to be called outside the loop and the context would need to be stored in a declared context.
Otherwise, this case does not work:
`
nt_array_ptr p : count(0);
int i;
if (*p) {
  i = 0;
  p++;
  if (*(p + 1)) {}
}
Here, the BlockState.UC is updated with p:1 inside the block `if (*p)`. Then the bounds are killed at `p++`. But the BlockState.UC still has `p:1`. So in this case we also need to reset the context with the declared bounds for p which are `p:0`.

mgrang:
In the latest change set, I have fixed the handling of bounds killed by a statement. For bounds killed by a statement, we reset the bounds to the declared bounds. So we first gather the declared bounds and then store them in a declared bounds context. Then when we find that the bounds for a variable are killed by a statement we reset those bounds from the declared context.

We still do not handle the following case:
```
void f(nt_array_ptr p : count(len), int len) {
    if (*(p + len + 1) {} // We need an out-of-bounds error as the upper bound of p is (p + len). But the checker does not currently understand that the dereference should only be at the upper bound.
}
```

mgrang:
Closing this in favor of https://github.com/microsoft/checkedc-clang/pull/821

PR #812, opened 2020-03-21
Fix the handling of killed bounds with loops

mgrang:
The dataflow equation for the Out set is:
`
  Out[B1->B2] = (In[B1] - Kill[B1]) ∪ Gen[B1->B2]
`

So if a variable is in the Gen set for an edge then it would be added to the
Out set for the edge even if it is killed inside the outgoing block. This
happens because of the union operation with Gen. This becomes problematic for
loops and we end up widening the bounds in the following case:

```
  while (*p)
    p++;
```

So we handle this as:
  1. Compute the Out set for a block `B` as usual.
  2. Then, if a variable `V` is in `Kill[B]`, remove `V` from the `Gen[B->B']`.

This would have no effect on liner code but in loops this would ensure we do
not widen bounds in the above case.

### DIFF
--- a/clang/lib/Sema/BoundsAnalysis.cpp
+++ b/clang/lib/Sema/BoundsAnalysis.cpp
@@ -645,6 +645,34 @@ void BoundsAnalysis::ComputeOutSets(ElevatedCFGBlock *EB,
 
     if (Differ(OldOut, EB->Out[succ]))
       WorkList.append(BlockMap[succ]);
+
+    // The dataflow equation for the Out set is:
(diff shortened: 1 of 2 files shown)


mgrang:
Had a discussion with @dtarditi and he clarified that we indeed need to widen the bounds upon entry to the block after the while condition. The bounds may be killed later inside the block and thus the Out of the block may result in empty set. But this does not preclude us from widening the bounds upon entry.
So I will go ahead and close this PR as this is no longer needed.

Issue #810, opened 2020-03-20
Generalize assignment compatibility for bounds-safe interfaces


kkjeer:
The current rules for assignment compatibility of expressions involving bounds-safe interfaces are missing two cases.

The first case is related to assignments involving nested pointers (see #797). GetCheckedCInteropType should be extended to handle dereferences and pointer arithmetic.

The second case is related to assignments of an unchecked pointer with an `_Nt_array_ptr` bounds-safe interface to an `_Nt_array_ptr` variable (see #806). The type checker should check the bounds-safe interface of the unchecked pointer in this case.

kkjeer:
This issue tracks the implementation of the spec updates described in [microsoft/checkedc/#397](https://github.com/microsoft/checkedc/issues/397).

kkjeer:
This issue was fixed in the [2020-07-31 release](https://github.com/microsoft/checkedc-clang/releases/tag/v0.8-dev-build-2020-07-31) of Checked C.

Issue #398, opened 2017-10-19
Rename bounds(none) to bounds(unknown)


dtarditi:
We need to change the compiler to recognize bounds(unknown) instead of bounds(none) and update the tests.   This will match the specification change made in https://github.com/Microsoft/checkedc/pull/209.

dtarditi:
This is a duplicate of #375.

PR #813, opened 2020-03-31
Fill in some missing cases for bounds-safe interface assignments

kkjeer:
This PR addresses issue #810 by filling in two missing cases for bounds-safe interface assignment compatibility:

1. Extend GetCheckedCInteropType to get the bounds-safe interface type for pointer dereferences and array subscripts
2. Allow assignment of an unchecked nt_array_ptr with a bounds-safe interface to a checked nt_array_ptr

Testing:
* Added Checked C tests to test the new assignment compatibility cases (PR [#398](https://github.com/microsoft/checkedc/pull/398))


clang/lib/Sema/SemaExpr.cpp
```diff
@@ -9124,6 +9152,43 @@ QualType Sema::GetCheckedCInteropType(ExprResult LHS) {
...
 }
 
+/// Get the bounds-safe interface type for an rvalue expression, if the
+/// rvalue expression has a bounds-safe interface. Return a null QualType
+/// otherwise.  The rvalue expression may appear as part of the left-hand
+/// side of an assignment - for example, as the subexpression of a pointer
+/// deference or an array subscript.  For rvalue expressions appearing as
+/// part of the left-hand side of an assignment, only lvalue-to-rvalue casts
+/// and pointer arithmetic have bounds-safe interfaces.
+QualType Sema::GetCheckedCRValueInteropType(ExprResult RHS) {
```
kkjeer: In addition to returning the bounds-safe interface type for `*e` and `p +/- i`, should this method also return the bounds-safe interface type for `&e`? If `e` has bounds-safe interface type `T`, should `&e` have bounds-safe interface type `ptr<T>`?

As an example, making this change would allow the following to compile:
```
void address_of(int *interop_ptr : itype(ptr<int>), ptr<int> checked_ptr) {
  *(&interop_ptr) = checked_ptr;
}
```
Currently, this example does not compile since `&interop_ptr` has an unchecked pointer type `int **`. If GetCheckedCRValueInteropType were to return the bounds-safe interface type for address-of expressions, then `&interop_ptr` would have a bounds-safe interface type of `ptr<ptr<int>>` and so `*(&interop_ptr)` would have a bounds-safe interface type of `ptr<int>`.

clang/lib/Sema/SemaExpr.cpp
```diff
@@ -8479,6 +8479,18 @@ Sema::CheckAssignmentConstraints(QualType LHSType, ExprResult &RHS,
   LHSType = Context.getCanonicalType(LHSType).getUnqualifiedType();
   RHSType = Context.getCanonicalType(RHSType).getUnqualifiedType();
 
+  // If the LHS is a checked nt array type and the RHS is an unchecked
+  // nt array type with a bounds-safe interface, use the bounds-safe
+  // interface of the RHS to check the types.
+  if ((LHSType->isNtCheckedArrayType() ||
```
dtarditi, line 8485: Is the test isNtCheckedArrayType needed?   I don't think C allows you to make assignments to values with array type.

clang/lib/Sema/SemaExpr.cpp
```diff
@@ -9116,6 +9128,23 @@ QualType Sema::GetCheckedCInteropType(ExprResult LHS) {
         IsParam = isa<ParmVarDecl>(Var);
       }
     }
+    // If `e` has bounds-safe interface type with referent type T, then
```
dtarditi, line 9131: I think the wording on this comment is somewhat unclear.   It's subtle to say that e has a referent type.   How about `e` has a bounds-safe interface type that is a pointer to T?

clang/lib/Sema/SemaExpr.cpp
```diff
@@ -9116,6 +9128,23 @@ QualType Sema::GetCheckedCInteropType(ExprResult LHS) {
...
+      if (Unary->getOpcode() == UnaryOperatorKind::UO_Deref) {
+        QualType T = GetCheckedCRValueInteropType(Unary->getSubExpr());
+        if (!T.isNull())
+          return QualType(T->getPointeeOrArrayElementType(), 0);
+      }
+    }
+    // If `e1` has bounds-safe interface type with referent type T and `e2`
+    // is an integer, then `e1[e2]` and `e2[e1`] have bounds-safe interface
+    // type T.
+    else if (ArraySubscriptExpr *Array = dyn_cast<ArraySubscriptExpr>(LHSExpr)) {
```
dtarditi, line 9143: I think the same change applies here.

clang/include/clang/Sema/Sema.h
```diff
@@ -10565,6 +10565,10 @@ class Sema {
   /// Returns a null QualType if there isn't one.
   QualType GetCheckedCInteropType(ExprResult LHS);
```
dtarditi, line 10566: Does this need a clarifying comment?  Or to be renamed?

Issue #814, opened 2020-04-02
Omit tautologies from UEQ equivalent expressions set


kkjeer:
For a compound assignment to a variable, e.g. `i += 2`, the UEQ set currently will contain `{ (i - 2) + 2, i}`. This is the same as saying that `i == i` since `(i - 2 + 2) == i` (ignoring integer overflow), so this is a tautology and should not be included in UEQ.

The UEQ set should not contain equality facts implied by assignments to variables where the variable also appears on the right-hand side of the assignment.

PR #817, opened 2020-04-07
Equivalent expressions: omit tautologies [4/n]

kkjeer:
See issue #814 

This PR modifies the UpdateAfterAssignment method to prevent adding sets such as `{ (i - 2) + 2, i }` to UEQ. These kinds of sets record equality between two mathematically equivalent expressions, which is a tautology and does not add any meaningful equality information to UEQ.

The approach is to modify the GetOriginalValue method to set a flag OVUsesV if the original value of a source expression with respect to a variable uses the value of the variable. UpdateAfterAssignment uses this flag to determine whether expressions in G that use the value of the variable should be adjusted using the original value or should be removed from G.

Testing:
* Modified the equiv-expr-sets.c Clang test to reflect the omissions of tautologies from UEQ.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -3706,7 +3706,8 @@ namespace {
     // value as the only subexpression of e.
     //
     // Val is an optional expression that may be contained in the updated G.
-    // If Val is not provided, e is used instead.
+    // If Val is not provided, e is used instead.  If Val and e are null,
+    // G is not updated.
     void UpdateG(Expr *E, const EqualExprTy SubExprG,
                  EqualExprTy &G, Expr *Val = nullptr) {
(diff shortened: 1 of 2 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3719,6 +3735,8 @@ namespace {
       G.clear();
 
       if (!Val) Val = E;
+      if (!Val)
```
dtarditi, line 3738: This change doesn't line up with the comment for UpdateG that says Val is an optional expression and that e is used if Val is not provided.

kkjeer: If the optional Val is not provided, e is used instead. This is an additional check so that if Val and e are both null, UpdateG early returns. I've updated the comments for UpdateG to reflect this.

PR #820, opened 2020-04-17
Equivalent expressions: various fixes [5/n]

kkjeer:
This PR makes several changes to equivalent expression set handling that will be necessary to update the bounds context during assignments.

#### Context:
This is part of the ongoing work to improve the soundness of bounds checking involving assignments to variables. The next step in this process will be to update the bounds context mapping variables to their observed bounds after assignments to variables. That process depends on fixing some issues with the current approach to tracking equivalent expression sets. The process of getting the original value of a variable after an assignment partially depends on the contents of the equivalent expression sets. This PR fixes some issues with these sets that would affect the original value for updating the bounds context.

##### When searching for a set F in UEQ containing the target variable v:
Consider the assignments:
```
array_ptr<int> arr_int;
array_ptr<const int> arr_const = arr_int;
arr_int = 0;
```
At the assignment `arr_int  =0`, UEQ contains the set F = `{ (array_ptr<const int>)arr_int, arr_const }`. When getting the original value of `arr_int`, GetOriginalValue should find the set F containing the value-preserving cast `(array_ptr<const int>)arr_int`, and return `arr_const` as the original value of `arr_int`.

##### When searching for a variable w != v:
Consider the assigments:
```
array_ptr<const int> arr_const;
array_ptr<int> arr_int = arr_const;
arr_int = 0;
```
At the assignment `arr_int = 0;` UEQ contains the set F = `{ (array_ptr<int>)arr_const, arr_int }`. When getting the original value of `arr_int`, GetOriginalValue should find the value-preserving cast `(array_ptr<int>)arr_const` in F and return `(array_ptr<int>)arr_const` as the original value of `arr_int`.

#### Add a type compatibility check to GetOriginalValue when searching for a variable w != v:
Consider the following assignments:
```
int val = 0;
array_ptr<int> arr = NullToPointer(0);
array_ptr<int> uses_arr = arr + 1;
arr = some_other_value;
```
Since `0` and `NullToPointer(0)` are considered equivalent, UEQ contains the set `{ 0, val, arr }` after these assignments. When getting the original value for `arr` in the fourth assignment, the variable `val` appears in the same set as `arr` and would be the original value of `arr` without this change. Then equality between `uses_arr` and `val + 1` would be recorded. This is incorrect since `val` is an integer and should not be used as the original value of the array_ptr `arr`. These errors can happen since UEQ does not consider type information when recording equality between expressions, so expressions that do not have the same type can be included in the same set in UEQ. With the change in this PR to add the type compatibility check, the original value of the array_ptr `arr` would no longer be the integer `val`.
One subtlety here is assignments involving value-preserving casts. Consider the following assignments:
```
array_ptr<int> arr, array_ptr<const int> arr_const;
arr_const = (array_ptr<const int>)arr;
```
After these assignments, UEQ contains the set `{ (array_ptr<const int>)arr, arr_const }`. When getting the original value of `arr`, GetOriginalValue finds the set F = `{ (array_ptr<const int>)arr, arr_const }` that contains the variable `arr` (since `(array_ptr<const int>)` is a value-preserving cast. The original value of `arr` should be `arr_const`. Since the type compatibility check does not consider `array_ptr<int>` (the type of `arr`) and `array_ptr<const int>` (the type of arr_const) to be compatible, GetOriginalValue would ordinarily not return `arr_const` as the original value of `arr`. To mitigate this, the type compatibility check uses the type of the value-preserving cast `(array_ptr<const int>)arr` expression that appears as `arr` in the set in UEQ, and checks for type compatibility between `array_ptr<const int>` (the type of the value-preserving cast of `arr`) and `array_ptr<const int>` (the type of `arr_const`). 

#### Testing;
* Added tests to equiv-expr-sets.c to value-preserving casts and type compatibility of original values

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -2610,7 +2610,7 @@ namespace {
 
         // Update UEQ and G for assignments to `e1` where `e1` is a variable.
         if (DeclRefExpr *V = GetLValueVariable(LHS)) {
-          Expr *OV = GetOriginalValue(V, Src, State.UEQ);
+          Expr *OV = GetOriginalValue(V, Target, Src, State.UEQ);
           UpdateAfterAssignment(V, Target, OV, CSS, State, State);
         }
         // Update UEQ and G for assignments where `e1` is not a variable.
@@ -3021,7 +3021,7 @@ namespace {
           // expressions that produce the same value as the variable `e1`,
           // and these expressions should not be added to UEQ.
           State.G.clear();
-          Expr *OV = GetOriginalValue(V, RHS, State.UEQ);
+          Expr *OV = GetOriginalValue(V, Target, RHS, State.UEQ);
           UpdateAfterAssignment(V, Target, OV, CSS, State, State);
         }
 
@@ -3770,7 +3770,11 @@ namespace {
 
     // GetOriginalValue returns the original value (if it exists) of the
     // expression Src with respect to the variable V in an assignment V = Src.
-    Expr *GetOriginalValue(DeclRefExpr *V, Expr *Src, const EquivExprSets EQ) {
+    //
+    // Target is the target expression of the assignment (that accounts for
+    // any necessary casts of V).
+    Expr *GetOriginalValue(DeclRefExpr *V, Expr *Target, Expr *Src,
+                           const EquivExprSets EQ) {
       // Check if Src has an inverse expression with respect to v.
       Expr *IV = nullptr;
       if (IsInvertible(V, Src))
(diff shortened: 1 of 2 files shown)


kkjeer:
> That's an interesting problem that you hit with NullToPointer casts, where a type mismatch results. I think the problem could occur in other ways because expression equality ignores types and looks for expressions that produce equivalent values. An alternative fix is to change GetOriginalValue to take into account type compatibility. If two variables x and y are in the same equality set, but y's type is not compatible with x's type, don't allow them to the original value for each other.

I'm running automated tests on this approach to check the performance impact, since this means that a) the UEQ set will be larger and b) there is additional work to check type compatibility in GetOriginalValue. If the performance isn't unacceptably impacted, I prefer this approach since it doesn't include special logic for NullToPointer casts.

kkjeer:
> Do you intend to keep the tests but not run them? When would you re-enable them?

I temporarily disabled these tests in order to run automated tests on the type compatibility check. The type compatibility check would cause these tests to fail without other changes being made to compensate, but I wanted to test the type compatibility check in isolation. Once all necessary changes have been made and tested, I'll reenable these tests and mark the PR as non-draft.

PR #821, opened 2020-04-17
Use widened bounds to update bounds in context

mgrang:
Use the widened bounds calculated by the analysis in BoundsAnalysis.cpp to
widen the bounds of nt_array_ptr.


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2123,6 +2123,71 @@ namespace {
       }
    }
 
+   void ResetKilledBounds(StmtDeclSetTy &KilledBounds,
```
dtarditi, line 2126: Could you add a comment describing what this function is doing?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2157,11 +2222,29 @@ namespace {
...
+
+       // Get the widened bounds for the current block as computed by the
+       // bounds widening analysis invoked by WidenBounds above.
+       BoundsMapTy WidenedBounds = BA.GetWidenedBounds(Block);
+       // Also get the bounds killed (if any) by each statement in the current
+       // block.
+       StmtDeclSetTy KilledBounds = BA.GetKilledBounds(Block);
+       // Update the bounds in context (BlockState.UC) using the bounds
+       // widening info calculated above.
+       UpdateBoundsInContext(WidenedBounds, BlockState.UC);
```
kkjeer, line 2246: After more discussion on the relationship between widened bounds, target bounds, and the bounds context, I'm currently updating the way the bounds context is handled. In particular, CheckDeclRefExpr will no longer use the BlockState.UC context to return the lvalue target bounds (since the widened bounds should not affect the target bounds for a variable). We may need to reevaluate how to use the widened bounds as the observed rvalue bounds for the value of a variable in light of my proposed updates to the bounds context.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2123,6 +2123,71 @@ namespace {
...
+
+       const llvm::APInt
+         APIntOff(Context.getTargetInfo().getPointerWidth(0), Offset);
+       IntegerLiteral *WidenedOffset = CreateIntegerLiteral(APIntOff);
+
+       Expr *Lower = RBE->getLowerExpr();
+       Expr *Upper = RBE->getUpperExpr();
+
+       // NewUpperBound = UpperBound + WidenedOffset.
+       Expr *NewUpper = ExprCreatorUtil::CreateBinaryOperator(
```
kkjeer, line 2181: I tested this locally and ran into some issues with checking an assignment using widened bounds.

In this example:
```
void widen_bounds_bug(nt_array_ptr<int> a : bounds(a, a + 1),
                      nt_array_ptr<int> p : bounds(p, p + 0),
                      nt_array_ptr<int> b : bounds(b, b + (0 + 1))) {
  if (*p) {
    // Bounds of p have been widened to (p, (p + 0) + 1).
    // The target bounds of a are (a, a + 1).
    // Because the base of p's bounds is p + 0,
    // ProveBoundsDeclValidity is unable to create a base range for p's bounds,
    // and is therefore unable to prove the validity of a's bounds after the assignment.
    a = p;

    // Bounds of b are (b, b + (0 + 1)).
    // The target bounds of a are (a, a + 1).
    // Because the base of b's bounds is b,
    // ProveBoundsDeclValidity is able to create a base range for b's bounds,
    // and is therefore able to prove the validity of a's bounds after the assignment.
    a = b;
  }
}
```
The assignment `a = p` results in a warning "unable to prove the declared bounds of a are valid after the assignment". The reason is that the new upper expression for p's widened bounds `(p, (p + 0) + 1)` is left-associative, so the base of the widened bounds is `p + 0`. If the upper expression were right-associative `(p, p + (0 + 1))`, as in the declared bounds for `b`, then the assignment `a = p` would be fine (as in the assignment `a = b`).

I quickly put together a local fix involving `SplitIntoBaseAndOffset` to split the upper expression of a variable's bounds and use the resulting base and offset to create the new upper expression for the widened bounds, but I haven't fully tested it. Below is the rough approach:
```
Expr *UpperBase;
llvm::APSInt  UpperOffsetConstant(1, true);
Expr *UpperOffsetVariable = nullptr;
SplitIntoBaseAndOffset(Upper, UpperBase, UpperOffsetConstant, UpperOffsetVariable);

Expr *WidenedUpper;
if (UpperOffsetVariable) {
  WidenedUpper = ExprCreatorUtil::CreateBinaryOperator(S, UpperOffsetVariable, WidenedOffset, BinaryOperatorKind::BO_Add);
} else {
  IntegerLiteral *UpperConst = CreateIntegerLiteral(UpperOffsetConstant);
  WidenedUpper = ExprCreatorUtil::CreateBinaryOperator(S, UpperConst, WidenedOffset, BinaryOperatorKind::BO_Add);
}

Expr *NewUpper = ExprCreatorUtil::CreateBinaryOperator(S, UpperBase, WidenedUpper, BinaryOperatorKind::BO_Add);
```

mgrang: Yes, SplitIntoBaseAndOffset does currently handle associativity and multiple arithmetic operators correctly. In BoundsAnalysis I ran into a similar issue wherein I had to correctly determine the offset of the expression being dereferenced. So I wrote an implementation of SplitIntoBaseAndOffset (see https://github.com/microsoft/checkedc-clang/blob/master/clang/lib/Sema/BoundsAnalysis.cpp#L340). This function associates all variables to the LHS and all constants to the RHS. Ideally, we want to merge both these functions into one.

One thing to keep in mind with changing associativity of operators is overflow. The original expression may not overflow but if we change the associativity then an overflow/underflow may occur.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2123,6 +2123,71 @@ namespace {
...
+                              BoundsContextTy &BoundsCtx) {
+     for (const auto item : WidenedBounds) {
+       VarDecl *V = const_cast<VarDecl *>(item.first);
+       unsigned Offset = item.second;
+
+       BoundsExpr *Bounds = V->getBoundsExpr();
+       if (!Bounds)
+         continue;
+
+       // If the declared bounds are CountBoundsExpr, normalize them to
```
kkjeer, line 2154: After merging #827, all bounds in the bounds context should already be a normalized RangeBoundsExpr.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2123,6 +2123,71 @@ namespace {
...
+       BoundsExpr *Bounds = VD->getBoundsExpr();
+       assert(Bounds && "Invalid bounds");
+
+       if (isa<CountBoundsExpr>(Bounds))
+         Bounds = ExpandToRange(VD, Bounds);
+       BoundsCtx[VD] = Bounds;
+     }
+   }
+
+   void UpdateBoundsInContext(BoundsMapTy &WidenedBounds,
```
kkjeer, line 2144: Could you add a comment to this function as well as ResetKilledBounds?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2223,6 +2288,10 @@ namespace {
             BoundsContextTy InitialObservedBounds = BlockState.ObservedBounds;
             BlockState.G.clear();
 
+            // If any bounds are killed by statement S, remove their bounds
+            // from the ObservedBounds.
+            ResetKilledBounds(KilledBounds, S, BlockState.ObservedBounds);
```
kkjeer, line 2293: I think this should be done before saving `InitialObservedBounds = BlockState.ObservedBounds`. `InitialBounds` should contain the contents of `ObservedBounds` before calling `Check(S, CSS, BlockState)` in order for the observed bounds to be reset after the `Check` call.

For example:
```
void pointer_widening() {
  nt_array_ptr<char> p : count(0) = "a";

  if (p[0]) {
    // This assignment kills the widened bounds of p.
    // InitialObservedBounds should contain p => bounds(p, p + 0),
    // not p => bounds(p, p + 0 + 1).
    // After checking this assignment, BlockState.ObservedBounds should contain
    // p => bounds(p, p + 0).
    p = "";
    if (p[1]) // observed bounds of p are (p, p + 0), so this is an out-of-bounds memory access
  }
}
``

File: clang/lib/Sema/SemaBounds.cpp [hunk 7]
```diff
@@ -2142,6 +2142,56 @@ namespace {
...
+     if (I == KilledBounds.end())
+       return;
+
+     // KilledBounds stores a mapping of statements to all variables whose
+     // bounds are killed by each statement. Here are remove the bounds of all
+     // such variables from ObservedBounds whose bounds are killed by the
+     // statement S. After removal, the bounds of these variables would default
+     // to the user declared bounds in DeclaredBounds.
+     for (const VarDecl *V : I->second)
+       ObservedBounds.erase(V);
```
kkjeer, line 2157: I think this should reset `ObservedBounds[V]` to the normalized declared bounds of `V`, rather than erasing `V` from the context. Before calling `Check` on a statement `S`, the `ObservedBounds` context should contain, for each variable `V` that is in scope at `S`:
1. If `V` has widened bounds `(lb, ub + i`) at `S`, `ObservedBounds[V] => bounds(lb, ub + i)`
2. Otherwise, if `V` has declared bounds `(lb, ub)`, `ObservedBounds[V] => bounds(lb, ub)`

For example:
```
void pointer_widening() {
  nt_array_ptr<char> p : count(0) = "a";

  if (p[0]) {
    // This assignment kills the widened bounds (p, p + 1) of p.
    // Before checking this assignment, ObservedBounds should contain p => (p, p + 0).
    // The original value of p in this assignment is p - 1. p - 1 should replace p in the observed bounds of p.
    // After checking this assignment, ObservedBounds should contain p => (p - 1, p -1 + 0).
    p = p + 1;
  }
}
```

mgrang:
This depends on https://github.com/microsoft/checkedc-clang/pull/829

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2218,8 +2283,11 @@ namespace {
             // bounds for each variable v that is in scope are the widened
             // bounds for v (if any), or the declared bounds for v (if any).
             GetDeclaredBounds(this->S, BlockState.ObservedBounds, S);
-            // TODO: update BlockState.ObservedBounds to reset any widened
-            // bounds that are killed by S to the declared variable bounds.
+
+            // If any bounds are killed by statement S, remove their bounds
```
kkjeer, line 2287: This comment no longer seems quite accurate, since the bounds that are killed by S are being reset to their declared bounds rather than removed from ObservedBounds.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2142,6 +2142,56 @@ namespace {
...
+     // the offset to the declared bounds of the _Nt_array_ptr and update its
+     // bounds in ObservedBounds.
+
+     for (const auto item : WidenedBounds) {
+       const VarDecl *V = item.first;
+       unsigned Offset = item.second;
+
+       // We normalize the declared bounds to RangBoundsExpr here so that we
+       // can easily apply the offset to the upper bound.
+       BoundsExpr *Bounds = S.ExpandBoundsToRange(V, V->getBoundsExpr());
```
dtarditi, line 2173: Here too we may be allocating AST data structures multiple times.  We want to avoid doing that, and do the expansion only once.

mgrang:
This PR depends on https://github.com/microsoft/checkedc/pull/400

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2142,6 +2142,56 @@ namespace {
...
+     auto I = KilledBounds.find(St);
+     if (I == KilledBounds.end())
+       return;
+
+     // KilledBounds stores a mapping of statements to all variables whose
+     // bounds are killed by each statement. Here we reset the bounds of all
+     // variables killed by the statement S to the declared bounds.
+     for (const VarDecl *V : I->second) {
+       if (const BoundsExpr *Bounds = V->getBoundsExpr())
+         ObservedBounds[V] = S.ExpandBoundsToRange(V, Bounds);
```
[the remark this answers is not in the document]
mgrang: Created issue to track this: https://github.com/microsoft/checkedc-clang/issues/830

PR #824, opened 2020-04-24
[WIP] Bounds context: update after assignments [2/n]

kkjeer:
This PR updates the bounds context mapping variables to their observed bounds after assignments to variables. It is currently a work in progress meant to show the proposed updates to the bounds context handling, which is relevant to the bounds widening work in #821. This PR also includes the changes in #820, which makes changes to UEQ handling that are necessary in order for this bounds context updating work.

#### Updates to bounds context handling
The UC bounds context has been replaced with DeclaredBounds and ObservedBounds contexts in CheckingState. DeclaredBounds maps variable declarations to the bounds the programmer has declared for the variable. These bounds should never be affected by assignments or bounds widening. ObservedBounds maps variable declarations to the bounds that are inferred during bounds checking. These bounds are affected by assignments, and may be affected by bounds widening.

Previously, CheckDeclRefExpr used the CheckingState.UC bounds context to get the target bounds of a variable. This allowed the widened bounds to be used as the observed rvalue bounds of the value of a variable, since CheckCastExpr returns the target bounds of the subexpression of an LValueToRValue cast as the rvalue bounds. This is no longer the case in CheckDeclRefExpr. The reason for this is that lvalue target bounds should never be affected by assignments or bounds widening. As a result, the observed rvalue bounds for the value of a variable that has widened bounds need to be inferred by another method rather than modifying CheckDeclRefExpr.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -2610,7 +2610,7 @@ namespace {
 
         // Update UEQ and G for assignments to `e1` where `e1` is a variable.
         if (DeclRefExpr *V = GetLValueVariable(LHS)) {
-          Expr *OV = GetOriginalValue(V, Src, State.UEQ);
+          Expr *OV = GetOriginalValue(V, Target, Src, State.UEQ);
           UpdateAfterAssignment(V, Target, OV, CSS, State, State);
         }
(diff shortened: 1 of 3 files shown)


kkjeer:
Closing in favor of #827

PR #827, opened 2020-04-30
Bounds context: observed bounds [2/n]

kkjeer:
This PR replaces the UC bounds context member of CheckingState with ObservedBounds, and uses the observed bounds as the rvalue bounds for the value of a variable, rather than as the lvalue target bounds for a variable.

#### Context:
This is part of the ongoing work to improve the soundness of bounds checking involving assignments to variables. This work is also related to the bounds widening analysis (e.g. #821). After this PR is merged, and the ObservedBounds context is updated to reflect the widened bounds of nt_array_ptrs, the bounds checking can use the widened bounds as the rvalue bounds of the value of an nt_array_ptr variable.

#### Notable changes:
* Rename the UC bounds context to ObservedBounds. This makes the meaning of the bounds expressions in the bounds context more clear: the observed bounds are the bounds that are currently known to be valid for a variable. These bounds are updated during bounds checking and will later be used to verify that the observed bounds imply the declared bounds for all variables after an assignment.
* Do not use the observed bounds of a variable as the lvalue target bounds. The target bounds of a variable should always be the programmer-declared bounds.
* Use the observed bounds of a variable as the rvalue bounds by modifying RValueCastBounds for LValueToRValue casts.

#### Tests:
* Modified the bounds-context.c tests to reflect changes in how the bounds context is printed.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -703,10 +703,9 @@ namespace {
           return true;
         if (D->isInvalidDecl())
           return true;
-        const BoundsExpr *Bounds = D->getBoundsExpr();
         // The bounds expressions in the bounds context should be normalized
         // to range bounds.
-        if (Bounds)
+        if (const BoundsExpr *Bounds = D->getBoundsExpr())
           BoundsContextRef[D] = SemaRef.ExpandBoundsToRange(D, Bounds);
         return true;
       }
@@ -4019,7 +4018,8 @@ namespace {
     // Taking the intersection of the observed bounds contexts of the block's
     // predecessors ensures that, before checking a statement S in the block,
     // the block's observed bounds context contains only variables with bounds
-    // that are in scope at S.
+    // that are in scope at S.  At the beginning of the block, each variable in
+    // scope is mapped to its normalized declared bounds.
     CheckingState GetIncomingBlockState(const CFGBlock *Block,
                                         llvm::DenseMap<unsigned int, CheckingState> BlockStates) {
       CheckingState BlockState;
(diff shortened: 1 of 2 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -841,13 +845,13 @@ namespace {
 
         OS << "{\n";
         for (auto I = OrderedDecls.begin(); I != OrderedDecls.end(); ++I) {
-          DeclaratorDecl *Variable = *I;
-          if (!UC[Variable])
+          const VarDecl *Variable = *I;
+          if (!Context[Variable])
```
mgrang, line 849: I think this would insert the Variable in Context if it does not exist.

If you only want to test existence of key in a map please use this:
`if (Context.count(Variable))`

If you want to test existence *and* use the value, then use this:
```
auto it = Context.find(Variable);
if (it != Context.end()) {
  // use *it
}
```

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3976,47 +3994,53 @@ namespace {
...
-    BoundsContextTy IntersectUC(BoundsContextTy UC1, BoundsContextTy UC2) {
-      BoundsContextTy IntersectedUC;
-      for (auto Pair : UC1) {
-        if (!Pair.second || !UC2[Pair.first])
+    // intersection of the contexts Context1 and Context2.
+    BoundsContextTy IntersectBoundsContexts(BoundsContextTy Context1,
+                                            BoundsContextTy Context2) {
+      BoundsContextTy IntersectedContext;
+      for (auto Pair : Context1) {
+        if (!Pair.second || !Context2[Pair.first])
```
mgrang, line 4039: Same comment here about checking for existence of key in map.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3976,47 +3994,53 @@ namespace {
...
     }
 
     // IntersectUC returns a bounds context resulting from taking the
-    // intersection of the contexts UC1 and UC2.
-    BoundsContextTy IntersectUC(BoundsContextTy UC1, BoundsContextTy UC2) {
-      BoundsContextTy IntersectedUC;
-      for (auto Pair : UC1) {
-        if (!Pair.second || !UC2[Pair.first])
+    // intersection of the contexts Context1 and Context2.
+    BoundsContextTy IntersectBoundsContexts(BoundsContextTy Context1,
```
mgrang, line 4035: Can this be a templated function which can be used to intersect sets holding items of any type? See https://github.com/microsoft/checkedc-clang/blob/master/clang/lib/Sema/BoundsAnalysis.cpp#L686.

Going forward we should move these intersect, union, etc functions to a common file so that we can re-use them.

mgrang:
The names UEQ, G, OV, etc sound very cryptic. Is it possible to rename them to more descriptive names in a separate PR?

clang/lib/Sema/SemaBounds.cpp
kkjeer: I updated IntersectBoundsContext to reset each in-scope variable to its normalized declared bounds. Now that it's more specialized to rely on variable declarations and bounds expressions, I don't think it lends itself as well to templating.

kkjeer:
> The names UEQ, G, OV, etc sound very cryptic. Is it possible to rename them to more descriptive names in a separate PR?

Created #828 to track this.

PR #834, opened 2020-05-12
Equivalent expressions: bounds cast inverse [7/n]

kkjeer:
This PR extends the IsInvertible and Inverse methods to return the inverse of a bounds cast expression.

#### Context:
This work is part of the ongoing work for soundness of bounds declaration checking. In order to update the bounds context after assignments that involve bounds cast expressions, IsInvertible and Inverse need to be able to handle bounds casts.

Consider:
```
void bounds_casts_correct(array_ptr<int> a : count(1), array_ptr<char> b : count(4)) {
  // Source bounds for dynamic cast: (temp(a), temp(a) + 1)
  // Original value of a: a
  // Updated observed bounds for a: (temp(a), temp(a) + 1)
  a = _Dynamic_bounds_cast<array_ptr<int>>(a, count(1));

  // Source bounds of dynamic bounds cast are (a, a + 1)
  // Original value of a: a
  // Updated observed bounds for a: bounds(a, a + 1)
  a = _Dynamic_bounds_cast<array_ptr<int>>(a, bounds(a, a + 1));
}

void bounds_casts_incorrect(array_ptr<int> a : count(1), array_ptr<char> b : count(4)) {
  // Source bounds for dynamic cast: (temp(a), temp(a) + 1)
  // Original value of a: null
  // Updated observed bounds for a: (temp(a), temp(a) + 1)
  a = _Dynamic_bounds_cast<array_ptr<int>>(a, count(1));

  // Source bounds of dynamic bounds cast are (a, a + 1)
  // Original value of a: null
  // Updated observed bounds for a: bounds(unknown)
  a = _Dynamic_bounds_cast<array_ptr<int>>(a, bounds(a, a + 1));
}
```
In the incorrect case, where the original value of `a` is `null`, the observed bounds for `a` will become `bounds(unknown)` after the second assignment. In the correct case (as implemented by this PR), where the original value of `a` is `a`, the observed bounds for `a` will become `bounds(a, a + 1)` after the second assignment. This matches the behavior of the first assignment.

#### Major changes:
* Update IsInvertible to determine that a bounds cast expression `BoundsCast<T1>(temp(e1))` is invertible with respect to a variable `x` if and only if `e1` is invertible with respect to `x`
* Update Inverse to return `(T2)e1` as the inverse of a bounds cast expression `BoundsCast<T1>(temp(e1))`, where `e1` has type `T2`.

#### Testing;
* Add tests for bounds cast inverses in equiv-expr-sets.c

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -4091,7 +4091,7 @@ namespace {
         case CastKind::CK_BooleanToSignedIntegral:
         case CastKind::CK_IntegralToFloating:
           return IsInvertible(X, E->getSubExpr());
-        // Bounds casts may be invertible
+        // Bounds casts may be invertible.
         case CastKind::CK_DynamicPtrBounds:
         case CastKind::CK_AssumePtrBounds: {
           CHKCBindTemporaryExpr *Temp =
@@ -4104,7 +4104,7 @@ namespace {
         case CastKind::CK_PointerToIntegral:
         case CastKind::CK_IntegralCast:
           return Size1 >= Size2 && IsInvertible(X, E->getSubExpr());
-        // All other casts are considered narrowing
+        // All other casts are considered narrowing.
         default:
           return false;
       }
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -3910,6 +3910,7 @@ namespace {
           return IsBinaryOperatorInvertible(X, cast<BinaryOperator>(E));
         case Expr::ImplicitCastExprClass:
         case Expr::CStyleCastExprClass:
+        case Expr::BoundsCastExprClass:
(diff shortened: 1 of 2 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4088,6 +4091,14 @@ namespace {
         case CastKind::CK_BooleanToSignedIntegral:
         case CastKind::CK_IntegralToFloating:
           return IsInvertible(X, E->getSubExpr());
+        // Bounds casts may be invertible
```
mgrang, line 4094: nit: period needed at the end of comment.

kkjeer:
> Looks good, thanks! Please address the feedback pointed out by @mgrang. Please also make a PR to update the Checked C spec to cover this case.

Created checked-c issue [401](https://github.com/microsoft/checkedc/issues/401) to track the spec updates.

PR #835, opened 2020-05-13
Add bigobj compile flag to checked-c-convert RewriteUtils.cpp

kkjeer:
This PR fixes a build failure with checked-c-convert by adding the bigobj compile flag to the CMake file for RewriteUtils.cpp.

Build failure message:
clang\tools\checked-c-convert\RewriteUtils.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj

Testing:
* Passed manual testing on Windows X64
* Passed automated testing on Windows X64 and Linux
* Automated testing in progress on Windows X86


mgrang:
Thanks for fixing this. LGTM.

PR #836, opened 2020-05-15
Bounds context: update after assignments [3/n]

kkjeer:
This PR updates the context mapping variable declarations to their current known bounds after an assignment to a variable.

#### Context:
This work is part of the soundness of bounds declaration checking. After this PR is merged, the next step is to validate that each variable declaration's observed bounds implies the variable's declared bounds after checking a top-level CFG statement.

#### Notable changes:
* Modify UpdateAfterAssignment to update the observed bounds context after an assignment to a variable
* Modify UpdateAfterAssignment to return updated rvalue bounds for the source of an assignment
* Use the modified source bounds as the rvalue bounds of an assignment or increment/decrement operator. For example, the rvalue bounds of the assignment `p = p + 2`, if `p` has declared bounds `(p, p + 1)` will be (`p - 2, p - 2 + 1)`, since the original value of `p` is `p - 2`.
* Check that the modified source bounds of an increment/decrement operator on a variable imply the declared bounds for the variable. For example, if `p` has declared bounds `(p, p + 1)`, after `p++`, check that the updated result bounds `(p - 1, p - 1 + 1)` imply the declared bounds `(p, p + 1)`.
* Account for equality information from nested assignments when checking bounds declarations at assignments, initializers, and call arguments.

#### Implications for pointer arithmetic:
One important change in this PR is that the source bounds used to check an assignment are the updated observed bounds. This means that pointer arithmetic on variables with declared bounds will frequently result in bounds warnings or errors.

For example, these assignments will all result in a warning:
* cannot prove declared bounds for p are valid after assignment 
* (expanded) declared bounds: bounds(p, p + i)
* (expanded) inferred bounds: bounds(p - 1, p - 1 + i)
```
void f(array_ptr<int> p : bounds(p, p + i), int i) {
  // Original value of p: p - 1
  // Updated source bounds: (p - 1, (p - 1) + i)
  // Updated source bounds are also the result bounds for the assignment
  p = p + 1;
  p += 1;
  p++;
}
```

These assignments will all result in an error:
* declared bounds for p are invalid after assignment
* (expanded) declared bounds are bounds(p, p)
* (expanded) inferred bounds are bounds(p - 1, p - 1)
* source bounds are an empty range
* destination upper bound is above source upper bound
```
void f(array_ptr<int> p : bounds(p, p)) {
  // Original value of p: p - 1
  // Updated source bounds: (p - 1, p - 1)
  // Updated source bounds are also the result bounds for the assignment
  p = p + 1;
  p += 1;
  p++;
}
```

#### Testing:
* Updated bounds-context.c to test updating observed bounds
* Modified checked-c tests to account for new warnings and errors when checking the updated source bounds of an assignment (checkedc PR [402](https://github.com/microsoft/checkedc/pull/402))
* Passed manual testing on Windows X64


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4433,10 +4542,16 @@ namespace {
       return false;
     }
 
-    // If E is a (possibly parenthesized) lvalue variable V,
+    // If E is a possibly parenthesized lvalue variable V,
     // GetLValueVariable returns V. Otherwise, it returns nullptr.
+    //
+    // V may have value-preserving operations applied to it.  For example,
```
dtarditi, line 4548: It would be helpful in the comment to describe what value-preserving casts can be applied to lvalue expressions.

clang/test/CheckedC/inferred-bounds/widened-bounds-check.c
```diff
@@ -13,7 +13,7 @@ void f1() {
   {}
 
   if (*p) {
-    p++;
+    p++; // expected-error {{declared bounds for p are invalid after assignment}}
```
dtarditi, line 16: This message could confuse programmers because they might think you are referring to an assignment operator.   There is no assignment operator in that expression.   You might want a message that says increment or decrement instead.  It is fine to address this in a follow-up PR.

clang/test/CheckedC/inferred-bounds/bounds-context.c
```diff
@@ -265,12 +280,1856 @@ void f2(int flag, int x, int y) {
...
+  // CHECK-NEXT: }
+}
+
+////////////////////////////////////////////////////////////////////
+// Setting the observed bounds of a variable to the source bounds //
+////////////////////////////////////////////////////////////////////
+
+// Scalar-typed variable declarations (array_ptr, nt_array_ptr) set the observed bounds to the initializer bounds
+void source_bounds1(array_ptr<int> a: count(1)) {
+  // Observed bounds context before declaration: { a => bounds(a, a + 1), arr => bounds(arr, arr + 0) }
```
dtarditi, line 666: Should `arr` be in the observed bounds context *before* the declaration?

I don't think this test is checking that bounds context before the declaration, also.

clang/test/CheckedC/inferred-bounds/bounds-context.c
```diff
@@ -265,12 +280,1856 @@ void f2(int flag, int x, int y) {
...
+  // CHECK-NEXT: RangeBoundsExpr
+  // CHECK-NEXT:   ImplicitCastExpr {{.*}} <LValueToRValue>
+  // CHECK-NEXT:     DeclRefExpr {{.*}} 'a'
+  // CHECK-NEXT:   BinaryOperator {{.*}} '+'
+  // CHECK-NEXT:     ImplicitCastExpr {{.*}} <LValueToRValue>
+  // CHECK-NEXT:       DeclRefExpr {{.*}} 'a'
+  // CHECK-NEXT:     IntegerLiteral {{.*}} 1
+  // CHECK-NEXT: }
+
+  // Observed bounds context before declaration: { a => bounds(a, a + 1), arr => bounds(arr, arr + 0), buf => bounds(buf, buf + 2) }
```
dtarditi, line 705: Should `buf` be in the observed bound context before the declaration?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4685,11 +4800,20 @@ namespace {
     // CreateIntegerLiteral returns an integer literal with Ty type.
     // If Ty denotes a pointer to an integer type (char *, ptr<int>, etc.),
     // CreateIntegerLiteral returns an integer literal with Ty's pointee type.
+    // If Ty denotes a pointer to a non-integer type (float *, ptr<double>,
```
dtarditi, line 4803: I found this comment and code confusing.   Are you overloading creation of the integer literal to account for the different bitwidth of integer literals required for integer arithmetic vs. pointer arithmetic involving a literal.  For pointer arithmetic, if the pointee type is an integer type, why would you make the type of the integer literal match the pointee type?  An example, for "char *", it is valid to add an integer larger than a char to the pointer.

clang/test/CheckedC/inferred-bounds/widened-bounds-check.c
kkjeer: Created issue #844 to track this.

clang/test/CheckedC/inferred-bounds/bounds-context.c
```diff
@@ -67,218 +77,2061 @@ void f1(array_ptr<int> arr : count(len), int len, int size) {
...
+  // CHECK-NEXT:       DeclRefExpr {{.*}} 'b'
+  // CHECK-NEXT:     ImplicitCastExpr {{.*}} <LValueToRValue>
+  // CHECK-NEXT:       DeclRefExpr {{.*}} 'size'
+  // CHECK-NEXT: }
+}
+
+// If statement, redeclared variable
+void declared2(int flag, int x, int y) {
+  // Observed bounds context: { a => bounds(a, a + x) }
+  int a checked[] : count(x) = (int checked[]){ 0 };
```
dtarditi, line 120: The observed bounds declaration for an array variable should be based on the declared type of the array variable.  It seems to be based on the declared bounds declaration here, which is incorrect.  I checked the code in SemaBounds.cpp.    Handling of this case seems to be missing, and there is a even a "TODO" that hints at this.   

Could you open a new issue for this case?   I think this should be handled as a separate change.   You should not try to fix it in this PR.  There's no reason for this to block further.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3288,7 +3351,7 @@ namespace {
         } else {
           BoundsExpr *NormalizedDeclaredBounds = ExpandToRange(D, DeclaredBounds);
```
dtarditi, line 3352: I believe the observed bounds need to be set in this case.   Please open an issue to track this, as a separate follow-up.  Addressing this case I believe will handled the TODO that is just above (at line 3273).

clang/test/CheckedC/inferred-bounds/bounds-context.c
kkjeer: Created #862 to track this.

Issue #818, opened 2020-04-09
Allow bounds widening for the default case of a switch-case


mgrang:
We need to allow bounds widening for the default case of a switch-case in some circumstances. If we establish that another label in a switch statement tests for 0, then the default case will handle non-zero case, and the bounds can be widened there. Programmers will use all the different forms of case statements and may depend on that behavior.

mgrang:
PR https://github.com/microsoft/checkedc-clang/pull/838

mgrang:
This is now fixed in master.

PR #838, opened 2020-05-21
Allow bounds widening for the default case of a switch-case

mgrang:
If we establish that another label in a switch statement tests for 0, then the
default case will handle non-zero case, and the bounds can be widened there.
Programmers will use all the different forms of case statements and may depend
on that behavior.

See https://github.com/microsoft/checkedc-clang/issues/818

### DIFF
--- a/clang/lib/Sema/BoundsAnalysis.cpp
+++ b/clang/lib/Sema/BoundsAnalysis.cpp
@@ -150,8 +150,8 @@ void BoundsAnalysis::ComputeGenSets() {
       if (pred->succ_size() == 0)
         continue;
 
-      const CFGBlock *FalseDefaultBlock = *(pred->succs().end() - 1);
-      if (EB->Block == FalseDefaultBlock)
+      const CFGBlock *FalseOrDefaultBlock = *(pred->succs().end() - 1);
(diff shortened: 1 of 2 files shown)


clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -147,34 +150,47 @@ void BoundsAnalysis::ComputeGenSets() {
...
+        // expression of a switch shall have integer type.
+        // If we have switch(*p) where p is _Nt_array_ptr<char> then it is
+        // casted to integer type and an IntegralCast is generated. Here we
+        // strip off the IntegralCast.
+        if (auto *CE = dyn_cast<CastExpr>(E)) {
+          if (CE->getCastKind() == CastKind::CK_IntegralCast)
+            E = CE->getSubExpr();
+        }
+
+        // If the current block has a non-null case label, we cannot widen the
```
kkjeer, line 179: The comment says "If the current block has a non-null case label", but the statement checks whether the case is null. Should the comment say "If the current block has a null case label"?

mgrang:
Windows ADO jobs keep getting stuck whenever triggered manually. I would go ahead and commit this and have the nightly Windows runs catch any breakages, if any.

PR #840, opened 2020-05-29
Inverse of address-of array subscripts

kkjeer:
This PR extends the work done in #839 by implementing the inverse of address-of array subscript expressions. Since `e1[e2]` is equivalent to `*(e1 + e2)`, the inverse of `&e1[e2]` is the inverse of `e1 + e2`. The inverse of `&*(e1 + e2)` is also the inverse of `e1 + e2`.

#### Testing:
* Added tests for the inverse of address-of array subscript expressions to equiv-expr-sets.c
* Passed automated testing on Windows and Linux

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -4057,11 +4057,15 @@ namespace {
         // &e1[e2] is invertible with respect to x if e1 + e2 is invertible
         // with respect to x.
         else if (ArraySubscriptExpr *ArraySubExpr = dyn_cast<ArraySubscriptExpr>(SubExpr)) {
-          BinaryOperator *Sum = ExprCreatorUtil::CreateBinaryOperator(S,
-                                  ArraySubExpr->getBase(),
-                                  ArraySubExpr->getIdx(),
-                                  BinaryOperatorKind::BO_Add);
-          return IsInvertible(X, Sum);
+          Expr *Base = ArraySubExpr->getBase();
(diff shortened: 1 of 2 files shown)


mgrang:
LGTM.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4197,12 +4204,18 @@ namespace {
...
-      // Inverse(f, &*e1) = Inverse(f, e1)
       if (Op == UnaryOperatorKind::UO_AddrOf) {
+        // Inverse(f, &*e1) = Inverse(f, e1)
         if (UnaryOperator *UnarySubExpr = dyn_cast<UnaryOperator>(SubExpr)) {
           if (UnarySubExpr->getOpcode() == UnaryOperatorKind::UO_Deref)
             return Inverse(X, F, UnarySubExpr->getSubExpr());
         }
+        // Inverse(f, &e1[e2]) = Inverse(f, e1 + e2)
+        else if (ArraySubscriptExpr *ArraySubExpr = dyn_cast<ArraySubscriptExpr>(SubExpr))
+          return Inverse(X, F, ExprCreatorUtil::CreateBinaryOperator(S,
```
dtarditi, line 4215: This allocates a new binary operator expression in the AST context, which causes the expression to live forever.   We want to avoid analysis routines allocating indefinitely-lived data when possible.  You might want to allocate the binary operator in a local variable here (that is, on the stack) instead.

PR #841, opened 2020-06-02
Save normalized VarDecl bounds

### LINKED ISSUE #830: Avoid repeated calls to ExpandBoundsToRange
Throughout clang in general (and inside dataflow analysis in particular) we repeatedly invoke ExpandBoundsToRange in order to canonicalize the bounds of a variable to RangeBoundsExpr. Sometimes we do this multiple times for the same variable. This is very inefficient because ExpandBoundsToRange can allocate AST data structures that are permanently allocated and increase the memory usage of the compiler.

The solution is to canonicalize the bounds once and attach it to the VarDecl.

kkjeer:
Fixes #830 

This PR attaches a normalized bounds expression to a VarDecl to avoid repeated calls to ExpandBoundsToRange, which can allocate AST structures and increase the memory usage of the compiler. This is relevant for updating and validating the observed context and using widened bounds, which is currently recomputing the normalized bounds of a VarDecl.

#### Testing:
* Passed automated testing on Linux

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -5543,6 +5543,22 @@ void Sema::WarnDynamicCheckAlwaysFails(const Expr *Condition) {
   }
 }
 
+// If the VarDecl D has a byte_count or count bounds expression,
+// NormalizeBounds expands it to a range bounds expression.  The expanded
+// range bounds are attached to the VarDecl D to avoid recomputing the
+// normalized bounds for D.
+BoundsExpr *Sema::NormalizeBounds(const VarDecl *D) {
+  // If D already has a normalized bounds expression, do not recompute it.
+  if (const BoundsExpr *NormalizedBounds = D->getNormalizedBounds()) 
+    return const_cast<BoundsExpr *>(NormalizedBounds);
+
+  // Normalize the bounds of D to a RangeBoundsExpr and attach the normalized
+  // bounds to D to avoid recomputing them.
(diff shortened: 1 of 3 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -5543,6 +5523,22 @@ void Sema::WarnDynamicCheckAlwaysFails(const Expr *Condition) {
   }
 }
 
+// If the VarDecl D has a byte_count or count bounds expression,
+// NormalizeBounds expands it to a range bounds expression.  The expanded
+// range bounds are attached to the VarDecl D to avoid recomputing the
+// normalized bounds for D.
+BoundsExpr *Sema::NormalizeBounds(const VarDecl *D) {
+  // If D already has a normalized bounds expression, do not recompute it.
+  if (const BoundsExpr *NormalizedBounds = D->getNormalizedBounds())
```
mgrang, line 5532: `setNormalizedBounds` already set non-const `BoundsExpr`. So I guess you do not need the const_cast here.

kkjeer:
> Do you plan to cleanup the invocations of ExpandBoundsToRange in other files as a follow up?

Created #848 to track this.

clang/lib/Sema/SemaBounds.cpp
kkjeer: Since the `VarDecl D` is `const`, `D->getNormalizedBounds()` seems to return a `const BoundsExpr`. I could put the `const_cast` on `D` instead.

kkjeer: As discussed offline, I removed the `const` from `VarDecl::getNormalizedBounds`.

PR #842, opened 2020-06-02
Map top-level statements to variables with killed widened bounds

kkjeer:
This PR adjusts bounds widening to map top-level CFG statements to the set of nt_array_ptr variables whose widened bounds are killed by the statement.

This is applicable to statements with multiple assignments. For example:
```
void f(nt_array_ptr<char> p : count(i), int i, int unrelated) {
  if (p[i])) {
    // p has widened bounds (p, p + i + 1).

    // This statement kills the widened bounds of p.
    // After this statement, p should have observed bounds (p, p + i).
    i = 0, unrelated = 0;
  }
}
```

Bounds declaration checking (CheckBoundsDeclarations::TraverseCFG) only checks top-level statements that are not nested within other top-level statements. The statement `i = 0, unrelated = 0;` will be checked, but `i = 0` and `unrelated = 0` will not be checked. In order for bounds checking to detect that the widened bounds of `p` have been killed, and that the observed bounds of `p` should be reset to the expanded declared bounds `(p, p + i)`, the killed bounds should map `i = 0, unrelated = 0` to `p`.

#### Context:
This work is related to validating the updated observed bounds context against the declared bounds after each top-level statement. In order to correctly detect errors in bounds checking, the bounds checker needs to know when a variable no longer has widened bounds.

#### Testing:
* Added tests to bounds-context.c to test killing widened bounds after statements with multiple assignments.

### DIFF
--- a/clang/include/clang/Sema/BoundsAnalysis.h
+++ b/clang/include/clang/Sema/BoundsAnalysis.h
@@ -97,7 +97,7 @@ namespace clang {
   using StmtDeclSetTy = llvm::DenseMap<const Stmt *, DeclSetTy>;
 
   // StmtSet denotes a set of Stmts.
-  typedef llvm::SmallPtrSet<const Stmt *, 16> StmtSet;
+  using StmtSet = llvm::SmallPtrSet<const Stmt *, 16>;
 
   class BoundsAnalysis {
   private:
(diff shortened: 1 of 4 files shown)


clang/include/clang/Sema/BoundsAnalysis.h
```diff
@@ -96,6 +96,9 @@ namespace clang {
   // 2. any variable used in the bounds expr of V is assigned to in S.
   using StmtDeclSetTy = llvm::DenseMap<const Stmt *, DeclSetTy>;
 
+  // StmtSet denotes a set of Stmts.
+  typedef llvm::SmallPtrSet<const Stmt *, 16> StmtSet;
```
mgrang, line 100: Please use `using` instead of `typedef` for consistency in this file.

PR #850, opened 2020-06-09
[WIP] Compare expressions semantically

mgrang:
Compare expressions E1 and E2 to determine whether they are “equal”. Two expressions are equal if they evaluate either to the same value or the same resultant expression. We make use of an n-ary preorder AST to compare the expressions.


mgrang:
Closing in favor of https://github.com/microsoft/checkedc-clang/pull/858

PR #857, opened 2020-06-16
Record equality between target and source

### LINKED ISSUE #845: Record equality between target and source for checking bounds
After the updated bounds context is used to check bounds for variables, there will be an issue with the way that expression equality is recorded. Certain kinds of expressions (expressions that create a new object or read memory via a pointer, such as `"abc"`, `*p`, `s->f`, etc.) are not included in the set State.EquivExprs of expressions that produce the same value. This introduces proof failures when checking bounds. Consider:

```
void missing_equality(array_ptr<int> a : count(1)) {
  // s->f reads memory via a pointer
  // s->f is not included in State.EquivExprs
  // Bounds checking does not know that a and s->f produce the same value
  // The inferred bounds (s->f, s->f + 1) do not provably imply
  // the declared bounds (a, a + 1)
  a = s->f;
}
```
When checking that the inferred bounds of `a` imply the declared bounds of `a`, equality between `a` and `s->f` should be temporarily recorded so that the bounds checker is able to prove that the inferred bounds imply the declared bounds.

kkjeer:
Fixes #845 

This PR fixes regressions introduced by #843 where bounds checking warnings would be emitted for assignments to variables where equality between the target variable and the assignment source was not recorded. For example:

```
struct S {
  array_ptr<int> f : count(2);
};

void test(array_ptr<int> p : count(2), struct S *s) {
  // Observed bounds of p are (s->f, s->f + 2)
  // Declared bounds of p are (p, p + 2)
  // Equality between p and s->f is not recorded => bounds warning
  p = s->f;
}
```

Since `s->f` reads memory via a pointer, it is not included in the checking state set of sets of equivalent expressions. When validating the bounds context, if the bounds checker does not regard `p` and `s->f` as equivalent, it is unable to prove that the observed bounds `(s->f, s->f + 2)` imply the declared bounds `(p, p + 2)`.

This PR introduces a TargetSrcEquality map in the CheckingState class that records equality between a target variable and a source expression that is not included in the checking state EquivExprs. In ValidateBoundsContext, this map is used to construct a local set of sets of equivalent expressions that includes all equality information in the checking state EquivExprs set, as well as each mapped pair `<Target, Src>` in the TargetSrcEquality map. This prevents expressions such as `*p`, `arr[i]`, `s.f`, `s->f`, `"abcde"`, `{ 0, 1, 2 }`, etc. from being included in the checking state EquivExprs, while allowing them to inform the bounds validation logic about equality implied by initialization or assignment.


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3812,6 +3821,29 @@ namespace {
...
     void ValidateBoundsContext(Stmt *S, CheckingState State, CheckedScopeSpecifier CSS) {
+      // Construct a set of sets of equivalent expressions that contains all
+      // the equality facts in State.EquivExprs, as well as any equality facts
+      // implied by State.TargetSrcEquality.  These equality facts will only
+      // be used to validate the bounds context and will not persist across
+      // CFG statements.  The source expressions in State.TargetSrcEquality
+      // do not meet the criteria for persistent inclusion in State.EquivExprs:
+      // for example, they may create new objects or read memory via pointers.
+      EquivExprSets EquivExprs = State.EquivExprs;
+      for (auto Pair : State.TargetSrcEquality) {
```
mgrang, line 3832: Instead of copying `State.TargetSrcEquality` in the `for` loop, you want to create a reference to it:
`for (auto &Pair : State.TargetSrcEquality)`

PR #858, opened 2020-06-17
Compare Expressions: Preoder AST to semantically compare expressions for equality [1/n]

mgrang:
We implement a preoder AST data structure to compare two expressions.
This AST is a mirror of the clang AST with the following properties:
-	It is an n-ary tree
-	It represents an expression in the preorder form
-	Each node of the clang AST is a node in the preorder AST
-	Its internal nodes represent the operator nodes of the clang AST
-	Its leaf nodes represent the operands of the clang AST

We traverse the ASTs for the expressions E1 and E2 and create the respective
preorder AST’s. Then we perform tree transforms to "normalize" the expressions
and finally match the tree node-by-node to determine if E1 and E2 are equal.

### DIFF
--- a/clang/include/clang/AST/CanonBounds.h
+++ b/clang/include/clang/AST/CanonBounds.h
@@ -119,7 +119,8 @@ namespace clang {
     /// bounds expressions.
     Result CompareExpr(const Expr *E1, const Expr *E2);
     /// \brief Semantic comparison of expressions that can occur in
-    /// bounds expressions.
+    /// bounds expressions. A return value of true indicates that the two
+    /// expressions are equivalent semantically.
     bool CompareExprSemantically(const Expr *E1, const Expr *E2);
 
     /// \brief Compare declarations that may be used by expressions or
--- a/clang/include/clang/AST/CanonBounds.h
+++ b/clang/include/clang/AST/CanonBounds.h
@@ -120,7 +120,7 @@ namespace clang {
     Result CompareExpr(const Expr *E1, const Expr *E2);
     /// \brief Semantic comparison of expressions that can occur in
     /// bounds expressions.
-    Result CompareExprSemantically(const Expr *E1, const Expr *E2);
+    bool CompareExprSemantically(const Expr *E1, const Expr *E2);
 
     /// \brief Compare declarations that may be used by expressions or
     /// or types.
--- a/clang/include/clang/AST/CanonBounds.h
+++ b/clang/include/clang/AST/CanonBounds.h
@@ -50,8 +50,7 @@ namespace clang {
     enum class Result {
       LessThan,
       Equal,
-      GreaterThan,
-      NotEqual,
+      GreaterThan
     };
 
   private:
--- a/clang/include/clang/AST/CanonBounds.h
+++ b/clang/include/clang/AST/CanonBounds.h
@@ -52,7 +52,6 @@ namespace clang {
       Equal,
       GreaterThan,
       NotEqual,
-      Unknown,
     };
 
   private:
--- a/clang/include/clang/AST/CanonBounds.h
+++ b/clang/include/clang/AST/CanonBounds.h
@@ -50,7 +50,9 @@ namespace clang {
     enum class Result {
       LessThan,
       Equal,
-      GreaterThan
+      GreaterThan,
+      NotEqual,
+      Unknown,
     };
 
   private:
@@ -118,6 +120,9 @@ namespace clang {
     /// \brief Lexicographic comparison of expressions that can occur in
     /// bounds expressions.
     Result CompareExpr(const Expr *E1, const Expr *E2);
+    /// \brief Semantic comparison of expressions that can occur in
+    /// bounds expressions.
+    Result CompareExprSemantically(const Expr *E1, const Expr *E2);
 
     /// \brief Compare declarations that may be used by expressions or
     /// or types.
--- a/clang/include/clang/AST/PreorderAST.h
+++ b/clang/include/clang/AST/PreorderAST.h
@@ -22,6 +22,8 @@
 
 namespace clang {
 
+  using Result = Lexicographic::Result;
+
   // Each binary operator of an expression results in a new node of the
   // PreorderAST. Each node contains 3 fields:
   // Opc: The opcode of the operator.
@@ -110,7 +112,9 @@ namespace clang {
     // @return Returns a bool indicating whether the two ASTs are equal.
     bool IsEqual(PreorderAST &P) { return IsEqual(Root, P.Root); }
 
-    // Check if an error has occurred during transformation of the AST.
+    // Check if an error has occurred during transformation of the AST. This
+    // is intended to be called from outside this class to check if an error
+    // has occurred during comparison of expressions.
     // @return Whether an error has occurred or not.
     bool GetError() { return Error; }
 
--- a/clang/include/clang/AST/PreorderAST.h
+++ b/clang/include/clang/AST/PreorderAST.h
@@ -115,6 +115,12 @@ namespace clang {
     // @param[in] N is the root node of the AST.
     void Cleanup(ASTNode *N);
 
+    // Set Error in case an error occurs during transformation of the AST.
+    // @param[in] Err is the value to be set for the Error field.
+    void SetError(bool Err) {
+      Error = Err;
+    }
+
     // Print the preorder AST.
     // @param[in] N is the root node of the AST.
     void PrettyPrint(ASTNode *N);
@@ -143,12 +149,6 @@ namespace clang {
     // @return Whether an error has occurred or not.
     bool GetError() {
       return Error;
-    }
-
-    // Set Error in case an error occurs during transformation of the AST.
-    // @param[in] Err is the value to be set for the Error field.
-    void SetError(bool Err) {
-      Error = Err;
     }
 
     // Compare the current AST with the given AST. This in turn, invokes
--- a/clang/include/clang/AST/PreorderAST.h
+++ b/clang/include/clang/AST/PreorderAST.h
@@ -1,4 +1,4 @@
-//===------- PreorderAST.h: An n-ary preorder abstract syntax tree -------===//
+//===------r PreorderAST.h: An n-ary preorder abstract syntax tree -------===//
 //
 //                     The LLVM Compiler Infrastructure
 //
@@ -95,17 +95,15 @@ namespace clang {
     void Create(ASTNode *N, Expr *E, ASTNode *Parent = nullptr);
 
     // Normalize the input expression through a series of transforms on the
-    // preorder AST.
-    // @param[in] N is the root of the AST.
-    // @param[out] Error is populated if an error is encountered during
+    // preorder AST. The Error field is set if an error is encountered during
     // normalization.
-    void Normalize(ASTNode *N, bool &Error);
-
-    // Sort the variables in a node of the AST.
     // @param[in] N is the root of the AST.
-    // @param[out] Error is populated if an error is encountered during
-    // sorting.
-    void Sort(ASTNode *N, bool &Error);
+    void Normalize(ASTNode *N);
+
+    // Sort the variables in a node of the AST. The Error field is set if an
+    // error is encountered during normalization.
+    // @param[in] N is the root of the AST.
+    void Sort(ASTNode *N);
 
     // Check if the two ASTs N1 and N2 are equal.
     // @param[in] N1 is the first AST.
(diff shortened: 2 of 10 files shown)


mgrang:
I will implement the coalescing of nodes of the AST and performing arithmetic optimizations on the nodes in follow-up PRs.

clang/include/clang/AST/PreorderAST.h
```diff
@@ -0,0 +1,174 @@
...
+  using VarListTy = std::vector<VarTy>;
+  using OpcodeTy = BinaryOperator::Opcode;
+  using ConstTy = llvm::APSInt;
+  using Result = Lexicographic::Result;
+
+  class PreorderAST {
+    class ASTNode {
+    public:
+      ASTContext &Ctx;
+      OpcodeTy opcode;
```
kkjeer, line 52: Is there a precedent for using camelCase names elsewhere in clang? Other files such as SemaBounds, BoundsAnalysis, CanonBounds, etc. seem to use PascalCase.

clang/include/clang/AST/PreorderAST.h
```diff
@@ -0,0 +1,174 @@
...
+        left(nullptr), right(nullptr), parent(Parent),
+        HasConstant(false) {
+          // We initialize the constant value for each node to 0. So we need a
+          // way to indicate the absence of a constant. So we have a boolean
+          // HasConstant to indicate whether a node has a constant or not.
+          constant = getConstVal(0);
+        }
+
+      llvm::APSInt getConstVal(unsigned Val) {
+      return llvm::APSInt(llvm::APInt(Ctx.getTypeSize(Ctx.IntTy), Val));
```
kkjeer, line 69: Nit: indentation

clang/include/clang/AST/PreorderAST.h
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    // Print the preorder AST.
+    // @param[in] N is the root node of the AST.
+    void print(ASTNode *N);
+
+    // Invoke IgnoreValuePreservingOperations to strip off casts.
+    // @param[in] E is the expression whose casts must be stripped.
+    // @return E with casts stripped off.
+    Expr *IgnoreCasts(const Expr *E);
```
kkjeer, line 156: The name IgnoreCasts is somewhat confusing since Lex.IgnoreValuePreservingOperations also ignores non-cast operations such as `&*e`.

clang/include/clang/AST/PreorderAST.h
```diff
@@ -0,0 +1,174 @@
...
+    // @param[in] An expression E.
+    // @return Whether E is an expression containing a reference to an array
+    // subscript or a pointer dereference.
+    bool IsDeclOperand(Expr *E);
+
+    // Get the DeclRefExpr from an expression E.
+    // @param[in] An expression E which is known to be either an LValueToRValue
+    // cast or an ArrayToPointerDecay cast.
+    // @return The DeclRefExpr from the expression E.
+    DeclRefExpr *GetDeclOperand(Expr *E);
```
kkjeer, line 170: Can IsDeclOperand and GetDeclOperand be combined into a single method that returns nullptr if E does not fit the criteria for IsDeclOperand?

clang/lib/AST/PreorderAST.cpp
```diff
@@ -0,0 +1,202 @@
...
+    if (isa<BinaryOperator>(RHS))
+      // Create the RHS as the right child of the current node.
+      Create(N->Right, RHS, /* parent node */ N);
+    else
+      // Create the RHS in the current node.
+      Create(N, RHS);
+  }
+}
+
+void PreorderAST::Sort(ASTNode *N, bool &Error) {
```
kkjeer, line 99: I find it somewhat confusing that the out parameter Error is named the same as the PreorderAST member Error - is the member PreorderAST::Error ever set during Sort or Normalize, or only the out parameter Error?

clang/lib/AST/PreorderAST.cpp
```diff
@@ -0,0 +1,202 @@
...
+    else
+      Parent->Right = N;
+  }
+
+  E = Lex.IgnoreValuePreservingOperations(Ctx, E);
+
+  // If E is a variable, store its name in the variable list for the current
+  // node. Initialize the count of the variable to 1.
+  DeclRefExpr *D;
+  if (IsDeclOperand(E, D)) {
```
kkjeer, line 59: Could `IsDeclOperand` return a `DeclRefExpr *` rather than a boolean with a `DeclRefExpr *` out parameter? Then the caller could do `if (DeclRefExpr *D = IsDeclOperand(E))` (or potentially rename `IsDeclOperand` to `GetDeclOperand`).

clang/test/CheckedC/inferred-bounds/widened-bounds-semantic-compare.c
```diff
@@ -0,0 +1,36 @@
+// Tests for bounds widening of _Nt_array_ptr's using function to semantically
```
kkjeer, line 1: Should there also be some tests that involve constants?

mgrang: There are 3 reasons why I did not add tests involving constants yet:
1. Currently I only call CompareExprSemantic from BoundsAnalysis where it is invoked to compare only the bases of expressions. So we do not have any callers which invoke CompareExprSemantic for offsets (constants).
2. I intend to add tests for constants in my PR which does constant folding. That way I would be able to test constants with +/* operators.
3. A more full-fledged testing of constants can happen once I integrate the semantic comparison with SplitIntoBaseOffset logic.

clang/include/clang/AST/CanonBounds.h
```diff
@@ -50,7 +50,8 @@ namespace clang {
     enum class Result {
       LessThan,
       Equal,
-      GreaterThan
+      GreaterThan,
```
dtarditi, line 53: I think this overloads Result in a confusing way.  Result is meant to represent the result of a lexicographic comparison, which produces total ordering.  It is confusing to now have NotEqual in Result.

clang/include/clang/AST/CanonBounds.h
```diff
@@ -118,6 +119,9 @@ namespace clang {
     /// \brief Lexicographic comparison of expressions that can occur in
     /// bounds expressions.
     Result CompareExpr(const Expr *E1, const Expr *E2);
+    /// \brief Semantic comparison of expressions that can occur in
+    /// bounds expressions.
+    Result CompareExprSemantically(const Expr *E1, const Expr *E2);
```
dtarditi, line 124: Could this just return a boolean?

clang/include/clang/AST/PreorderAST.h
```diff
@@ -0,0 +1,165 @@
...
+#include "clang/AST/ASTContext.h"
+#include "clang/AST/CanonBounds.h"
+#include "clang/AST/Expr.h"
+
+namespace clang {
+  // Each node of the PreorderAST has a variable-count list which stores the
+  // Name of each variable along with a count of the number of times the
+  // variable appears in the node.
+  struct VarTy {
+    std::string Name;
```
dtarditi, line 28: It would be better to use a VarDecl instead.   Because of the way bounds expressions work, you could have occurrences of two variables in the same scope with the same name.  Consider
```
array_ptr<int> p : count(len) = ...
int len = ...
... *p ...  // has a bounds expression that uses the first occurrence of len.
```

clang/include/clang/AST/PreorderAST.h
```diff
@@ -0,0 +1,165 @@
...
+
+  // Each node of the PreorderAST contains 3 fields:
+  // - Opcode: The Opcode for the node.
+  // - variable-count list: A list of variables and their counts.
+  // - Constant value: All Constants occurring in the node are Constant folded.
+
+  // For example, consider an expression E = a + 3 + b + 4 + a
+  // A node of the PreorderAST for E would contain the following:
+  // - Opcode: +
+  // - variable-count list: [a:2, b:1]
```
dtarditi, line 40: I think this AST is overly specialized to the particular transformation that you are doing.   That's fine for this change, but I think you'll want to generalize it for symbolic evaluation.
- Some operators are commutative, but not associative.  An example is signed integer arithmetic, which has undefined behavior on overflow.  You can't represent them properly in this AST.
- Some arithmetic operators are not commutative or associative, such as division.
- It would be helpful to have leaf expressions that are non-variable/non-constant expressions.

clang/include/clang/AST/PreorderAST.h
```diff
@@ -0,0 +1,165 @@
...
+    DeclRefExpr *GetDeclOperand(Expr *E);
+
+  public:
+    PreorderAST(ASTContext &Ctx, Expr *E) :
+      Ctx(Ctx), Lex(Lexicographic(Ctx, nullptr)),
+      OS(llvm::outs()), Error(false) {
+
+      AST = new ASTNode(Ctx);
+      Create(AST, E);
+      Normalize(AST);
```
dtarditi, line 142: It would be better to not have the constructor normalize expressions.  That makes code harder to read and understand because there's an implicit operation at construction of a PreorderAST.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -0,0 +1,202 @@
...
+      // Create the LHS in the current node.
+      Create(N, LHS);
+
+    if (isa<BinaryOperator>(RHS))
+      // Create the RHS as the right child of the current node.
+      Create(N->Right, RHS, /* parent node */ N);
+    else
+      // Create the RHS in the current node.
+      Create(N, RHS);
+  }
```
dtarditi, line 92: If an expression isn't a constant or variable, this code just discard the expression.   Shouldn't you call SetError in this case?  This would result in `e1 * a * 5` being considered to `e2 * a * 5`, for example.  The alternative is to add a node to your AST to represent expressions that aren't variables or constants.

clang/include/clang/AST/CanonBounds.h
```diff
@@ -118,6 +118,9 @@ namespace clang {
     /// \brief Lexicographic comparison of expressions that can occur in
     /// bounds expressions.
     Result CompareExpr(const Expr *E1, const Expr *E2);
+    /// \brief Semantic comparison of expressions that can occur in
+    /// bounds expressions.
+    bool CompareExprSemantically(const Expr *E1, const Expr *E2);
```
dtarditi, line 123: You might want to describe the meaning of the return value: that true means two expressions are equivalent semantically.   Clang has this unusual convention where the boolean returns from many functions are inverted in their meaning.

clang/include/clang/AST/PreorderAST.h
```diff
@@ -0,0 +1,124 @@
...
+    // @param[in] this is the first AST.
+    // @param[in] P is the second AST.
+    // @return Returns a bool indicating whether the two ASTs are equal.
+    bool IsEqual(PreorderAST &P) { return IsEqual(Root, P.Root); }
+
+    // Check if an error has occurred during transformation of the AST.
+    // @return Whether an error has occurred or not.
+    bool GetError() { return Error; }
+
+    // Cleanup the memory consumed by the AST. This is intended to be called
```
dtarditi, line 117: This is a public method: do you need to mention that it is intended to be called from outside this class?

clang/include/clang/AST/PreorderAST.h
dtarditi: This looks it is fixed.

clang/lib/AST/PreorderAST.cpp
dtarditi: Do you need to call SetError() for other unhandled cases?

clang/lib/AST/PreorderAST.cpp
```diff
@@ -0,0 +1,200 @@
...
+  if (llvm::APSInt::compareValues(N1->Const, N2->Const) != 0)
+    return false;
+
+  // Match each variable occurring in the two nodes.
+  for (size_t I = 0; I != N1->Vars.size(); ++I) {
+    auto &V1 = N1->Vars[I];
+    auto &V2 = N2->Vars[I];
+
+    // If any variable differs between the two nodes.
+    if (V1->getQualifiedNameAsString().compare(
```
dtarditi, line 140: You should use CanonBounds::CompareDecl instead.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+  if (!N->IsOpCommutativeAndAssociative()) {
+    SetError();
+    return;
+  }
+
+  // Sort the variables in the node lexicographically.
+  llvm::sort(N->Vars.begin(), N->Vars.end(),
+             [](const VarDecl *V1, const VarDecl *V2) {
+               return V1->getQualifiedNameAsString().compare(
```
dtarditi, line 105: Two VarDecls may have the same name, but represent different variables from different scopes.  In this case, your sort may be unstable and put them into a different orders for different trees, which would cause the comparison to fail.

clang/lib/AST/PreorderAST.cpp
mgrang: I am calling `SetError()` on line 87. This acts as the fall-through case for all unhandled cases in this function.

mgrang:
@dtarditi I have addressed your latest comments.

mgrang:
There are some failures observed on ADO which don't seem to have any relation to this PR. I was not able to reproduce the errors on the ADO build machine manually. So I will merge this PR now and if ADO still breaks then I will look into fixing the failures.

PR #860, opened 2020-06-19
Fixing intermittent test failures

Machiry:
Reference: https://github.com/microsoft/checkedc-clang/pull/854

We fixed root cause memory corruption and enabled all the tests.


mgrang:
I have verified the Windows X86 ADO build: https://dev.azure.com/msresearch/CheckedC/_build/results?buildId=513155&view=results

PR #871, opened 2020-07-06
Compare Expressions: More normalizations of the preorder AST

mgrang:
We perform more normalizations of the preorder AST:
    1. Coalesce nodes with the same commutative and associate operator. This
    involves moving variables to the parent node.
    2. While coalescing, constant-fold the constant with the parent constant.
    3. Sort the children nodes of a node lexicographically.


mgrang:
This is a follow-up to https://github.com/microsoft/checkedc-clang/pull/858

mgrang:
If this PR merges first I would have to re-implement the sorting of nodes when I add support for having more than 2 child nodes to the AST. So I would first like to implement that support and will make this a draft PR until then.

mgrang:
Closing in favor of https://github.com/microsoft/checkedc-clang/pull/879

PR #875, opened 2020-07-09
Compare Expressions: Make the preorder AST an n-ary tree [2/n]

mgrang:
Previously, the preorder AST was a binary tree with left and right children.
But after coalescing, a node may have more than two children. Here we replace
left and right pointers with a vector representing the list of child nodes of a
given node.

### DIFF
--- a/clang/include/clang/AST/PreorderAST.h
+++ b/clang/include/clang/AST/PreorderAST.h
@@ -42,7 +42,7 @@ namespace clang {
     llvm::APSInt Const;
     bool HasConst;
     Node *Parent;
-    llvm::SetVector<Node *> Children;
+    llvm::SmallVector<Node *, 2> Children;
 
     Node(Node *Parent) :
       Opc(BO_Add), HasConst(false), Parent(Parent) {}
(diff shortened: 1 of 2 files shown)


clang/include/clang/AST/PreorderAST.h
```diff
@@ -25,24 +25,27 @@ namespace clang {
...
     BinaryOperator::Opcode Opc;
     std::vector<const VarDecl *> Vars;
     llvm::APSInt Const;
-    // HasConst indicates whether there is a constant in the node. This is used
-    // to differentiate between an absence of a constant and a constant of value
-    // 0.
     bool HasConst;
-    Node *Parent, *Left, *Right;
+    Node *Parent;
+    llvm::SetVector<Node *> Children;
```
dtarditi, line 45: How about using a SmallVector instead?   Why use a SetVector?

mgrang: I used a SetVector to reduce make it easier to remove an element from the list. But SetVector has a larger space complexity than a vector. I will change this to a SmallVector.

PR #876, opened 2020-07-10
Compare Expressions: Add a field for non-variable, non-constant expressions [3/n]

mgrang:
Currently, a node of the preorder AST has fields for variables and constants. However, an input expression may contain sub-expressions which are neither variables nor constants. We add a field "Others" of type Expr to store all such expressions.

### DIFF
--- a/clang/include/clang/AST/PreorderAST.h
+++ b/clang/include/clang/AST/PreorderAST.h
@@ -30,6 +30,7 @@ namespace clang {
   // Opc: The opcode of the operator.
   // Vars: A list of variables in the sub expression.
   // Const: Constants of the sub expression are folded.
+  // Others: A list of all other non-variable, non-constant expressions.
   // HasConst: Indicates whether there is a constant in the node. It is used to
   // differentiate between the absence of a constant and a constant value of 0.
   // Parent: A link to the parent node of the current node.
@@ -40,6 +41,7 @@ namespace clang {
     BinaryOperator::Opcode Opc;
     std::vector<const VarDecl *> Vars;
     llvm::APSInt Const;
+    std::vector<const Expr *> Others;
     bool HasConst;
     Node *Parent;
     std::vector<Node *> Children;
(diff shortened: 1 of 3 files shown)


clang/include/clang/AST/PreorderAST.h
```diff
@@ -25,24 +25,29 @@ namespace clang {
   using Result = Lexicographic::Result;
 
   // Each binary operator of an expression results in a new node of the
-  // PreorderAST. Each node contains 3 fields:
+  // PreorderAST. Each node contains the following fields:
+
   // Opc: The opcode of the operator.
   // Vars: A list of variables in the sub expression.
   // Const: Constants of the sub expression are folded.
+  // Others: A list of all other non-variable, non-constant expressions.
```
dtarditi, line 33: I think it would be better in the long run to generalize the node type into a class hierarchy.  Something along the lines of:
```
Node
 |  \ 
 |    \
 |      \
 |        \
 |          \
BinaryNode  LeafExprNode
```
where `LeafExpr` could just wrap a clang AST expression.  A representation that separates binary expressions and regular clang expression loses the order of evaluation information, which we'll want later on.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -130,6 +130,14 @@ bool PreorderAST::IsEqual(Node *N1, Node *N2) {
   if (llvm::APSInt::compareValues(N1->Const, N2->Const) != 0)
     return false;
 
+  // If the number of other expressions in the two nodes mismatch.
+  if (N1->Others.size() != N2->Others.size())
+    return false;
```
dtarditi, line 135: It would avoid duplicated logic like this.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -140,9 +148,26 @@ bool PreorderAST::IsEqual(Node *N1, Node *N2) {
       return false;
   }
 
-  // Recursively match the left and the right subtrees of the AST.
-  return IsEqual(N1->Left, N2->Left) &&
-         IsEqual(N1->Right, N2->Right);
+  // Match each of the other expressions occurring in the two nodes.
```
dtarditi, line 151: This is also duplicated logic.

mgrang:
Superseded by https://github.com/microsoft/checkedc-clang/pull/901

PR #879, opened 2020-07-14
Compare Expressions: Coalesce nodes, constant-fold, sort child nodes [4/n]

mgrang:
1. We coalesce nodes having the same operator. Coalescing involves moving the
variables, constants and other expressions to the parent and removing the
current node.

2. As part of coalescing, we also constant-fold the constant of the current
node with that of the parent.

3. We also sort the child nodes based on several criteria.


mgrang:
Superseded by https://github.com/microsoft/checkedc-clang/pull/901 and https://github.com/microsoft/checkedc-clang/pull/906

PR #885, opened 2020-07-22
Compare Expressions: Add Sema parameter to PreorderAST class [5/n]

mgrang:
The logic for splitting an expression into a base and an offset would use the
preorder AST. There we would need to construct BinaryOperator expressions and
pass FPOptions. We would use the Sema to access FPOptions.


mgrang:
This draft PR is no longer needed.

PR #889, opened 2020-07-24
Compare Expressions: Use a SetVector for variables and other expressions [6/n]

mgrang:
For splitting an expression into a base and offset we would need to compare the
given expression and match it with its declared upper bound expression. In
doing so, we would match variables and other sub-expressions. Using a SetVector
would make the matching more efficient as compared to a vector.


mgrang:
Superseded by #901 and #906

Issue #870, opened 2020-07-06
Introduce temporary bindings for struct-typed compound literals


kkjeer:
PR #864 binds array-typed and function-typed compound literals to temporaries. This would allow bounds to be inferred for lvalue expressions such as `&(int [1]){ 0 }`. (see issue #859). The lvalue bounds of a struct-typed compound literal expression `e` should be `(&temp(e), &temp(e) + 1)`.

There are two issues that currently result from attempting to bind struct-typed compound literals to temporaries which should be fixed:

#### Assert failure in code generation
The assert `!E->getSubExpr()->isLValue())` fails in ScalarExprEmitter::VisitCHKCBindTemporaryExpr (CGExprScalar.cpp).

String literals and array-typed compound literals (which are both lvalues) are bound to temporaries. However, ScalarExprEmitter::VisitCHKBindTemporaryExpr is not called on these expressions since they are wrapped in ArrayToPointerDecay cast expressions rather than LValueToRValue cast expressions. For ArrayToPointerDecay casts, ScalarExprEmitter::VisitCastExpr calls CGF.EmitArrayToPointerDecay on the cast's subexpression. For LValueToRValue casts, ScalarExprEmitter::VisitCastExpr calls ScalarExprEmitter::Visit on the cast's subexpression.

#### Compile error for assignments involving a struct-typed literal
Assigning a struct-typed compound literal to a variable results in a compile error: "cannot compile this aggregate expression yet". For example:

```
struct S s;
s = (struct S){ 0 }; // error: cannot compile this aggregate expression yet

struct S s1 = (struct S){ 0 }; // no error
```

This error occurs because the AggExprEmitter class currently has no VisitCHKCBindTemporaryExpr method.

PR #890, opened 2020-07-24
Fix expected error in compound-literals test

kkjeer:
Fix the expected error in the compound-literals test to reflect the changes introduced by validating the bounds context.

#### Testing:
* Update compound-literals.c to expect the correct error message
* Automated testing in progress


kkjeer:
> LGTM. Although it would be good to explain why the bounds of `arr` are unknown. I guess Yahui's work would be help us get more verbose errors.

The bounds for `arr` are currently unknown because we do not infer bounds for struct-typed compound literals. Issue #870 tracks this work - I added a TODO to the test.

PR #893, opened 2020-08-03
Target a single assignment in bounds checking error messages

dopelsunce:
- In bounds checking related diagnostic messages, highlight the assignment within a top-level statement that causes the error or warning. This PR considers two types of bounds checking error messages for a variable V:
  1. The compiler cannot prove or can disprove the declared bounds for V are valid after an assignment to a variable in the bounds of V; and
  2. The inferred bounds of V become unknown after an assignment to a variable used in the declared bounds for V.
- Print 'assignment', 'increment', 'decrement' or 'initialization' in the error message depending on the "type" of the target expression. 

#### An example

```
array_ptr<int> p : count(len) = ...;
other = 0, len = len * 2;
```

Error message before the change (highlight the whole statement):
```
error: inferred bounds for 'p' are unknown after statement
  other = 0, len = len * 2;
  ^~~~~~~~~~~~~~~~~~~~~~~~~
```

After the change (blame the embedded assignment that caused the error; print 'assignment' instead of 'statement'):
```
error: inferred bounds for 'p' are unknown after assignment
  other = 0, len = len * 2;
             ^~~~~~~~~~~~~
```

### DIFF
--- a/clang/include/clang/Sema/Sema.h
+++ b/clang/include/clang/Sema/Sema.h
@@ -5187,7 +5187,9 @@ public:
   enum BoundsDeclarationCheck {
       BDC_Assignment,
       BDC_Initialization,
-      BDC_Statement
+      BDC_Statement,
+      BDC_Increment,
+      BDC_Decrement
   };
 
   /// \brief Check that address=of operation is not taking the
--- a/clang/include/clang/Sema/Sema.h
+++ b/clang/include/clang/Sema/Sema.h
@@ -5186,10 +5186,10 @@ public:
 
   enum BoundsDeclarationCheck {
       BDC_Assignment,
+      BDC_Decrement,
+      BDC_Increment,
       BDC_Initialization,
       BDC_Statement,
-      BDC_Increment,
-      BDC_Decrement
   };
 
   /// \brief Check that address=of operation is not taking the
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -4056,8 +4056,9 @@ namespace {
         SrcRange = BlameExpr->getSourceRange();
 
         // Choose the type of assignment E to show in the diagnostic messages
-        // from: assignment (=), decrement (--) or increment (++). These are all
-        // modifying operators that can update observed bounds.
+        // from: assignment (=), decrement (--) or increment (++). If none of
+        // these cases match, the diagnostic message reports that the error is
+        // for a statement.
         if (UnaryOperator *UO = dyn_cast<UnaryOperator>(BlameExpr)) {
           if (UO->isIncrementOp())
             BDCType = Sema::BoundsDeclarationCheck::BDC_Increment;
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -3941,21 +3941,8 @@ namespace {
     void DiagnoseUnknownObservedBounds(Stmt *St, const VarDecl *V,
                                        BoundsExpr *DeclaredBounds,
                                        CheckingState State) {
-      // Blame the assignment in St that causes V's observed bounds to be
-      // unknown if we can find it in map BlameAssignments.  Otherwise, blame
-      // the whole St in the error message.
-      SourceLocation Loc = St->getBeginLoc();
-      SourceRange SrcRange = St->getSourceRange();
-      auto BDCType = Sema::BoundsDeclarationCheck::BDC_Statement;
-      auto It = State.BlameAssignments.find(V);
-      if (It != State.BlameAssignments.end()) {
-        Expr *BlameExpr = It->second;
-        Loc = BlameExpr->getBeginLoc();
-        SrcRange = BlameExpr->getSourceRange();
-        BDCType = GetBDCTypeFromExpr(BlameExpr);
-      }
-      S.Diag(Loc, diag::err_unknown_inferred_bounds)
-        << BDCType << V << SrcRange;
+      BlameAssignmentWithinStmt(St, V, State,
+                                diag::err_unknown_inferred_bounds);
       S.Diag(V->getLocation(), diag::note_declared_bounds)
         << DeclaredBounds << DeclaredBounds->getSourceRange();
 
@@ -4022,13 +4009,35 @@ namespace {
           return;
       }
 
+      unsigned DiagId = (Result == ProofResult::False) ?
+        diag::error_bounds_declaration_invalid :
+        (CSS != CheckedScopeSpecifier::CSS_Unchecked?
+          diag::warn_checked_scope_bounds_declaration_invalid :
+          diag::warn_bounds_declaration_invalid);
+      SourceLocation Loc = BlameAssignmentWithinStmt(St, V, State, DiagId);
+      if (Result == ProofResult::False)
+        ExplainProofFailure(Loc, Cause, ProofStmtKind::BoundsDeclaration);
+      S.Diag(V->getLocation(), diag::note_declared_bounds)
+        << DeclaredBounds << DeclaredBounds->getSourceRange();
+      S.Diag(Loc, diag::note_expanded_inferred_bounds)
+        << ObservedBounds << ObservedBounds->getSourceRange();
+    }
+
+    // BlameAssignmentWithinStmt prints a diagnostic message that highlights the
+    // assignment expression in St that causes V's observed bounds to be unknown
+    // or invalid.  If St is a DeclStmt, St itself and V are highlighted.
+    // BlameAssignmentWithinStmt returns the source location of the blamed
+    // assignment.
+    SourceLocation BlameAssignmentWithinStmt(Stmt *St, 
+                 const VarDecl *V, CheckingState State, unsigned DiagId) const {
+      SourceRange SrcRange = St->getSourceRange();
+      auto BDCType = Sema::BoundsDeclarationCheck::BDC_Statement;
+      
       // For a declaration, the diagnostic message should start at the
       // location of v rather than the beginning of St.  If the message
       // starts at the beginning of a declaration T v = e, then extra
       // diagnostics may be emitted for T.
       SourceLocation Loc = St->getBeginLoc();
-      SourceRange SrcRange = St->getSourceRange();
-      auto BDCType = Sema::BoundsDeclarationCheck::BDC_Statement;
       if (isa<DeclStmt>(St)) {
         Loc = V->getLocation();
         BDCType = Sema::BoundsDeclarationCheck::BDC_Initialization;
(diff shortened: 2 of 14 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4017,11 +4058,35 @@ namespace {
         << ObservedBounds << ObservedBounds->getSourceRange();
     }
 
+    // Returns the type of assignment E to show in the diagnostic messages:
+    // assignment (=), decrement (--) or increment (++).
+    Sema::BoundsDeclarationCheck GetBDCTypeFromExpr(Expr *E) const {
+      if (UnaryOperator *UO = dyn_cast<UnaryOperator>(E)) {
+        switch (UO->getOpcode()) {
```
kkjeer, line 4065: You can use `UnaryOperator::isIncrementOp(UO->getOpcode())` and `UnaryOperator::isDecrementOp(E->getOpcode())` to determine whether `E` is an increment operator or a decrement operator, respectively.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -239,7 +239,7 @@ namespace {
 
 bool Sema::AbstractForFunctionType(
   BoundsAnnotations &Annots,
-  ArrayRef<DeclaratorChunk::ParamInfo> Params) {  
+  ArrayRef<DeclaratorChunk::ParamInfo> Params) {
```
kkjeer, line 242: Please remove whitespace changes from this PR.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3998,17 +4027,29 @@ namespace {
...
       SourceLocation Loc = St->getBeginLoc();
-      if (isa<DeclStmt>(St))
+      SourceRange SrcRange = St->getSourceRange();
+      auto BDCType = Sema::BoundsDeclarationCheck::BDC_Statement;
+      if (isa<DeclStmt>(St)) {
         Loc = V->getLocation();
+        BDCType = Sema::BoundsDeclarationCheck::BDC_Initialization;
+      }
+
+      // Find the last assignment to V to blame in the diagnostic message if it
```
kkjeer, line 4037: The expression in BlameAssignments (if it exists) might not be an assignment to V. It could be an assignment to a variable used in the bounds of V.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4069,8 +4134,10 @@ namespace {
         const VarDecl *W = Pair.first;
         BoundsExpr *Bounds = Pair.second;
         BoundsExpr *AdjustedBounds = ReplaceVariableInBounds(Bounds, V, OriginalValue, CSS);
-        if (!Pair.second->isUnknown() && AdjustedBounds->isUnknown())
+        if (!Pair.second->isUnknown() && AdjustedBounds->isUnknown()) {
```
kkjeer, line 4137: This seems to only update BlameAssignments if the updated observed bounds of W are unknown. For a case like:

```
unsigned int i;
array_ptr<int> p : count(i) = 0;
// Original value of i in i++ is i - 1
// i++ causes the updated observed bounds of p to be (p, p + i - 1)
i++, unrelated_variable = 0;
```

We want `BlameAssignments[p] => i++`.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -705,6 +705,21 @@ namespace {
       // user when diagnosing unknown bounds errors.
       llvm::DenseMap<const VarDecl *, SmallVector<Expr *, 4>> UnknownSrcBounds;
 
+      // BlameAssignments maps a variable declaration V to an expression in a
+      // top-level CFG statement that last updates any variable used in the
+      // declared bounds of V.
+      //
+      // BlameAssignments is used to provide more context for two types of
+      // diagnostic messages:
+      //   1. The compiler cannot prove or can disprove the declared bounds for
```
kkjeer, line 714: It is also possible that the compiler cannot prove or can disprove the declared bounds for V are valid after an assignment to a variable used in the declared bounds of V. For example:

```
unsigned int len;
array_ptr<int> p : count(len) = 0;
// Original value of len is len + 2
// Updated observed bounds of p are (p, p + len - 2)
// Cannot prove that bounds(p, p + len - 2) imply (p, p + len)
len = len + 2;
```

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4017,11 +4058,35 @@ namespace {
...
+          break;
+        case UnaryOperatorKind::UO_PreInc:
+        case UnaryOperatorKind::UO_PostInc:
+          return Sema::BoundsDeclarationCheck::BDC_Increment;
+          break;
+        default:
+          break;
+        }
+      } else if (dyn_cast<BinaryOperator>(E)) {
+        // Must be an assignment or a compounds assignment, because E is
```
kkjeer, line 4078: If you're not using the result of `dyn_cast<BinaryOperator>(E)`, use `isa<BinaryOperator>(E)` instead.

kkjeer, line 4078: Typo: compounds => compound.

clang/lib/Sema/SemaBounds.cpp
dopelsunce: Thanks for the suggestion. I found a more succinct shorthand for this: `UO->isDecrementOp()`.

clang/lib/Sema/SemaBounds.cpp
dopelsunce: I've changed the comment to this general case.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3998,17 +4027,28 @@ namespace {
...
-      if (isa<DeclStmt>(St))
+      SourceRange SrcRange = St->getSourceRange();
+      auto BDCType = Sema::BoundsDeclarationCheck::BDC_Statement;
+      if (isa<DeclStmt>(St)) {
         Loc = V->getLocation();
+        BDCType = Sema::BoundsDeclarationCheck::BDC_Initialization;
+      }
+
+      // Find the assignment (if it exists) to blame for the error or warning.
+      auto It = State.BlameAssignments.find(V);
```
kkjeer, line 4038: Can this logic move into a helper method that sets a SourceLocation, SourceRange, and BDCType based on `St` and `BlameAssignments`?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3920,13 +3936,26 @@ namespace {
...
                                        BoundsExpr *DeclaredBounds,
                                        CheckingState State) {
-      S.Diag(St->getBeginLoc(), diag::err_unknown_inferred_bounds)
-        << V << St->getSourceRange();
+      // Blame the assignment in St that causes V's observed bounds to be
+      // unknown if we can find it in map BlameAssignments.  Otherwise, blame
+      // the whole St in the error message.
+      SourceLocation Loc = St->getBeginLoc();
+      SourceRange SrcRange = St->getSourceRange();
+      auto BDCType = Sema::BoundsDeclarationCheck::BDC_Statement;
```
kkjeer, line 3949: This could miss the case where the bounds of `V` are unknown after an initialization, e.g.

```
array_ptr<int> unknown_src : bounds(unknown);
// Bounds of arr are unknown after initialization
array_ptr<int> arr : count(1) = unknown_src;
```

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4017,11 +4023,55 @@ namespace {
         << ObservedBounds << ObservedBounds->getSourceRange();
     }
 
+    // BlameAssignmentWithinStmt prints a diagnostic message that highlights the
+    // assignment expression in St that causes V's observed bounds to be unknown
+    // or invalid.  If St is a DeclStmt, St itself and V are highlighted.
+    // BlameAssignmentWithinStmt returns the source location of the blamed
+    // assignment.
+    SourceLocation BlameAssignmentWithinStmt(Stmt *St, 
+                 const VarDecl *V, CheckingState State, unsigned DiagId) const {
```
kkjeer, line 4032: Please align `const VarDecl *V` underneath `Stmt *St` (and ensure the remaining parameters are on separate lines if necessary in order to prevent lines longer than 80 characters).

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4017,11 +4023,55 @@ namespace {
         << ObservedBounds << ObservedBounds->getSourceRange();
     }
 
+    // BlameAssignmentWithinStmt prints a diagnostic message that highlights the
+    // assignment expression in St that causes V's observed bounds to be unknown
+    // or invalid.  If St is a DeclStmt, St itself and V are highlighted.
```
kkjeer, line 4028: Nit: BlameAssignmentWithinStmt can also be used to print a diagnostic message if V's bounds are not necessarily invalid, but are not provably valid. I suggest updating the comment to something like "not provably valid" instead of "invalid".

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4017,11 +4023,55 @@ namespace {
...
+                 const VarDecl *V, CheckingState State, unsigned DiagId) const {
+      SourceRange SrcRange = St->getSourceRange();
+      auto BDCType = Sema::BoundsDeclarationCheck::BDC_Statement;
+      
+      // For a declaration, the diagnostic message should start at the
+      // location of v rather than the beginning of St.  If the message
+      // starts at the beginning of a declaration T v = e, then extra
+      // diagnostics may be emitted for T.
+      SourceLocation Loc = St->getBeginLoc();
+      if (isa<DeclStmt>(St)) {
```
kkjeer, line 4041: Is is possible for a `DeclStmt` to have an entry in `State.BlameAssignments`? If not, you can early return here rather than searching for `V` in `BlameAssignments` below.

clang/lib/Sema/SemaBounds.cpp
dopelsunce: Thanks for the suggestion!

clang/lib/Sema/SemaBounds.cpp
dopelsunce: `DeclStmt` does not have entries in `State.BlameAssignments`. I made the function return early inside the if block.

clang/include/clang/Sema/Sema.h
```diff
@@ -5187,7 +5187,9 @@ class Sema {
   enum BoundsDeclarationCheck {
       BDC_Assignment,
       BDC_Initialization,
-      BDC_Statement
+      BDC_Statement,
```
mgrang, line 5190: Should the enums be listed alphabetically?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4017,11 +4023,59 @@ namespace {
...
     }
 
+    // BlameAssignmentWithinStmt prints a diagnostic message that highlights the
+    // assignment expression in St that causes V's observed bounds to be unknown
+    // or not provably valid.  If St is a DeclStmt, St itself and V are
+    // highlighted.  BlameAssignmentWithinStmt returns the source location of
+    // the blamed assignment.
+    SourceLocation BlameAssignmentWithinStmt(Stmt *St, const VarDecl *V,
+                                             CheckingState State,
+                                             unsigned DiagId) const {
```
mgrang, line 4033: We should assert that `St` is non-null inside this function.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4017,11 +4023,59 @@ namespace {
...
+        Expr *BlameExpr = It->second;
+        Loc = BlameExpr->getBeginLoc();
+        SrcRange = BlameExpr->getSourceRange();
+        
+        // Choose the type of assignment E to show in the diagnostic messages:
+        // assignment (=), decrement (--) or increment (++).
+        if (UnaryOperator *UO = dyn_cast<UnaryOperator>(BlameExpr)) {
+          if (UO->isIncrementOp())
+            BDCType = Sema::BoundsDeclarationCheck::BDC_Increment;
+          else if (UO->isDecrementOp())
```
mgrang, line 4062: Are there plans to handle other unary operators like `unary minus`, `deref`, `address-of`, etc?

clang/include/clang/Sema/Sema.h
dopelsunce: I can make this adjustment. Just curious, is it a C++ coding convention to list items alphabetically in an enum?

clang/lib/Sema/SemaBounds.cpp
dopelsunce: No. Here `UO` must be a self-modifying unary operator, and I can only think of `++` and `--`.

I can also put an assert here.

clang/include/clang/Sema/Sema.h
mgrang: No, this is not a C++ coding convention. Listing enums/includes, etc alphabetically can help with code readability.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4017,11 +4023,61 @@ namespace {
...
+
+        // Choose the type of assignment E to show in the diagnostic messages
+        // from: assignment (=), decrement (--) or increment (++). These are all
+        // modifying operators that can update observed bounds.
+        if (UnaryOperator *UO = dyn_cast<UnaryOperator>(BlameExpr)) {
+          if (UO->isIncrementOp())
+            BDCType = Sema::BoundsDeclarationCheck::BDC_Increment;
+          else if (UO->isDecrementOp())
+            BDCType = Sema::BoundsDeclarationCheck::BDC_Decrement;
+        } else if (isa<BinaryOperator>(BlameExpr)) {
```
dtarditi, line 4066: If both the increment and decrement tests fail, the code would silently fall off the end of this block and not issue an error.  You should add an else clause that contains an llvm_unreachable statement, since this is not expected to happen.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4017,11 +4023,61 @@ namespace {
...
+      // for the error or warning.
+      auto It = State.BlameAssignments.find(V);
+      if (It != State.BlameAssignments.end()) {
+        Expr *BlameExpr = It->second;
+        Loc = BlameExpr->getBeginLoc();
+        SrcRange = BlameExpr->getSourceRange();
+
+        // Choose the type of assignment E to show in the diagnostic messages
+        // from: assignment (=), decrement (--) or increment (++). These are all
+        // modifying operators that can update observed bounds.
```
dtarditi, line 4060: Could you a comment that says "If none of these cases match, the diagnostic message reports that the error is for a statement."?   When I first read the code, I noticed there were several ways we could fall off the end of an if-statement chain.  It doesn't matter because the code just reports a diagnostic message for a statement type.

PR #897, opened 2020-08-11
Conditional operators: update checking state [1/n]

kkjeer:
This PR updates the checking state for conditional operators of the form `e1 ? e2 : e3`.

#### Notable changes:
The `CheckConditionalOperator` method is modified to update the checking state members:

##### ObservedBounds
The observed bounds context is updated as follows:
* Update the observed bounds context `State.ObservedBounds` by checking the test expression `e1`.
* Get the observed bounds context `StateTrueArm.ObservedBounds` by checking the true arm expression `e2`. The observed bounds context `State.ObservedBounds` from checking the test expression `e1` is used as the initial context to check `e2`.
* Get the observed bounds context `StateFalseArm.ObservedBounds` by checking the false arm expression `e3`. The observed bounds context `State.ObservedBounds` from checking the test expression `e1` is used as the initial context to check `e3`.
* If the contexts `StateTrueArm.ObservedBounds` and `StateFalseArm.ObservedBounds` are identical, the final observed bounds context is `StateTrueArm.ObservedBounds`.
* Otherwise:
  * For each context `UC` of `StateTrueArm.ObservedBounds` and `StateFalseArm.ObservedBounds`:
    * Get the set of variables whose bounds were modified in the arm context, i.e. the set of all variables `v` for which `UC[v] != State.ObservedBounds[v]`.
    * TODO: should this be the set of all variables `v` for which `UC[v] != declared bounds of v`? If so, then consider the example `(a = b) ? (c = d) : c`. The set of updated variables in the true arm `e2` would be `{ a, c  }` and the set of updated variables in the false arm `e3` would be `a`. If the bounds of `b` do not imply the declared bounds of `a`, an error will be emitted twice (once for each arm `e2` and `e3`).
    * For each variable `v` in the set, check that the observed bounds `UC[v]` that were updated in the arm `e2` or `e3` imply the declared bounds of `v`.
    * For each variable `v` in the set, set `State.ObservedBounds[v]` to the declared bounds of `v`.
  * The final observed bounds context `StateObservedBounds` will contain, for each variable `x`:
    * If the bounds of `x` were not modified in either arm `e2` or `e3`, the observed bounds of `x` will be the observed bounds of `x` from checking the test expression `e1`.
    * If the bounds of `x` were modified in either arm `e2` or `e3`, the observed bounds of `x` will be the declared bounds of `x`.

##### EquivExprs
The set of sets of equivalent expressions is updated as follows:
* Update the equivalent expressions set `State.EquivExprs` by checking the test expression `e1`.
* Update the equivalent expressions set `StateTrueArm.EquivExprs` by checking the true arm expression `e2`. The equivalent expressions set `State.EquivExprs` from checking the test expression `e1` is used as the initial set to check `e2`.
* Update the equivalent expressions set `StateFalseArm.EquivExprs` by checking the false arm expression `e3`. The equivalent expressions set `State.EquivExprs` from checking the test expression `e1` is used as the initial set to check `e3`.
* The final set of equivalent expressions `State.EquivExprs` is the intersection of `StateTrueArm.EquivExprs` and `StateFalseArm.EquivExprs`.

##### SameValue
The set of expressions that produce the same value as `e1 ? e2 : e3` is updated as follows:
* Update the set `StateTrueArm.SameValue` by checking the true arm `e2`.
* Update the set `StateFalseArm.SameValue` by checking the false arm `e3`.
* Update the final set `State.SameValue` to be the intersection of `StateTrueArm.SameValue` and `StateFalseArm.SameValue`.
* If `e1 ? e2 : e3` is a non-modifying expression that does not create a new object or read memory via a pointer, add `e1 ? e2 : e3` to the final set `State.SameValue`.

#### Note on resetting updated arm bounds to declared bounds:
The current implementation resets the observed bounds of any variables whose bounds were updated in either arm to the variable's declared bounds. For example:

```
int test;
array_ptr<int> large : count(3) = 0;
array_ptr<int> medium : count(2) = 0;
array_ptr<int> small : count(1) = 0;
(medium = small), test ? (medium = large) : medium;
```
At the end of checking the statement, `medium` has observed bounds `bounds(medium, medium + 2)`. The initial assignment `medium = small` does not result in an bounds checking error.

Note: if the answer to the TODO question above is true, i.e. the set of variables `v` whose bounds were modified in either arm `e2` or `e3` should be the set of all variables `v` for which `UC[v] != declared bounds of v`, then:
* The observed bounds context that is validated in the true arm `e2` would be `{ medium => bounds(large, large + 3) }` (no errors or warnings).
* The observed bounds context that is validated in the false arm `e3` would be `{ medium => bounds(small, small + 1) }` (error: inferred bounds `(small, small + 1)` do not imply declared bounds `(medium, medium + 2)`).
* The final observed bounds context after checking the statement would be `{ medium => bounds(medium, medium + 2), small => bounds(small, small + 1), large => bounds(large, large + 3) }` (no errors or warnings).

#### Testing:
* Added tests to bounds-context.c and equiv-expr-sets.c to test that the observed bounds context and set of sets of equivalent expressions are correctly updated when checking conditional operators.

#### Next steps:
* Infer rvalue bounds for conditional operators. The rvalue bounds for a conditional operator `e1 ? e2 : e3` is the greatest lower bound of the rvalue bounds of `e2` and `e3`.
* Handle uses of temporaries bound in only one arm.
* Treat logical operators `e1 && e2` and `e1 || e2` as forms of conditional operators.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -2409,7 +2409,8 @@ namespace {
 
             // For each variable v in ObservedBounds, check that the
             // observed bounds of v imply the declared bounds of v.
-            ValidateBoundsContext(S, BlockState, CSS);
+            ValidateBoundsContext(S, BlockState, WidenedBounds,
+                                  KilledBounds, CSS);
 
             // The observed bounds that were updated after checking S should
             // only be used to check that the updated observed bounds imply
@@ -3527,19 +3528,25 @@ namespace {
         // If checking each arm produces two different bounds contexts,
         // validate each arm's context separately.
 
+        BoundsMapTy WidenedBounds;
+        StmtDeclSetTy KilledBounds;
+
         // Validate the variables whose bounds were updated in the true arm.
         BoundsContextTy TrueUpdatedBounds = ContextDifference(
                                               StateTrueArm.ObservedBounds,
                                               State.ObservedBounds);
         StateTrueArm.ObservedBounds = TrueUpdatedBounds;
-        ValidateBoundsContext(E->getTrueExpr(), StateTrueArm, CSS);
+        ValidateBoundsContext(E->getTrueExpr(), StateTrueArm, WidenedBounds,
+                              KilledBounds, CSS);
 
         // Validate the variables whose bounds were updated in the false arm.
         BoundsContextTy FalseUpdatedBounds = ContextDifference(
                                                StateFalseArm.ObservedBounds,
                                                State.ObservedBounds);
         StateFalseArm.ObservedBounds = FalseUpdatedBounds;
-        ValidateBoundsContext(E->getFalseExpr(), StateFalseArm, CSS);
+        ValidateBoundsContext(E->getFalseExpr(), StateFalseArm, WidenedBounds,
+                              KilledBounds, CSS);
+
       }
 
       State.EquivExprs = IntersectEquivExprs(StateTrueArm.EquivExprs,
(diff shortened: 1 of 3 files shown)


mgrang (approved):
LGTM.

dtarditi (commented):
I think the logic for checking bounds declarations is missing a case where the test expression changes a variable involved in bounds, but an arm does not change the variable.

dtarditi (approved):
Looks good, thanks!

Issue #895, opened 2020-08-08
Assignments within conditional expressions should kill widened bounds


kkjeer:
Expressions within a conditional operator that modify variables used in a widened bounds expression currently do not kill the widened bounds. For example,

```
void f(nt_array_ptr<char> p : count(i), unsigned i, int flag) {
  if (*(p + i)) {
    // The i++ expression should kill the widened bounds of p.
    flag ? i++ : i;
  }
}
```

mgrang:
Fixed in https://github.com/microsoft/checkedc-clang/pull/898

mgrang:
@kkjeer Could you please verify if the above PR fixes the issue?

PR #898, opened 2020-08-12
Bounds Widening: Fix handling of bounds variables for nt_array_ptrs

mgrang:
There were 3 issues with the handling of variables used in the declared bounds
of an `nt_array_ptr`:

1. Bounds variables were being determined only for `nt_array_ptr`s declared in
the current block. As a result, if a variable is killed in a block then the
widened bounds in the successor blocks could not be killed. This is fixed by
storing the bounds variables along with the `nt_array_ptr`s for the entire
function instead of per block.

2. Bounds variables for function parameters were not getting determined
correctly. This is fixed by uniforming the logic to determine bounds
variables for function parameters and local variables.

3. Bounds declared as `CountBoundsExpr` were not being handled. This is fixed by
reading the `NormalizedBounds` instead of invoking `ExpandBoundsToRange`.

Also performed minor NFC code cleanups.

Fixes https://github.com/microsoft/checkedc-clang/issues/895

### DIFF
--- a/clang/lib/Sema/BoundsAnalysis.cpp
+++ b/clang/lib/Sema/BoundsAnalysis.cpp
@@ -217,44 +217,39 @@ void BoundsAnalysis::CollectBoundsVars(const Expr *E, DeclSetTy &BoundsVars) {
     CollectBoundsVars(BO->getRHS(), BoundsVars);
   }
 
-  if (IsDeclOperand(E)) {
-    const DeclRefExpr *D = GetDeclOperand(E);
+  if (DeclRefExpr *D = GetDeclOperand(E))
     if (const auto *V = dyn_cast<VarDecl>(D->getDecl()))
       BoundsVars.insert(V);
-  }
 }
 
 DeclRefExpr *BoundsAnalysis::GetDeclOperand(const Expr *E) {
-  if (!E || !isa<CastExpr>(E))
-    return nullptr;
-  auto *CE = dyn_cast<CastExpr>(E);
-  assert(CE->getSubExpr() && "invalid CastExpr expression");
-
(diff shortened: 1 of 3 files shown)


clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -273,13 +268,20 @@ void BoundsAnalysis::CollectNtPtrsInScope(FunctionDecl *FD) {
...
+  // Collect the variables used in the declared bounds expr for each ntptr.
+  for (const VarDecl *V : VarList) {
+    DeclSetTy BoundsVars;
+    CollectBoundsVars(S.NormalizeBounds(V), BoundsVars);
+    NtPtrsInScope[V] = BoundsVars;
+  }
 }
 
-void BoundsAnalysis::FillGenSetAndGetBoundsVars(const Expr *E,
+void BoundsAnalysis::FillGenSetForEdge(const Expr *E,
```
dtarditi, line 284: It looks like indentation needs to be fixed here.

clang/test/CheckedC/inferred-bounds/widened-bounds.c
```diff
@@ -490,7 +489,7 @@ void f20() {
 // CHECK:  [B9]
 // CHECK: upper_bound(q) = 1
 
-  // Declared bounds (INT_MIN) and deref offset (INT_MAX - 1). No sequential deref tests. No widening.
+  // Declared bounds
```
dtarditi, line 492: I don't understand why you deleted the rest of the comment.  Same goes for other changes in this file.

mgrang: Thanks for catching this. The comments should not have been deleted. Not sure how that happened. I will fix this.

PR #899, opened 2020-08-13
Bind all compound literals to temporaries

### LINKED ISSUE #870: Introduce temporary bindings for struct-typed compound literals

kkjeer:
Fixes #870 

This PR introduces temporary bindings for all compound literals. It also infers lvalue bounds for struct-typed compound literals: for a struct-typed compound literal expression `e`, the lvalue bounds of `e` are `bounds(&value(temp(e)), &value(temp(e)) + 1)`.

#### Testing:
* Modified bounds-context and compound-literals tests to expect compound literals and lvalue bounds for struct-typed compound literals
* Added checkedc tests in [checkedc/414](https://github.com/microsoft/checkedc/pull/414)

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -3689,12 +3689,7 @@ namespace {
         if (E->getType()->isStructureType()) {
           Expr *TempUse = CreateTemporaryUse(E);
           Expr *Addr = CreateAddressOfOperator(TempUse);
-          const llvm::APInt One(Context.getTargetInfo().getPointerWidth(0), 1);
-          IntegerLiteral *Size = CreateIntegerLiteral(One);
-          CountBoundsExpr *CBE = new (Context)
-              CountBoundsExpr(BoundsExpr::Kind::ElementCount, Size,
-                              SourceLocation(), SourceLocation());
-          return ExpandToRange(Addr, CBE);
+          return ExpandToRange(Addr, Context.getPrebuiltCountOne());
         }
 
         // Only expressions with array or function type can have a decayed
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -3692,9 +3692,9 @@ namespace {
           return ExpandToRange(Addr, Context.getPrebuiltCountOne());
         }
 
-        // Only expressions with array or function type can have a decayed
-        // type, which is used to create the lvalue bounds.
-        if (E->getType()->isArrayType() || E->getType()->isFunctionType()) {
+        // The lvalue bounds of an array-typed compound literal expression e
+        // are based on the dimension size of e.
+        if (E->getType()->isArrayType()) {
           BoundsExpr *BE = CreateBoundsForArrayType(E->getType());
           QualType PtrType = Context.getDecayedType(E->getType());
           Expr *ArrLValue = CreateTemporaryUse(E);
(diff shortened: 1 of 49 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3684,21 +3684,33 @@ namespace {
...
+        // are bounds(&value(temp(e), &value(temp(e)) + 1).
+        if (E->getType()->isStructureType()) {
+          Expr *TempUse = CreateTemporaryUse(E);
+          Expr *Addr = CreateAddressOfOperator(TempUse);
+          const llvm::APInt One(Context.getTargetInfo().getPointerWidth(0), 1);
+          IntegerLiteral *Size = CreateIntegerLiteral(One);
+          CountBoundsExpr *CBE = new (Context)
+              CountBoundsExpr(BoundsExpr::Kind::ElementCount, Size,
+                              SourceLocation(), SourceLocation());
+          return ExpandToRange(Addr, CBE);
```
dtarditi, line 3697: There's a prebuilt Count(1) expression available in ASTContext.  See ASTContext::getPrebuiltCountOne().   This avoids an extra allocation in ASTContext.  The code constructs a count(1) expression, but then calls ExpandToRange, which produces a range expression.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -3684,21 +3684,33 @@ namespace {
...
+
         // Only expressions with array or function type can have a decayed
-        // type, which is used to create the lvalue bounds.  Compound literals
-        // with non-array, non-function types do not have lvalue bounds.
-        // TODO: checkedc-clang issue #870: bind all compound literals to
-        // temporaries and infer lvalue bounds for struct compound literals.
-        if (!(E->getType()->isArrayType() || E->getType()->isFunctionType()))
-          return CreateBoundsAlwaysUnknown();
+        // type, which is used to create the lvalue bounds.
+        if (E->getType()->isArrayType() || E->getType()->isFunctionType()) {
```
dtarditi, line 3702: I don't think function types should have bounds, so I'm confused by the prior code.  Should function types be given unknown bounds instead?

PR #901, opened 2020-08-14
Compare Expressions: Create a class hierarchy for AST node type [4/n]

mgrang:
We generalize the node type into the following class hierarchy:

```
       Node
      /    \
     /      \
 BinaryNode LeafExprNode
```

`BinaryNode` can have two or more `Node`s as children. The `LeafExpreNode` type wraps
a clang AST expression.


mgrang:
I am working on adding unit tests to this PR.

PR #903, opened 2020-08-18
Explain bounds proof failure caused by free variables in inferred/declared bounds

dopelsunce:
This PR enables the compiler to emit an error when a _free variable_ (defined below) is found in the declared/inferred bounds on bounds assignments, and passing arguments. A note is printed for each free variable found.

### Free variables w.r.t. expressions
Assume the helper function `VarSet(E)` returns the set of variables in expression `E`.

Given two expressions `SrcExpr` and `DstExpr`, a set of equivalent expression sets `EQ`. A variable `V` in `VarSet(SrcExpr)` is _free_ if:
1. (`V` does not equal to some constant) `V` is not an integer type, or there is no set in `EQ` that contains a constant and `V`; and,
1. (`V` does not have an alias in `VarSet(DstExpr)`) For each `U` in `VarSet(DstExpr)`, there is not set in `EQ` that contains both `V` and `U`.

### Free variables w.r.t. declared/inferred bounds

We extend the above definition for free variables with respect to the lower (or upper bounds) of the declared bounds (or inferred bounds). In other words, a variable used by the declared lower bounds may be free even if it is used in the inferred upper bounds.

### Affected cases
Free variable is detected in the following cases:
1. On variable initialization or assignment, inferred bounds do not imply declared bounds. Error msg: "it is not possible to prove that the inferred bounds of p imply the declared bounds of p after ...".
1. On passing arguments, the inferred bounds do not imply the expected argument bounds. Error msg: "it is not possible to prove argument meets declared bounds for n-th parameter"

__Note__: we do not check free variables in static bounds cast.

### Note for each free variable

We use the template: "note: the declared/inferred bounds use the value of the variable 'q', and there is no relational information between 'q' and any of the variables used by the inferred/declared bounds".


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1238,6 +1373,149 @@ namespace {
...
+      //   1. either (1) V is not an integer type, or (2) there is no set in
+      //   EquivExprs that contains both V and a constant (i.e. an integer-typed
+      //   value); and
+      //   2. for each variable U in DstVars, there is not set in EquivExprs
+      //   that contains both V and U.
+      //
+      // GetFreeVariables returns true if any free variable is found in SrcVars,
+      // and appends the free variables to FreeVariables.
+      static bool GetFreeVariables(Sema &S, const EqualExprTy &SrcVars,
+                            const EqualExprTy &DstVars,
```
kkjeer, line 1436: Align the start of the parameters on the same column.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1172,6 +1270,43 @@ namespace {
         return ProofResult::Maybe;
       }
 
+      // InRangeWithFreeVars is the same as InRange, but gathers free variables
```
kkjeer, line 1273: Why do both `InRangeWithFreeVars` and `InRange` exist? Is there a context in which `InRange` would be called but `InRangeWithFreeVars` would not?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1238,6 +1373,149 @@ namespace {
         return ProofResult::Maybe;
       }
 
+      // CompareLowerOffsets first calls CompareLowerOffsetsImpl. If
+      // CompareLowerOffsetsImpl returns Maybe, it continues to collect free
+      // variables in the lower offsets of this and R.
+      ProofResult
+      CompareLowerOffsets(BaseRange &R, ProofFailure &Cause,
```
kkjeer, line 1380: Align the method name on the same line as the return type.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1238,6 +1373,149 @@ namespace {
...
+        return HasFreeVariables;
+      }
+
+      // Check free variables between E1 and E2. Append any found free variables
+      // to FreeVars with each free variable in E1 (resp. E2) having Pos1 (resp.
+      // Pos2).
+      static bool CheckFreeVarInExprs(Sema &S, Expr *E1, Expr *E2, FVPos Pos1,
+                                      FVPos Pos2, EquivExprSets *EquivExprs,
+                                      FreeVariableListTy &FreeVars) {
+        // Do not check free variables in bounds that contain indirect accesses.
```
kkjeer, line 1467: This comment is somewhat confusing since `E1` and `E2` are not necessarily bounds expressions. I suggest changing to "Do not check free variables in expressions that contain indirect accesses.".

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1238,6 +1373,149 @@ namespace {
...
+        if ((E1 && ReadsMemoryViaPointer(E1)) || (E2 && ReadsMemoryViaPointer(E2)))
```
kkjeer, line 1468: What happens if `E1` (or `E2`) is null?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1655,12 +1933,14 @@ namespace {
...
-                                        const BoundsExpr *SrcBounds,
-                                        ProofFailure &Cause,
-                                        EquivExprSets *EquivExprs,
-                                        ProofStmtKind Kind =
-                                          ProofStmtKind::BoundsDeclaration) {
+    // 
+    // If any free variable is found in SrcBounds or DeclaredBounds, return
+    // False and add the free variables to FreeVariables.
+    ProofResult ProveBoundsDeclValidity(
+        const BoundsExpr *DeclaredBounds, const BoundsExpr *SrcBounds,
```
kkjeer, line 1940: Please align the parameters with the method name.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1238,6 +1373,149 @@ namespace {
...
+      // both Variable and a constant (i.e. IntegerType).
+      static bool IsEqualToConstant(Sema &S, Expr *Variable,
+                                    const EquivExprSets *EquivExprs) {
+        if (Variable->getType()->isPointerType() || !EquivExprs)
+          return false;
+
+        EqualExprTy EquivSet =
+            GetEqualExprSetContainingExpr(S, Variable, *EquivExprs);
+        for (const auto &E : EquivSet) {
+          if (E->getType()->isIntegerType())
```
kkjeer, line 1512: This function will return true if there is a set in `EquivExprs` that contains `Variable` and any other integer-typed expression (e.g. an integer-typed variable), not necessarily a constant. For example:

```
void f(int a, int b) {
    a = b;
}
```

`EquivExprs` contains the set `{ a, b }`. Since `b` has integer type, `IsEqualToConstant(a, EquivExprs)` will return true.

I suggest looking at the `Expr::isIntegerConstantExpr` method to test whether an expression is an integer constant.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -655,6 +703,31 @@ namespace {
   // expressions that produce the same value as e.
   using ExprEqualMapTy = llvm::DenseMap<Expr *, EqualExprTy>;
 
+  // Describes the position of a free variable (FR).
+  enum class FVPos {
+    Lower = 0x1,    // The FR appears in (any) lower bounds.
+    Upper = 0x2,    // The FR appears in (any) upper bounds.
+    Base = Lower | Upper, // The FR appears in (any) base bounds.
```
kkjeer, line 710: Instead of defining each combination here, I suggest only defining `Lower`, `Upper`, `Base`, and `Declared`, and creating the other combinations where they are needed.

clang/lib/Sema/SemaBounds.cpp
dopelsunce: `InRange` is also called in `ProveMemoryAccessInRange` (https://github.com/microsoft/checkedc-clang/blob/master/clang/lib/Sema/SemaBounds.cpp#L1728), where we don't check free variables.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -5291,12 +5308,15 @@ namespace {
       }
     }
 
-    // Returns true if the expression e reads memory via a pointer.
-    // IncludeAllMemberExprs is used to modify the behavior to return true
-    // if e is or contains a pointer dereference, member reference, or
+    // Returns true if the expression e reads memory via a pointer, or e is
```
kkjeer, line 5311: The comment indicates that `ReadsMemoryViaPointer` returns true if `e` is null, but it actually returns false.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2081,19 +2442,30 @@ namespace {
...
+          TargetBounds, SrcBounds, Cause, nullptr, FreeVars, Kind);
       if (Result != ProofResult::True) {
-        unsigned DiagId = (Result == ProofResult::False) ?
-          diag::error_static_cast_bounds_invalid :
-          (CSS != CheckedScopeSpecifier::CSS_Unchecked ?
-           diag::warn_checked_scopestatic_cast_bounds_invalid :
-           diag::warn_static_cast_bounds_invalid);
+        // Which diagnostic message to print?
+        unsigned DiagId =
+            (Result == ProofResult::False)
```
kkjeer, line 2451: Upon reflection, I actually don't think it makes sense to print an error for free variables in static pointer casts. You can just print the existing `error_static_cast_bounds_invalid` error message instead.

clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -10229,6 +10249,9 @@ def err_bounds_type_annotation_lost_checking : Error<
   def error_static_cast_bounds_invalid : Error<
     "cast source bounds are too narrow for %0">;
 
+  def error_static_cast_bounds_unprovable : Error<
```
kkjeer, line 10252: I think we should only use the existing error message for static bounds casts instead of adding an error for free variables (see comment below).

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -565,6 +565,57 @@ namespace {
...
+      EqualExprTy VariableList;
+
+  public:
+      CollectVariableSetHelper(Sema &SemaRef)
+          : SemaRef(SemaRef), VariableList() {}
+
+      const EqualExprTy &GetVariableList() const { return VariableList; }
+
+      bool VisitDeclRefExpr(DeclRefExpr *E) {
+      // We cast variables to rvalues so they can be compared with rvalues in EquivExprSet.
```
kkjeer, line 598: Please indent code within this method.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1719,7 +2014,9 @@ namespace {
           }
         }
         return R;
-      }
+      } else
+        // Failed to extract a common base.
+        Cause = CombineFailures(Cause, ProofFailure::NoBaseRange);
```
kkjeer, line 2019: Is the `NoBaseRange` cause used to alter any diagnostic messages?

clang/test/CheckedC/static-checking/bounds-side-effects.c
```diff
@@ -91,11 +92,12 @@ void f5(int i) {
 // Test different forms of bounds declarations.
 void f20(int len, _Array_ptr<int> p : count(len), int i) {
   len = i, p = alloc(i * sizeof(int));  // correct
-  len = 5;                              // incorrect
+  len = 5;                              // expected-error {{it is not possible to prove that the inferred bounds of 'p' imply the declared bounds of 'p' after assignment}}
 }
 
 void f21(int len, _Array_ptr<int> p : byte_count(len), int i) {
-  len = i * sizeof(int), p = alloc(i * sizeof(int)); // correct
+  len = i * sizeof(int), p = alloc(i * sizeof(int)); // correct \
```
kkjeer, line 99: Remove `// correct` in cases where there is now an expected error.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4015,15 +4388,25 @@ namespace {
         if (I->second.find(V) == I->second.end())
           return;
       }
+      
+      // Which diagnostic message to print?
+      unsigned DiagId =
```
kkjeer, line 4393: As we discussed offline, we may want to prioritize width-related error messages over free variable-related error messages. This is a follow-up item tracked in #910.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1699,7 +1993,14 @@ namespace {
         llvm::outs() << "\nSource range:";
         SrcRange.Dump(llvm::outs());
 #endif
-        ProofResult R = SrcRange.InRange(DeclaredRange, Cause, EquivExprs, Facts);
+        // For static bounds cast, we do not check free variables.
```
kkjeer, line 1996: This comment is somewhat confusing since it's not clear that checking static bounds cast will always pass a nullptr EquivExprs parameter. I suggest moving this logic into the `InRangeWithFreeVars` function, so that in `InRangeWithFreeVars`, if `EquivExprs` is null, simply return `InRange` without checking any free variables.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2081,19 +2448,30 @@ namespace {
...
-          diag::error_static_cast_bounds_invalid :
-          (CSS != CheckedScopeSpecifier::CSS_Unchecked ?
-           diag::warn_checked_scopestatic_cast_bounds_invalid :
-           diag::warn_static_cast_bounds_invalid);
+        // Which diagnostic message to print?
+        unsigned DiagId =
+            (Result == ProofResult::False)
+                ? (TestFailure(Cause, ProofFailure::HasFreeVariables)
+                       ? diag::error_static_cast_bounds_unprovable
+                       : diag::error_static_cast_bounds_invalid)
```
kkjeer, line 2460: I believe the `error_static_cast_bounds_unprovable` error code should be deleted (delete this as well from DiagnosticSemaKinds.td). If the proof result is false, only the existing `errror_static_cast_bounds_invalid` message should be emitted.

clang/test/CheckedC/static-checking/bounds-decl-checking-bsi.c
```diff
@@ -64,7 +66,9 @@ _Array_ptr<int> f14(unsigned num) {
 }
 
 _Array_ptr<int> f15(unsigned num1, unsigned num2){
-  _Array_ptr<int> p : byte_count(num1) = test_f13(num2); // expected-warning {{cannot prove declared bounds for 'p' are valid after initialization}} \
+  _Array_ptr<int> p : byte_count(num1) = test_f13(num2); // expected-error {{it is not possible to prove that the inferred bounds of 'p' imply the declared bounds of 'p' after initialization}} \
+                                                         // expected-note {{the declared upper bounds use the value of the variable 'num1', and there is no relational information between 'num1' and any of the variables used by the inferred upper bounds}} \
```
dtarditi, line 70: The notes about the bounds should be displayed before the note about lack of relational information.  This will be confusing for a programmer to follow.   This can be addressed in follow-up work.

dtarditi (commented):
Thanks for the work.   I have another suggestion in addition to @kkjeer's suggestions.   This could be done as follow-up work.

clang/test/CheckedC/static-checking/bounds-decl-checking-bsi.c
kkjeer: Thanks for the suggestion @dtarditi. Created issue #912 to track this.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1066,15 +1131,27 @@ namespace {
       DstEmpty = 0x10,      // The destination bounds are empty (LB == UB).
       DstInvalid = 0x20,    // The destination bounds are invalid (LB > UB).
       Width = 0x40,         // The source bounds are narrower than the destination bounds.
-      PartialOverlap = 0x80 // There was only partial overlap of the destination bounds with
+      PartialOverlap = 0x80, // There was only partial overlap of the destination bounds with
                             // the source bounds.
+      HasFreeVariables = 0x200, // Source or destination has free variables.
```
kkjeer, line 1136: Nit: change this value to `0x100` and remove trailing comma.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1238,6 +1376,175 @@ namespace {
...
+        if (ReadsMemoryViaPointer(E1) || ReadsMemoryViaPointer(E2))
+          return false;
+
+        bool HasFreeVariables = false;
+        EqualExprTy Vars1 = CollectVariableSet(S, E1);
+        EqualExprTy Vars2 = CollectVariableSet(S, E2);
+
+        // EquivVars holds sets of DeclRefExpr and IntegerLiteral filtered from
+        // EquivExprs.
+        EquivExprSets EquivVars;
```
dtarditi, line 1492: An implication of my comment is that we shouldn't be filtering EquivVars.  We could be removing expressions that have equality relationships and that use variables.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1238,6 +1376,175 @@ namespace {
...
+      // GetFreeVariables gathers "free variables" in SrcVars.
+      //
+      // Given two variable sets SrcVars and DstVars, and a set of equivalent
+      // sets of Expr EquivExprs. A variable V in SrcVars is *free* if the two
+      // conditions are met:
+      //   1. either (1) V is not an integer type, or (2) there is no set in
+      //   EquivExprs that contains both V and a constant (i.e. an integer-typed
+      //   value); and
+      //   2. for each variable U in DstVars, there is not set in EquivExprs
+      //   that contains both V and U.
```
dtarditi, line 1444: I think condition 2 needs to be amended to work in terms of uses of variables.   There is not a set in EquivExprs that contains two different expressions, where one expression uses V and the other expression uses U.

clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -10172,6 +10172,23 @@ def err_bounds_type_annotation_lost_checking : Error<
...
 
+  def error_bounds_declaration_unprovable : Error<
+    "it is not possible to prove that the inferred bounds of %1 "
+    "imply the declared bounds of %1 after "
+    "%select{assignment|decrement|increment|initialization|statement}0">;
+
+  def note_free_variable_decl_or_inferred : Note<
+    "the %select{declared|inferred}0 %select{lower |upper |}1bounds use the value "
+    "of the variable '%2', and there is no relational information between '%2' "
+    "and any of the variables used by the %select{inferred|declared}0 "
```
dtarditi, line 10183: This may need to be reworded.  There is no relational information involving v and any of the expressions used by the other bounds.  Perhaps something along the lines of 
```
"the %select{declared|inferred}0 %slect{lower | upper |}bound use the variable %2' and there is no relational information involving '%2' and any of expressions used by the %select{inferred|declared}0 "

**Review by dtarditi — COMMENTED** [review]
The algorithm seems to have subtle complications that I didn't realize when we first discussed the idea.  The basic idea is just to look at sets of variables, but there are subtleties involving expressions using those variables.  Expressions that use those variables may have relational information between each other and may appear in equivalent expression sets.  This relational information is being lost when we boil things down to just sets of variables.

@kkjeer and @dopelsunce, it may be worth discussing next steps before making changes.   I think this may address some of the error messages that we're seeing be generated now.

File: clang/test/CheckedC/inferred-bounds/basic.c [hunk 23]
```diff
@@ -426,7 +426,7 @@ void f42(void) {
 // CHECK: | `-DeclRefExpr {{0x[0-9a-f]+}} '_Array_ptr<int>' lvalue Var {{0x[0-9a-f]+}} 'p' '_Array_ptr<int>'
 // CHECK: `-IntegerLiteral {{0x[0-9a-f]+}} 'int' 1
 
-  _Array_ptr<int> r : count(1) = &p[0];
+  _Array_ptr<int> r : count(1) = &p[0]; // expected-error {{it is not possible to prove that the inferred bounds of 'r' imply the declared bounds of 'r' after initialization}}
```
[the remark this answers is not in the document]
kkjeer: The set `EquivExprs` contains `{ (_Array_ptr<int>)(&p[0]), r }`. Without the changes in this PR, there would be a warning: "cannot prove declared bounds for 'r' are valid after initialization" since the bases `p` and `r` of the inferred bounds `(p, p + 1)` and declared bounds `(r, r + 1)` are not equal.

clang/test/CheckedC/inferred-bounds/basic.c
```diff
@@ -490,7 +490,7 @@ void f43(void) {
 // CHECK: | `-DeclRefExpr {{0x[0-9a-f]+}} 'int [5][5]' lvalue Var {{0x[0-9a-f]+}} 'arr' 'int [5][5]'
 // CHECK: `-IntegerLiteral {{0x[0-9a-f]+}} 'int' 5
 
-  p = arr[0];
+  p = arr[0]; // expected-error {{it is not possible to prove that the inferred bounds of 'p' imply the declared bounds of 'p' after assignment}}
```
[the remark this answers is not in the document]
kkjeer: `EquivExprs` here contains `{ arr[0], p }`. The bases `arr` and `p` of the inferred bounds `(arr, arr + 5)` and declared bounds `(p, p + 5)` are not equal, resulting in a warning without this PR and a free variable error with this PR.

clang/test/CheckedC/inferred-bounds/basic.c
```diff
@@ -517,7 +517,7 @@ void f43(void) {
 // CHECK:   | `-DeclRefExpr {{0x[0-9a-f]+}} 'int [5][5]' lvalue Var {{0x[0-9a-f]+}} 'arr' 'int [5][5]'
 // CHECK:   `-IntegerLiteral {{0x[0-9a-f]+}} 'int' 5
 
-  r = &p[0];
+  r = &p[0]; // expected-error {{it is not possible to prove that the inferred bounds of 'r' imply the declared bounds of 'r' after assignment}}
```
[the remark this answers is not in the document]
kkjeer: `EquivExprs` here contains `{ arr[0], p }`. The bases `p` and `arr` of the inferred bounds `(p, p + 5)` and declared bounds `(arr, arr + 1)` are not equal, resulting in a warning without this PR and a free variable error with this PR.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -565,6 +565,53 @@ namespace {
...
+  }
+
+  // Helper class for collecting a vector of unique variables as rvalues from an
+  // expression. We collect rvalues because CheckingState.EquivExprSet uses
+  // rvalues to check equality.
+  class CollectVariableSetHelper
+    : public RecursiveASTVisitor<CollectVariableSetHelper> {
+  private:
+    Sema &SemaRef;
+    EqualExprTy VariableList;
```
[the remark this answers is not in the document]
kkjeer: `EqualExprTy` should be renamed to `ExprSetTy` or something similar. I've opened issue #927 to track this.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1159,15 +1247,65 @@ namespace {
...
+      }
+
+      // InRangeWithFreeVars is the same as InRange, but gathers free variables
+      // betweeen the source and declared BaseRanges in addition.
+      ProofResult
+      InRangeWithFreeVars(BaseRange &R, ProofFailure &Cause,
+                          EquivExprSets *EquivExprs,
+                          std::pair<ComparisonSet, ComparisonSet> &Facts,
+                          FreeVariableListTy &FreeVariables) {
+        // If we do not have equivalence relational information (EquivExprs
```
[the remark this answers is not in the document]
dtarditi: @kkjeer, could you update this comment?   Thanks.

clang/lib/Sema/SemaBounds.cpp
dtarditi: @kkjeer, could you update this comment?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -565,6 +565,53 @@ namespace {
...
+    EqualExprTy VariableList;
+
+  public:
+    CollectVariableSetHelper(Sema &SemaRef)
+      : SemaRef(SemaRef), VariableList() {}
+
+    const EqualExprTy &GetVariableList() const { return VariableList; }
+
+    bool VisitDeclRefExpr(DeclRefExpr *E) {
+      if (!EqualExprsContainsExpr(SemaRef, VariableList, E, nullptr)) {
```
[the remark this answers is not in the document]
dtarditi: @kkjeer, is my comment still accurate?  If so, could you leave a comment in the code about the algorithm being quadratic.  I don't like to leave quadratic algorithms around in a compiler without calling them out.

dtarditi (commented):
There are a few suggestions around comments that I'd like to see resolved.  Also, there is a merge conflict according to GitHub.   Once both of these are done, I'll approve the merge.

PR #913, opened 2020-09-05
Add expected errors in widened-bounds-semantic-compare

kkjeer:
After the changes in #903, the compiler will emit an error where free variables are found in the declared or inferred bounds. There are several places in the widened-bounds-semantic-compare test where the declared bounds involve free variables. The other cases in this test file do not result in errors because the compiler is not able to create a base range for the declared bounds (so a warning is emitted, which the test file ignores).


mgrang (approved):
LGTM.

PR #915, opened 2020-09-11
Bounds widening for if (!*p)

emoham:
In this PR we add support for widening the bounds for if (!*p) along with a few tests.


mgrang:
This functionality is already present on master.

Issue #911, opened 2020-09-04
Handling C style casts in checking free variables


dopelsunce:
```
void f() {
  int a1 checked[5];
  short int t20 : byte_count(5 * sizeof(int)) = (short int)a1;
}
```

`a1` and `t20` are incorrectly reported as free variables by #903.

This causes one test to fail: checkedc/tests/typechecking/bounds.c

When checking the assignment,
`EquivExprs` contains the set `{ LValueToRValue(t20), (short int)LValueToRValue(a1) }` and we're searching for the variable `LValueToRValue(a1).` Thus `a1` and `t20` are treated unequal.

**Fix**:  this could be part of #909 to add DeclRefExprs to the variable set, and then search the EquivExprs for an expression that is the DeclRefExpr ignoring casts and parentheses.

Basically add a `IsRValueCastOfVariableIgnoringCasts` method that would return true for `(short int)LValueToRValue(a1)` and `DeclRefExpr(a1)`

kkjeer:
Fixed by https://github.com/microsoft/checkedc-clang/commit/867cf58ba22c37b7b6bb30051a10887ec78a74ce.

PR #919, opened 2020-09-28
Free variables: fix #909 and #911 by ignoring casts in EquivExprs

### LINKED ISSUE #909: Add lvalue DeclRefExprs to variable set for diagnosing free variables
Currently in #903, getting the set of all variables involved in an expression constructs an LValueToRValue ImplicitCastExpr for each variable to add to the set. This is done so that the expressions in the set are rvalues and can later be compared with other rvalue expressions in `CheckingState.EquivExprs`. We want to avoid constructing these ImplicitCastExprs.

`CollectVariableSetHelper` should collect a set of DeclRefExprs. To check whether there is a set in `CheckingState.EquivExprs` that contains an lvalue variable `v`, search each expression `e` in each set `F` in `CheckingState.EquivExprs` to see if `e` is an LValueToRValue cast of `v`. The `IsRValueCastOfVariable(e, v)` method can be used to check this.

dopelsunce:
On checking free variables, create a temporary `EquivVars` that only holds DecRefExpr and IntegerLiteral from the `EquivExprs`, with casts ignored. The variables in `EquivVars` are then used to compare with the variables (DecRefExpr) collected from bounds expressions.

IntegerLiterals from `EquivExprs` are also kept in `EquivVars` so we can check whether a variable equals to a constant.

### DIFF
--- a/clang/lib/AST/CanonBounds.cpp
+++ b/clang/lib/AST/CanonBounds.cpp
@@ -310,7 +310,7 @@ Result Lexicographic::CompareExpr(const Expr *Arg1, const Expr *Arg2) {
 #include "clang/AST/StmtNodes.inc"
        llvm_unreachable("cannot compare a statement");  
      case Expr::PredefinedExprClass: Cmp = Compare<PredefinedExpr>(E1, E2); break;
-     case Expr::DeclRefExprClass: return Compare<DeclRefExpr>(E1, E2);
+     case Expr::DeclRefExprClass: Cmp = Compare<DeclRefExpr>(E1, E2); break;
      case Expr::IntegerLiteralClass: return Compare<IntegerLiteral>(E1, E2);
      case Expr::FloatingLiteralClass: return Compare<FloatingLiteral>(E1, E2);
      case Expr::ImaginaryLiteralClass: break;
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -571,12 +571,12 @@ namespace {
   // EqualExprsContainsExpr returns true if the set Exprs contains E.
   bool EqualExprsContainsExpr(Sema &S, const EqualExprTy Exprs, Expr *E,
                               EquivExprSets *EquivExprs) {
-      for (auto I = Exprs.begin(); I != Exprs.end(); ++I) {
+    for (auto I = Exprs.begin(); I != Exprs.end(); ++I) {
       if (Lexicographic(S.Context, EquivExprs).CompareExpr(*I, E) ==
-          Lexicographic::Result::Equal)
-          return true;
-      }
-      return false;
+        Lexicographic::Result::Equal)
+        return true;
+    }
+    return false;
   }
 
   // Helper class for collecting a vector of unique variables as rvalues from an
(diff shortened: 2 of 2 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1492,10 +1487,24 @@ namespace {
...
 
-        if (AddFreeVariables(S, Vars1, Vars2, EquivExprs, Pos1, FreeVars))
+        // EquivVars holds sets of DeclRefExpr and IntegerLiteral filtered from
+        // EquivExprs.
+        EquivExprSets EquivVars;
+        for (auto ExprSet : *EquivExprs) {
+          EqualExprTy Vars;
+          auto It = ExprSet.begin();
+          for (; It != ExprSet.end(); It++) {
+            *It = (*It)->IgnoreCasts();
```
kkjeer, line 1497: This also needs to ignore parentheses: 

```
Expr *E = (*It)->IgnoreParenCasts();
if (isa<IntegerLiteral>(E) || isa<DeclRefExpr>(E))
  Vars.push_back(E);
``` 

Otherwise, the following test case fails:

```
int a1 checked[5];
short int t20 : byte_count(5 * sizeof(int)) = ((((short int)(a1))));
```

clang/lib/AST/CanonBounds.cpp
```diff
@@ -310,7 +310,7 @@ Result Lexicographic::CompareExpr(const Expr *Arg1, const Expr *Arg2) {
 #include "clang/AST/StmtNodes.inc"
        llvm_unreachable("cannot compare a statement");  
      case Expr::PredefinedExprClass: Cmp = Compare<PredefinedExpr>(E1, E2); break;
-     case Expr::DeclRefExprClass: return Compare<DeclRefExpr>(E1, E2);
+     case Expr::DeclRefExprClass: Cmp = Compare<DeclRefExpr>(E1, E2); break;
```
kkjeer, line 313: Why this change?

kkjeer (commented):
Thanks for working on this! I have a comment about removing parentheses.

clang/lib/AST/CanonBounds.cpp
dopelsunce: Previously the code just returns here. This adds a new path so that if `Cmp` is not Equal, we proceed to use `CheckEquivExprs` to check if E1 and E2 produce the same value using EquivExprSets. (https://github.com/microsoft/checkedc-clang/pull/919/files#diff-9ff223384e8cae5a65b76ee59e13437eR376)

kkjeer: Is there a test case that fails if this change is not made?

dopelsunce:
```
void f() {
  int i1, i2;
  i1 = i2;
  _Array_ptr<int> p1 : count(i1 + 1) = 0;
  _Array_ptr<int> p2 : count(i2) = p1;
}
```

Without this change, `i1` and `i2` will be incorrectly reported as free variables in the code above, because`Compare<DeclRefExpr>` does not use the information from EquivExprSet to check if `i1` and `i2` can produce the same value. Once  `Compare<DeclRefExpr>` returns false, `CheckEquivExprs` also returns without using the information from the sets of equivalent expressions.

Previously in #903, we compare two ImplicitCastExprs rather than two DeclRefExprs, so this change was not needed. Now we directly compare DeclRefExprs collected from bounds expression, which is a different case for CheckEquivExprs.

kkjeer: Thanks for clarifying!

kkjeer (approved):
Thanks for fixing this!

PR #920, opened 2020-09-29
Compare Expressions: Normalize subtract operator & expressions w/o const [7/n]

mgrang:
We perform two normalizations of expressions:

1. Normalize `(e1 - e2)` to `(e1 + -e2)`. This will avoid the use of subtract
operators and ease constant folding.

2. Normalize expressions which do not have any integer constants. For example:
  `p ==> p + 0`
  `p + i ==> p + i + 0`
  `p * i ==> p * i * 1`
  This will make comparison of two expressions for equality easier.

### DIFF
--- a/clang/lib/AST/PreorderAST.cpp
+++ b/clang/lib/AST/PreorderAST.cpp
@@ -93,9 +93,9 @@ void PreorderAST::Create(Expr *E, Node *Parent) {
     // and then check if the resultant expression -e2 overflows. If it
     // overflows, we undo the unary minus operator.
 
-    // TODO: Currently we only handle detection of overflow for integer
-    // constant expressions. We need to handle the case where e2 is a variable
-    // expression which is not known until run time.
+    // TODO: Currently, we can only prove that integer constant expressions do
+    // not overflow. We still need to handle proving that non-constant
+    // expressions do not overflow.
     if (BO->getOpcode() == BO_Sub &&
         RHS->isIntegerConstantExpr(Ctx)) {
       Expr *UOMinusRHS = new (Ctx) UnaryOperator(RHS, UO_Minus, RHS->getType(),
--- a/clang/lib/AST/PreorderAST.cpp
+++ b/clang/lib/AST/PreorderAST.cpp
@@ -29,24 +29,30 @@ void PreorderAST::AddNode(Node *N, Node *Parent) {
   }
 }
 
-void PreorderAST::CoalesceNode(BinaryNode *B, BinaryNode *Parent) {
+void PreorderAST::RemoveNode(Node *N, Node *Parent) {
+  assert(isa<BinaryNode>(Parent) && "Invalid parent");
+
+  auto *P = dyn_cast<BinaryNode>(Parent);
+
   // Remove the current node from the list of children of its parent.
-  for (auto I = Parent->Children.begin(),
-            E = Parent->Children.end(); I != E; ++I) {
-    if (*I == B) {
-      Parent->Children.erase(I);
+  for (auto I = P->Children.begin(),
+            E = P->Children.end(); I != E; ++I) {
+    if (*I == N) {
+      P->Children.erase(I);
       break;
     }
   }
 
-  // Move all children of the current node to its parent.
-  for (auto *Child : B->Children) {
-    Child->Parent = Parent;
-    Parent->Children.push_back(Child);
+  if (auto *B = dyn_cast<BinaryNode>(N)) {
+    // Move all children of the current node to its parent.
+    for (auto *Child : B->Children) {
+      Child->Parent = P;
+      P->Children.push_back(Child);
+    }
   }
 
   // Delete the current node.
-  delete B;
+  delete N;
 }
 
 void PreorderAST::Create(Expr *E, Node *Parent) {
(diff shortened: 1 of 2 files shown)


mgrang:
I cannot add test cases for this PR because currently there is no code that exercise this path. I would be able to add tests cases in the follow-up PR which will make use of the code in this PR.

mgrang:
I cannot add test cases for this PR because currently there is no code that exercise this path. I would be able to add tests cases in the follow-up PR which will make use of the code in this PR.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -230,6 +274,52 @@ void PreorderAST::ConstantFold(Node *N, bool &Changed) {
   Changed = true;
 }
 
+void PreorderAST::NormalizeExprsWithoutConst(Node *N) {
+  // Normalize expressions by adding an integer constant to expressions which
+  // do not have one. For example:
+  // p ==> (p + 0)
```
kkjeer, line 280: How does `p` become normalized to `p + 0`?

mgrang: I have added more comments inline explaining why we need to normalize `p` to `p + 0`.

kkjeer (approved):

mgrang:
@dtarditi I have addressed your comments. Could you please review this?

clang/lib/AST/PreorderAST.cpp
```diff
@@ -29,24 +29,30 @@ void PreorderAST::AddNode(Node *N, Node *Parent) {
...
+      P->Children.erase(I);
       break;
     }
   }
 
-  // Move all children of the current node to its parent.
-  for (auto *Child : B->Children) {
-    Child->Parent = Parent;
-    Parent->Children.push_back(Child);
+  if (auto *B = dyn_cast<BinaryNode>(N)) {
```
dtarditi, line 46: This is only correct if (1) the child is singleton node (2) the parent and the current node are associative and commutative operators, and are the same operator.  It would be good to have an assert to prevent misuse of this function.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -56,11 +62,58 @@ void PreorderAST::Create(Expr *E, Node *Parent) {
   E = Lex.IgnoreValuePreservingOperations(Ctx, E->IgnoreParens());
 
   if (const auto *BO = dyn_cast<BinaryOperator>(E)) {
-    auto *N = new BinaryNode(BO->getOpcode(), Parent);
+    BinaryOperator::Opcode BinOp = BO->getOpcode();
+    Expr *LHS = BO->getLHS();
+    Expr *RHS = BO->getRHS();
+
+    // We can convert (e1 - e2) to (e1 + -e2) if -e2 does not overflow.  One
```
dtarditi, line 69: I think you should only put this into a normal form if -e2 is guaranteed to not overflow.   That means that if e2 is not an integer constant, it should not be put in a normal form.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -56,11 +62,58 @@ void PreorderAST::Create(Expr *E, Node *Parent) {
...
+        for (auto &PD : Diag) {
+          if (PD.second.getDiagID() == diag::note_constexpr_overflow) {
+            Overflow = true;
+            break;
+          }
+        }
+      }
+
+      if (Overflow)
+        RHS = dyn_cast<UnaryOperator>(RHS)->getSubExpr();
```
dtarditi, line 99: This is more complex than needed.  How about saving the original value of RHS?

clang/lib/AST/PreorderAST.cpp
```diff
@@ -56,11 +62,58 @@ void PreorderAST::Create(Expr *E, Node *Parent) {
...
-    Create(BO->getLHS(), /*Parent*/ N);
-    Create(BO->getRHS(), /*Parent*/ N);
+    Create(LHS, /*Parent*/ N);
+    Create(RHS, /*Parent*/ N);
+
+  } else if (!Parent) {
+    // We will enter here for expressions like "p" which are not BinaryNode and
+    // which do not yet have a parent. So we create a parent BinaryNode with
+    // addition as the operator and add "p" as a LeafExprNode of this node.
+    auto *N = new BinaryNode(BO_Add, Parent);
```
dtarditi, line 114: Why is this invariant necessary?  The invariant seems to be that the root node must be a BinaryNode.  I'm not following why it is needed.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -115,9 +168,23 @@ void PreorderAST::Sort(Node *N) {
   [&](const Node *N1, const Node *N2) {
 
     if (const auto *L1 = dyn_cast<LeafExprNode>(N1)) {
-      if (const auto *L2 = dyn_cast<LeafExprNode>(N2))
+      if (const auto *L2 = dyn_cast<LeafExprNode>(N2)) {
```
dtarditi, line 171: This comparison function seems important enough that it should be a named function with comments describing what it does, not just a lambda function.

dtarditi (commented):
I have a number of suggestions and some comments.

clang/lib/AST/PreorderAST.cpp
mgrang: I was trying to avoid leaking memory here. If we save the original value of RHS which is an Expr we would not be able to reclaim memory because the destructor for Expr is protected.

clang/lib/AST/PreorderAST.cpp
mgrang: I am not sure I understand your comment. Basically, if N is a BinaryNode we want to move its children to the parent of N and then delete N. If not, then we can simply delete N.

clang/lib/AST/PreorderAST.cpp
mgrang: Yes, the invariant is that the root node must be a BinaryNode. So for expressions like `if (*p)`, we don't have a BinaryOperator. So when we enter the Create function there is no root and the parent is null. So in this condition we create a new BinaryNode with + as the operator and add p as a child of this BinaryNode. Later in NormalizeExprsWithoutConst we normalize p to p+0 by adding 0 as a sibling of p.

clang/lib/AST/PreorderAST.cpp
mgrang: I have moved the comparison code to a comparator function but I cannot move it outside of the Sort function because I need the lambda to capture the context to refer to the Ctx and Lex objects in the comparator.

mgrang:
@dtarditi I have addressed your latest comments in my updated PR.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -56,11 +83,65 @@ void PreorderAST::Create(Expr *E, Node *Parent) {
...
+    Expr *RHS = BO->getRHS();
+
+    // We can convert (e1 - e2) to (e1 + -e2) if -e2 does not overflow.  One
+    // instance where -e2 can overflow is if e2 is INT_MIN. Here, instead of
+    // specifically checking whether e2 is INT_MIN, we add a unary minus to e2
+    // and then check if the resultant expression -e2 overflows. If it
+    // overflows, we undo the unary minus operator.
+
+    // TODO: Currently we only handle detection of overflow for integer
+    // constant expressions. We need to handle the case where e2 is a variable
```
dtarditi, line 97: I think the last sentence in this comment could be clarified.  It is kind of confusing.   We're unlikely to be able to prove that a non-constant expression overflows.  However, we could prove it does not overflow.   So I'd like us to state that we can only prove the integer constant expressions don't overflow.  We still need to handle proving that non-constant expressions do no overflow.

dtarditi (approved):
LGTM.  Thanks!  I have a suggestion for improving wording on a comment.

PR #923, opened 2020-10-08
Miscellaneous fixes to silence build warnings

mgrang:
- Removed unused code
- Added a missing `break` in a `switch case`
- Changed type of RHS variable used in comparison to match the LHS type


kkjeer (approved):
LGTM

PR #924, opened 2020-10-20
Conditional operators: infer rvalue bounds [2/n]

kkjeer:
This PR infers rvalue bounds for conditional operators `e1 ? e2 : e3` as the greatest lower bound of the rvalue bounds of `e2` and `e3`:

* If the bounds of `e2` and `e3` are equivalent, the bounds of the conditional operator are the bounds of `e2`.
* If the bounds of `e2` are `bounds(any)`, the bounds of the conditional operator are the bounds of `e3`.
* If the bounds of `e3` are `bounds(any)`, the bounds of the conditional operator are the bounds of `e2`.
* Otherwise, the bounds of the conditional operator are `bounds(unknown)`.

#### Testing:
* Added a new test to verify the inferred bounds for conditional operators.
* Added expected warnings and errors for conditional operator bounds.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -3487,20 +3487,16 @@ namespace {
       Check(E->getCond(), CSS, State);
 
       // Check the "true" arm `e2`.
-      // TODO: save the rvalue bounds from checking `e2`.  These bounds will
-      // be used to determine the rvalue bounds of `e`.
       CheckingState StateTrueArm;
       StateTrueArm.EquivExprs = State.EquivExprs;
       StateTrueArm.ObservedBounds = State.ObservedBounds;
-      Check(E->getTrueExpr(), CSS, StateTrueArm);
+      BoundsExpr *BoundsTrueArm = Check(E->getTrueExpr(), CSS, StateTrueArm);
 
       // Check the "false" arm `e3`.
-      // TODO: save the rvalue bounds from checking `e3`.  These bounds will
-      // be used to determine the rvalue bounds of `e`.
       CheckingState StateFalseArm;
       StateFalseArm.EquivExprs = State.EquivExprs;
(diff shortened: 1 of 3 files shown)


dtarditi (approved):
Looks good, thanks!

mgrang (approved):
LGTM.

kkjeer:
Some notes from the PR review discussion:

1. Why are `bounds(p, p + 4)` not returned as the greatest lower bound of `bounds(p, p + 4)` and `bounds(q, q + 5)`?
A: The current implementation does not order bounds expressions beyond the following: `bounds(unknown) < B < bounds(any)`, where `B` is a non-unknown, non-any bounds expression. There are complications involved in trying to order bounds expressions, especially if at least one bounds expression is a variable-sized range. Not ordering these bounds expressions avoids the need to split the bounds into a base and offset in order to convert to a range.

2. What is the `CheckingState` data structure and what is it used for?
A: `CheckingState` is a class that holds several members that are used to maintain an internal state while recursively checking an expression. It is commented in [SemaBounds.cpp](https://github.com/microsoft/checkedc-clang/blob/master/clang/lib/Sema/SemaBounds.cpp#L661).

3. If a pointer `p` has inferred bounds of `any` (e.g. from an assignment `p = 0`), can `p` be used to access memory? What is the dynamic bounds check here?
A: The bounds used for the dynamic bounds check are the lvalue bounds of `p`. E.g. if `p` has declared bounds `bounds(p, p + 5)`, the runtime check for `p[i]` will check that `0 <= i < 5`.

4. If a pointer `p` with declared bounds `bounds(p, p + 3)` has inferred bounds `bounds(p, p + 4)`, do future assignments to `p` have to imply `bounds(p, p + 4)`?
A: The inferred bounds for an expression only last for one top-level clang CFG statement. The target bounds for a variable are always the programmer-declared bounds (never the inferred bounds). In this example, all assignments to `p` should imply the declared bounds of `bounds(p, p + 3)`.

PR #926, opened 2020-10-22
Reset killed bounds for in-scope variables only

### LINKED ISSUE #925: ResetKilledBounds adds out-of-scope variables to bounds context
The `ResetKilledBounds` method currently resets the bounds of *all* variables whose bounds are killed by a statement `S` to their declared bounds, even if the variable is not in scope at `S`.

Consider:
```
void f(void) {
  int i, j;
  for (;;) {
    i = 1; // this statement kills the widened bounds of p and adds p => bounds(p, p + (i + j)) to the bounds context
    j = 2; // error: bounds for 'p' are unknown after statement
    _Nt_array_ptr<char> p : count(i + j) = 0;
    if (p[i + j]) { // widen the bounds of p
      return;
    }
  }
}
```

The statement `i = 1` kills the widened bounds of `p`, but `p` should not be added to the observed bounds context since it is not in scope at `i = 1`. Adding `p` to the observed bounds context results in an error at `j = 2` since the value of the variable `j` is lost.

kkjeer:
Fixes #925 

This PR adjusts ResetKilledBounds to only reset a variable `v` whose widened bounds have been killed by statement `S` if `v` is in scope at `S`, i.e. if `v` is already present in `State.ObservedBounds`.

#### Testing:
* Added a function to widened-bounds.c to test this change.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -2266,10 +2266,13 @@ namespace {
 
      // KilledBounds stores a mapping of statements to all variables whose
      // bounds are killed by each statement. Here we reset the bounds of all
-     // variables killed by the statement S to the normalized declared bounds.
+     // variables that are in scope at the statement S and whose bounds are
+     // killed by S to the normalized declared bounds.
      for (const VarDecl *V : I->second) {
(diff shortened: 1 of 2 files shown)


mgrang (approved):
LGTM.

PR #928, opened 2020-10-23
Minor improvements to the bounds widening doc

mgrang:


sulekhark (commented):
The changes look very good. However, as I understand from David's clarification today, it would be more clear to modify the sentence on lines 61-62 as:
"Given `_Nt_array_ptr V` with declared bounds `(V + low, V + high)`, the bounds
of `V` can be widened by 1 if `V` is dereferenced at its current upper bound, **and if the dereferenced value can be inferred to be non-null.**"

mgrang:
> The changes look very good. However, as I understand from David's clarification today, it would be more clear to modify the sentence on lines 61-62 as:
> "Given `_Nt_array_ptr V` with declared bounds `(V + low, V + high)`, the bounds
> of `V` can be widened by 1 if `V` is dereferenced at its current upper bound, **and if the dereferenced value can be inferred to be non-null.**"

Thanks Sulekha. For the dataflow analysis the actual value at the upper bound does not matter and we don't explicitly check that value as part of the analysis. The actual value is determined at runtime and accordingly the `if` condition would be either true (which means we would widen) or could be false (in which case we would not widen). So I think mentioning that the value at upper bound should be non-null could confuse the reader.

OTOH, I will add go ahead and add the above clarification in the doc.

dtarditi (approved):

kkjeer (approved):

PR #929, opened 2020-10-29
Compare Expressions: Extract offset from dereference expression [8/n]

mgrang:
Extract the offset by which a pointer is dereferenced. For the pointer we
compare the dereference expr with the declared upper bound expr. If the
non-integer parts of the two exprs are not equal we say that a valid offset
does not exist and return false. If the non-integer parts of the two exprs are
equal we require that the integer part of the deref expr be numerically greater
than or equal to the integer part of the upper bound expr. If this condition is
not satisfied we return false. If the condition is satisfied we return true and
the offset in this case is calculated as:
`(integer part of deref expr - integer part of upper bound expr)`.

For example, given upper bound expr `(p + i)` and deref expr `*(p + i + 1)` we
return `1` as the offset.


clang/include/clang/AST/PreorderAST.h
```diff
@@ -122,6 +122,16 @@ namespace clang {
     // @param[in] N is current node of the AST. Initial value is Root.
     void NormalizeExprsWithoutConst(Node *N);
 
+    // Get the deref offset from the DerefExpr. The offset represents the
+    // possible amount by which the bounds of an ntptr should be widened.
+    // @param[in] UpperExpr is the upper bounds expr for the ntptr.
+    // @param[in] DerefExpr is the dereferenced expr for the ntptr.
+    // @param[out] Offset is the offset from the base by which the pointer is
+    // dereferenced.
+    // return Returns a boolean indicating whether a valid offset exists.
```
dtarditi, line 131: Could you clarify the meaning of true/false here?  Sometimes clang likes to reverse the meaning of booleans returned by functions, so it is helpful to be specific.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -377,6 +377,127 @@ void PreorderAST::NormalizeExprsWithoutConst(Node *N) {
...
+    // (p + i) and (p + i + j)
+    if (B1->Children.size() != B2->Children.size())
+      return false;
+
+    // Match each child of the two nodes.
+    for (size_t I = 0; I != B1->Children.size(); ++I) {
+      auto *Child1 = B1->Children[I];
+      auto *Child2 = B2->Children[I];
+
+      if (!GetDerefOffset(Child1, Child2, Offset))
```
dtarditi, line 420: I'm not following why a recursive function call is needed here.   Shouldn't the children of the two nodes be identical, except for the children that are integer offsets?

clang/lib/AST/PreorderAST.cpp
```diff
@@ -377,6 +377,127 @@ void PreorderAST::NormalizeExprsWithoutConst(Node *N) {
...
+    // the DerefNode is not, or vice versa.
+
+    // Return false if the types of the nodes mismatch.
+    if (!B2)
+      return false;
+
+    // Return false if the opcodes mismatch.
+    // This means the two exprs are like:
+    // (p + i) and (p * i)
+    if (B1->Opc != B2->Opc)
```
dtarditi, line 406: Should we require the opcode to be addition?  This is supposed to be an offset.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -377,6 +377,127 @@ void PreorderAST::NormalizeExprsWithoutConst(Node *N) {
   AddNode(L, B);
 }
 
+bool PreorderAST::GetDerefOffset(Node *UpperNode, Node *DerefNode,
```
dtarditi, line 380: It seems to me that this function could be simplified.  You could have one function that checks whether two leaf nodes are integer constant expressions and computes their difference if any.   This function could just walk children looking for two integer constant expressions.

I thought one of the points of normalization is to use associativity and commutivity to group nodes together, i.e. `a + (b + c)` and `(a + b) + c` are put into normal forms that match.   Another point is that it groups constant expressions together so that they can be constant-folded into a single expression.  This should then simplify this code.

dtarditi (commented):
I think this code could be made simpler by relying on the invariants from normalizing expressions.   I'm not following a recursive call that is being made or why the function isn't checking for addition.  I also think you should add more test cases, including negative test cases that aren't expected to result in an offset being found.

clang/lib/AST/PreorderAST.cpp
mgrang: Yes, this is true if we only had the addition operator. Then the tree would be flattened with a single BinaryNode (+) and all children sorted and the integers constant-folded. Then we could simply walk the tree matching expressions and compute the difference only for the constant expressions.

The reason I had to recursively match was to support multiplication. Consider two expressions like `a + (b * 2) + 1` and `a + (b * 2) + 2`. In this case we cannot coalesce the two BinaryNodes and the two trees are:
```
+                              +
 \                              \
  a, 1, *                        a, 2, *
          \                              \
           b, 2                           b, 2
```

With that said, I think I need to slightly change my logic to compute the difference only for integer constant expressions which are children of the addition operator and ensure all children of a multiplication operator are equal.

mgrang:
There are a lot of positive and negative tests for bounds widening in
https://github.com/microsoft/checkedc-clang/blob/master/clang/test/CheckedC/inferred-bounds/widened-bounds.c and
https://github.com/microsoft/checkedc-clang/blob/master/clang/test/CheckedC/inferred-bounds/widened-bounds-semantic-compare.c

mgrang:
Here are a couple of issues where some cases are failing. I will fix them as follow-up to this PR.
https://github.com/microsoft/checkedc-clang/issues/932
https://github.com/microsoft/checkedc-clang/issues/933

mgrang:
Some questions from the PR review meeting:

1. What happens with casts which are present in the clang AST? Do we consider those during normalization and comparison?
A: The preorder AST is a stripped version of the clang AST and we control the building of the preorder AST. So while creating the preorder AST we skip casts and simply create nodes for the variable and constant expressions. So, no we do not consider casts during comparison.

2. Are expressions like `if (!*p)` and `if (*p != '\0')` handled?
A: These and several other variations are TODO.

3. What happens to bounds widening if the case of `if (*(p - 1))`?
A: In this case we require the declared upper bounds are also `(p - 1)`. If they are then bounds widening would happen. Although, whether or not we actually enter that if condition is a question which could only be answered at run time.

kkjeer (approved):

dtarditi (approved):
Looks good, thank you!

PR #940, opened 2020-11-16
Detect indirect relationships between variables

kkjeer:
This PR modifies the definition of free variables so that, if:
* `x` is a variable appearing in declared (or inferred) bounds, and:
* `y` is a variable appearing in inferred (or declared) bounds, and:
* `e1` is an expression that uses the value of `x`, and:
* `e2` is an expression that uses the value of `y`, and:
* `EquivExprs` contains a set that contains both `e1` and `e2`

then `x` is *not* a free variable.

This change means that some bounds warnings that were converted to free variable errors are now back to being warnings. The free variable note messages are also reworded to mention expressions in declared/inferred bounds rather than variables.

#### Testing
* Modified tests to reflect free variable errors which are now back to being bounds warnings.
* Modified tests to reflect reworded free variable note messages.
* Modified checkedc tests in [checkedc/428](https://github.com/microsoft/checkedc/pull/428).

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -1432,6 +1432,49 @@ namespace {
         return ProofResult::Maybe;
       }
 
+      // CheckFreeVarInExprs returns true if there are any free variables in
+      // E1 w.r.t E2, or if there are any free variables in E2 w.r.t E1.
+      // Any free variables in E1 with position Pos1 and any free variables in
+      // E2 with position Pos2 are appended to FreeVars.
+      bool CheckFreeVarInExprs(Expr *E1, Expr *E2,
+                               FreeVariablePosition Pos1,
+                               FreeVariablePosition Pos2,
+                               EquivExprSets *EquivExprs,
+                               FreeVariableListTy &FreeVars) {
+        // If E1 or E2 accesses memory via pointer, we skip because we cannot
+        // determine aliases for two indirect accesses soundly yet.
+        if (ReadsMemoryViaPointer(E1) || ReadsMemoryViaPointer(E2))
+          return false;
+
+        bool HasFreeVariables = false;
+        EqualExprTy Vars1 = CollectVariableSet(S, E1);
+        EqualExprTy Vars2 = CollectVariableSet(S, E2);
+
+        if (AddFreeVariables(Vars1, Vars2, EquivExprs, Pos1, FreeVars))
+          HasFreeVariables = true;
+
+        if (AddFreeVariables(Vars2, Vars1, EquivExprs, Pos2, FreeVars))
+          HasFreeVariables = true;
+
+        return HasFreeVariables;
+      }
+
+      // AddFreeVariables creates a pair <Variable, Pos> for each free variable
+      // in SrcVars w.r.t. DstVars and appends the pair to FreeVariablesWithPos.
+      bool AddFreeVariables(const EqualExprTy &SrcVars,
+                            const EqualExprTy &DstVars,
+                            EquivExprSets *EquivExprs,
+                            FreeVariablePosition Pos,
+                            FreeVariableListTy &FreeVariablesWithPos) {
+        EqualExprTy FreeVariables;
+        if (GetFreeVariables(SrcVars, DstVars, EquivExprs, FreeVariables)) {
+          for (const auto V : FreeVariables)
+            FreeVariablesWithPos.push_back(std::make_pair(V, Pos));
+          return true;
+        }
+        return false;
+      }
+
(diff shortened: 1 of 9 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1453,23 +1453,78 @@ namespace {
 
         // Gather free variables.
         for (const auto &SrcV : SrcVars) {
-          if (IsEqualToConstant(S, SrcV, EquivExprs))
+          DeclRefExpr *SrcVar = cast<DeclRefExpr>(SrcV);
```
mgrang, line 1456: What happens if `SrcV` is not a `DeclRefExpr`?

kkjeer: The list `SrcVars` should always be a list of `DeclRefExprs` since it is obtained by searching for all the `DeclRefExprs` that occur in a bounds expression. As a separate cleanup item, the type of `SrcVars` should be changed from `EqualExprTy` (set of `Exprs`) to a set of `DeclRefExprs`.

kkjeer: Created #941 to track this.

kkjeer:
Some notes from the PR discussion:

* A potential performance improvement is to construct a map of each expression in `EquivExprs` to the set of variables the expression uses. This prevents the need to repeatedly search each expression in `EquivExprs` for expressions that use `SrcV` or a variable in `DstVars`. This is a follow-up item to consider if further testing data indicates that performance improvements are necessary.
* Which kinds of expressions can be used to find indirect relationships between variables?
The `EquivExprs` set (containing expressions that can be used to find variable relationships) can contain expressions that:
1. Are non-modifying (function calls, assignments, etc. **cannot** be in `EquivExprs`)
2. Do not read memory via a pointer (pointer dereferences, array subscripts, etc. **cannot** be in `EquivExprs`)
3. Create a new object (string literals, initializer lists, etc. **cannot** be in `EquivExprs`)
Some examples of expressions that can appear in `EquivExprs` are `x`, `3`, `y + 1`, `a && b`, `&p`, etc.
* How is `EquivExprs` modified?
`EquivExprs` is modified after assignments (`x = y`, `x++`, `x += z`, etc.) and initializers. For example, after checking the assignment `x = y`, `EquivExprs` will contain a set containing `x` and `y` (thus recording equality between `x` and `y`).
* Which kinds of expressions do we use to emit free variable errors?
We only consider variables, i.e. `DeclRefExprs` to be free. For example, in declared bounds `(p, p + s.len)` and inferred bounds `(p, p + 3)`, we do not consider `s.len` to be a "free expression". In declared bounds `(p, p + x)` and inferred bounds `(p, p + 3)`, we can consider `x` to be a free variable.

mgrang (approved):
LGTM.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1432,111 +1432,170 @@ namespace {
...
+      // sets of Expr EquivExprs. A variable V in SrcVars is *free* if these
       // conditions are met:
-      //   1. either (1) V is not an integer type, or (2) there is no set in
-      //   EquivExprs that contains both V and a constant (i.e. an integer-typed
-      //   value); and
-      //   2. for each variable U in DstVars, there is not set in EquivExprs
-      //   that contains both V and U.
+      //   1. V is not equal to an integer constant, i.e. there is no set in
+      //      EquivExprs that contains V and an IntegerLiteral expression, and:
+      //   2. For each variable U in DstVars, V is not equivlaent to U, i.e.
```
dtarditi, line 1485: This is a typo in the spelling of equivalent.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1432,111 +1432,170 @@ namespace {
         return ProofResult::Maybe;
       }
 
+      // CheckFreeVarInExprs returns true if there are any free variables in
+      // E1 w.r.t E2, or if there are any free variables in E2 w.r.t E1.
+      // Any free variables in E1 with position Pos1 and any free variables in
```
dtarditi, line 1437: The comment is somewhat confusing in its explanation of the uses of the Pos1 and Pos2 arguments.  I think it would better to expand the comment:  Any free variables in E1 and any free variables in E2 are appended to FreeVars.  Pos1 is the position of E1 and Pos2 is the position of E2.  Free variables are appended with the position information of the expression in which they are free (free variables in E1 are appended with the position of E1, for example).
```

dtarditi (approved):
Looks great, thanks!  I have some suggestions about comments.

Issue #932, opened 2020-11-04
Fix assert in bounds widening "BinaryNode operator must equal parent operator"


mgrang:
Handle the following case in bounds widening:

```
void f(int i) {
  _Nt_array_ptr<char> p : count(0) = "a";

  if (*(p + (0 * 1)))
  {}
}
```
Currently, we hit an assert:
```
clang-9: /usr/local/magrang/master/src/clang/lib/AST/PreorderAST.cpp:46: void clang::PreorderAST::RemoveNode(clang::Node*, clang::Node*): Assertion `B->Opc == P->Opc && "BinaryNode operator must equal parent operator"' failed.
```

This is because we have incorrectly added the requirement that in order to remove a BinaryNode its opcode should be equal to its parent opcode which is not valid in this case.

mgrang:
Fixed in https://github.com/microsoft/checkedc-clang/pull/943

PR #943, opened 2020-11-17
Compare Expressions: Fix the conditions for coalescing a node with its parent

mgrang:
We can coalesce a BinaryNode if any one of the following conditions are
true:
1. The parent has the same operator as the current node OR
2. The current node has just one child (for example, as a result of
constant folding) and the parent and current operators are commutative and
associative.

Fixes issue #932

### DIFF
--- a/clang/lib/AST/PreorderAST.cpp
+++ b/clang/lib/AST/PreorderAST.cpp
@@ -29,9 +29,12 @@ void PreorderAST::AddNode(Node *N, Node *Parent) {
   }
 }
 
-bool PreorderAST::CanCoalesceNode(BinaryNode *B, BinaryNode *P) {
-  if (!B || !isa<BinaryNode>(B) ||
-      !P || !isa<BinaryNode>(P))
+bool PreorderAST::CanCoalesceNode(BinaryNode *B) {
+  if (!B || !isa<BinaryNode>(B))
+    return false;
+
+  auto *P = dyn_cast<BinaryNode>(B->Parent);
+  if (!P)
     return false;
 
   // We can coalesce a BinaryNode if any one of the following conditions are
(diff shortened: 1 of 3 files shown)


mgrang:
> The comment here needs to be updated to be in sync with the fix.

Done.

mgrang:
Some notes from the PR review discussion:

1. Maybe we should remove the operator as part of constant folding. For example, given nodes `+,1,2,3` we should not replace them with `+,6` but `6`. This means we need to remove the operator as part of constant folding itself. We also need to handle the case where an operator has constant and non-constant nodes. For example, given `+,a,2,3` we should not get rid of the operator in this case.

2. The tree is not really a "binary" tree but an n-ary tree. So calling the operator nodes as `BinaryNode` sounds confusing. Maybe we should rename `BinaryNode` to `InnerNode`.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -29,51 +29,48 @@ void PreorderAST::AddNode(Node *N, Node *Parent) {
...
+  if (!B || !isa<BinaryNode>(B) ||
+      !P || !isa<BinaryNode>(P))
+    return false;
+
+  // We can coalesce a BinaryNode if any one of the following conditions are
+  // true:
+  // 1. The parent has the same operator as the current node OR
+  // 2. The current node has just one child (for example, as a result of
+  // constant folding) and the parent and current operators are commutative and
+  // associative.
```
sulekhark, line 45: My thoughts are as follows:
a) Should the condition 1 above not be "parent has the same operator as the current node, and the operator is commutative and associative" ?
b) If we implement constant folding to be complete in itself, then the current node can never have just one child because of constant folding. It is possible that the current node will have just one child if the operator at the current node is a unary operator.

mgrang: > My thoughts are as follows:
> a) Should the condition 1 above not be "parent has the same operator as the current node, and the operator is commutative and associative" ?
Could you elaborate why this should not be the case? We should not try to coalesce different operators and coalescing only makes sense if the operator is commutative and associative because after coalescing we also need to sort the nodes.

> b) If we implement constant folding to be complete in itself, then the current node can never have just one child because of constant folding. It is possible that the current node will have just one child if the operator at the current node is a unary operator.
We have now implemented constant folding to be complete in itself. As part of constant folding we now invoke this method to coalesce the node. So we need this condition here. Otherwise we would end up duplicating logic to coalesce nodes.

sulekhark (approved):
LGTM.

kkjeer (approved):

Issue #933, opened 2020-11-04
Correctly widen bounds for exprs not containing addition or multiplication


mgrang:
Handle the following case in bounds widening:
```
void f(int i) {
  _Nt_array_ptr<char> p : bounds(p, p - i + 0) = "a";

  if (*(p - i))
  {}
}
```

We call the function `NormalizeExprsWithoutConst` which adds 0 constant to expressions not containing an integer constant. But the problem is we only add `+ 0` to an expression which contains addition and `* 1` to an expression which contains multiplication. In the above case, we do not have addition or multiplication in the deref expression. This results in the bounds not getting widened.

The proper fix for this is to always add `+ 0` to every expression without checking the operator. This 0 constant would anyway be constant folded. So it should not be a problem.

mgrang:
Fixed in https://github.com/microsoft/checkedc-clang/pull/944

PR #944, opened 2020-11-17
Compare Expressions: Fix normalization of exprs by adding zero

mgrang:
Previously, in order to normalize exprs we added zero only to exprs not
containing any other constants. Also, for binary nodes with + operator we added
+0 and for nodes with * operator we added *1. This resulted in exprs like "p -
i" which do not contain constants but also do not have + or * operator to not
get normalized. So we simply the logic to add zero. Now we simply add a binary
node with the + operator and a 0 child at the root of the AST. This makes the
logic simpler and handles a wider variety of cases.

This fixes issue #933

### DIFF
--- a/clang/include/clang/AST/PreorderAST.h
+++ b/clang/include/clang/AST/PreorderAST.h
@@ -118,10 +118,6 @@ namespace clang {
     // this to control when to stop recursive constant folding.
     void ConstantFold(Node *N, bool &Changed);
 
-    // Normalize expressions which do not have any integer constants.
-    // @param[in] N is current node of the AST. Initial value is Root.
-    void NormalizeExprsWithoutConst(Node *N);
-
     // Get the deref offset from the DerefExpr. The offset represents the
     // possible amount by which the bounds of an ntptr could be widened.
     // @param[in] UpperExpr is the upper bounds expr for the ntptr.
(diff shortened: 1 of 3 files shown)


sulekhark:
Have we confirmed that constant folding will not fold `i + 0` to `i` ? Will there be some circumstances where we want `i + 0` to constant-fold to `i`, and some other circumstances where we do not want this to happen (ex: to maintain normalization)?

mgrang:
> Have we confirmed that constant folding will not fold `i + 0` to `i` ? Will there be some circumstances where we want `i + 0` to constant-fold to `i`, and some other circumstances where we do not want this to happen (ex: to maintain normalization)?

Yes, I have confirmed that `i + 0` will not constant fold to `i`. For example,
```
void f(int i) {
  _Nt_array_ptr<char> p : bounds(p, p + 0) = "a";

  if (*(p + 0))
  {}
}
```

On dumping the preorder AST we see that +0 is preserved:
```
+
ImplicitCastExpr 0x8927c50 '_Nt_array_ptr<char>' <LValueToRValue>
`-DeclRefExpr 0x8927c08 '_Nt_array_ptr<char>' lvalue Var 0x8927b68 'p' '_Nt_array_ptr<char>'
IntegerLiteral 0x8927f48 'int' 0
```

sulekhark:
LGTM.

sulekhark (approved):
LGTM.

PR #946, opened 2020-11-20
Compare Expressions: Rename BinaryNode to OperatorNode [NFC]

mgrang:
The preorder AST is an n-ary tree. Hence, calling an operator node as
BinaryNode is not correct. So we rename BinaryNode to OperatorNode.

### DIFF
--- a/clang/lib/AST/PreorderAST.cpp
+++ b/clang/lib/AST/PreorderAST.cpp
@@ -197,12 +197,12 @@ bool PreorderAST::CompareNodes(const Node *N1, const Node *N2) {
     return true;
 
   // Compare N1:OperatorNode and N2:OperatorNode.
-  const auto *B1 = dyn_cast<OperatorNode>(N1);
-  const auto *B2 = dyn_cast<OperatorNode>(N2);
+  const auto *O1 = dyn_cast<OperatorNode>(N1);
+  const auto *O2 = dyn_cast<OperatorNode>(N2);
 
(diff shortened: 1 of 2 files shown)


clang/lib/AST/PreorderAST.cpp
```diff
@@ -198,71 +188,71 @@ bool PreorderAST::CompareNodes(const Node *N1, const Node *N2) {
...
-  // Compare N1:BinaryNode and N2:BinaryNode.
-  const auto *B1 = dyn_cast<BinaryNode>(N1);
-  const auto *B2 = dyn_cast<BinaryNode>(N2);
+  // Compare N1:OperatorNode and N2:OperatorNode.
+  const auto *B1 = dyn_cast<OperatorNode>(N1);
+  const auto *B2 = dyn_cast<OperatorNode>(N2);
 
   if (B1->Opc != B2->Opc)
     return B1->Opc < B2->Opc;
   return B1->Children.size() < B2->Children.size();
```
sulekhark: Minor observation (feel free to ignore if you wish): In almost all places variable `B` has been renamed to `O` except in the few lines above (variables `B1` and `B2`). There are two other similar occurrences of `B1` and `B2`.

sulekhark (approved):
LGTM.

clang/lib/AST/PreorderAST.cpp
mgrang: I have fixed these in my latest commit.

PR #948, opened 2020-11-23
Rename EqualExprTy to ExprSetTy

### LINKED ISSUE #927: Rename EqualExprTy typedef
The name `EqualExprTy` was introduced as a typedef for the set `CheckingState.SameValue` of expressions that produce the same value. This typedef is now being used for other purposes that are not related to expression equality, so it should be renamed to something like `ExprSetTy`.

kkjeer:
Fixes #927 

This PR renames the `EqualExprTy` type to `ExprSetTy`. `EqualExprTy` was originally created as the type of the `SameValue` set in `CheckingState` that holds expressions that produce the same value as some expression `e`, but this is no longer the only use of the type.


sulekhark (approved):
LGTM.

mgrang (approved):
LGTM.

PR #949, opened 2020-11-30
Where Clause: Parse a where clause that redeclares bounds [1/n]

mgrang:
Added logic to parse a where clause that redeclares the bounds of a pointer.
For example,

```
_Nt_array_ptr<char> p : count(0);
int x = strlen(p) _Where p : bounds(p, p + x);
```

For the pointer we add a new annotation called `WClause` in the AST. `WClause` is
an instance of class `WhereClause` and holds all facts declared for a pointer
using a where clause.

In this PR:
1. We only parse a where clause declared after a function call, like `strlen` in
the above example.
2. We only parse facts which redeclare bounds for a checked pointer. For
example, we do not parse facts like `where x == e1`.
3. We only parse a single fact declared using a where clause. For example, we
do not parse multiple facts like `where x > e1 && x < e2`.


clang/test/CheckedC/parsing/where-clauses.c
```diff
@@ -0,0 +1,15 @@
...
+
+void f1(_Nt_array_ptr<char> p : count(0)) {
+  int a = func(p) _Where p : bounds(p, p + a);
+  int b = func(p) _Where p : count(b);
+  int c = func(p) _Where p;
+  int d = func(p) _Where; // expected-error {{expected expression}}
+  int e = func(p) _Where p :; // expected-error {{expected bounds expression}}
+  int f = func(p) _Where p bounds(); // expected-error {{expected ';' at end of declaration}}
+  int g = func(p) _Where p : bounds(); // expected-error {{expected expression}}
+}
```
sulekhark, line 15: Would it be possible to add test cases like:
1) unsigned long a;
    a = strlen(p)  _Where .....

2) long a;
    for (int i = 0, a = strlen(s) _Where  ... ; .... ; ....)

sulekhark: One more test case:
3. int a;
     a = func(p) _Where p : bounds(q, q + a);

mgrang:
David and I discussed the design for parsing where clauses. We decided to parse where clauses as follows:
1. For declarations: Parse where clause when parsing the declaration. In this case the where clause would be attached to the `Decl`.
2. For all others: Parse where clause when parsing the ExprStmt. In this case the where clause would be attached to the `ExprStmt`.

The checking of the function `strlen` would be handled inside `Sema`. The individual where clause facts can be of type `Expr`.

PR #950, opened 2020-12-02
Remove.gitattributes file

kkjeer:
The .gitattributes file causes some issues on Windows when attempting to upgrade to the latest clang/llvm sources. The cache fails to refresh for a random selection of files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-* text eol=lf


dtarditi (approved):
Looks great.  Please merge this as soon as you can, as this has caused issues for other people too: see [https://github.com/correctcomputation/checkedc-clang/issues/317](https://github.com/correctcomputation/checkedc-clang/issues/317)

PR #951, opened 2020-12-04
Fix x64 Windows builds break

mgrang:
Add /bigobj flag to fix x64 Windows compile error for the 3C tool.


mgrang:
With this fix there are no errors related to `/bigobj` flag although there are other unit test failures which the 3C team are currently fixing. I will go ahead and merge this PR.

Issue #947, opened 2020-11-23
Upgrade to LLVM 11


mattmccutchen-cci:
I believe @kkjeer is in the process of upgrading this repository to LLVM 11.  I'm creating this issue to have a public place to talk about the process that is more appropriate than https://github.com/microsoft/checkedc-clang/pull/930#issuecomment-726152033.

What's the latest plan for how the git history for the merge will be constructed?

Here's what I know so far about the options:

When I first spotted the creation of the `checkedc-clang-110-rc` branch on November 12, it appeared that the plan was to squash all previous Checked C and 3C development into one or a few commits on top of LLVM 11.  We (CCI) objected that this would make it a pain to trace the origin of Checked C or 3C code across the upgrade, and since it is a non-fast-forward change, it would disrupt dependent repositories such as our 5C repository as well.  @mgrang then agreed in an email to @mwhicks1 not to do it this way.

Then yesterday, I saw the `clang-11-upgrade` branch, which seems to be doing it the other way: squashing the difference between our current LLVM 9 baseline and LLVM 11 onto the current Checked C history.  (I took a look and it appears that that the diff that was squashed onto `clang-11-upgrade` may be incorrect, but there may be plans to fix that.)  This approach has an analogous problem of making it a pain to trace the origin of LLVM code across the upgrade, but I don't know if we ever do that, so we likely don't care.  And it is a fast-forward change, so it won't disrupt 5C.  So I think CCI is OK with it.

[...]

PR #955, opened 2020-12-08
Update llvm source upgrade doc

kkjeer:
* Update the url for the llvm project to https://github.com/llvm/llvm-project
* Add a step to the baseline branch process for resetting to the desired llvm/clang source commit
* Add a step to the master branch process for setting the local git merge rename limit

### DIFF
--- a/clang/docs/checkedc/Update-to-latest-LLVM-sources.md
+++ b/clang/docs/checkedc/Update-to-latest-LLVM-sources.md
@@ -23,13 +23,20 @@ The third step is to merge your changes back into your baseline and master branc
 First create remote to the mirrored GitHub repo that contains the updated sources
 for LLVM/Clang. Go to your LLVM/Clang repo and do:
 
-    git remote add mirror https://github.com/llvm-mirror/llvm-project
+    git remote add mirror https://github.com/llvm/llvm-project
 
 Then branch your baseline branch and merge changes into it:
 
     git checkout baseline
     git checkout -b updated-baseline
     git pull mirror master
+    git reset --hard <llvm-commit>
+
+`<llvm-commit>` is the commit on https://github.com/llvm/llvm-project that contains the desired version of the LLVM/Clang sources. For example, to update to LLVM/Clang version 11.0:
+
+    git reset --hard 176249bd6732a8044d457092ed932768724a6f06
+
+To help ensure stability, `<llvm-commit>` should be a commit on an `llvmorg-<version>` tag. These tags are final LLVM/Clang releases (see https://llvm.org/docs/HowToReleaseLLVM.html#tag-the-llvm-final-release). For example, the LLVM/Clang 11.0 commit is the [latest commit](176249bd6732a8044d457092ed932768724a6f06) on the [llvmorg-11.0.0](https://github.com/llvm/llvm-project/tree/llvmorg-11.0.0) tag.
 
 ## Run testing on your branched baseline branches.
 
@@ -46,6 +53,7 @@ that first.
 You can now branch your baseline branches to create a new master branch:
 
     git checkout -b updated-master
+    git config --local merge.renamelimit 26000
     git merge master
 
 You will very likely have merge conflicts and some test failures.  The test
--- a/clang/docs/checkedc/Update-to-latest-LLVM-sources.md
+++ b/clang/docs/checkedc/Update-to-latest-LLVM-sources.md
@@ -28,15 +28,12 @@ for LLVM/Clang. Go to your LLVM/Clang repo and do:
 Then branch your baseline branch and merge changes into it:
 
     git checkout baseline
-    git checkout -b updated-baseline
-    git pull mirror master
-    git reset --hard <llvm-commit>
+    git fetch mirror
+    git checkout -b updated-baseline <llvmorg-version>
 
-`<llvm-commit>` is the commit on https://github.com/llvm/llvm-project that contains the desired version of the LLVM/Clang sources. For example, to update to LLVM/Clang version 11.0:
+To help ensure stability, `<llvmorg-version>` is a final LLVM/Clang release tag (see https://llvm.org/docs/HowToReleaseLLVM.html#tag-the-llvm-final-release). For example, to update to LLVM/Clang version 11.0.0 sources:
 
-    git reset --hard 176249bd6732a8044d457092ed932768724a6f06
-
-To help ensure stability, `<llvm-commit>` should be a commit on an `llvmorg-<version>` tag. These tags are final LLVM/Clang releases (see https://llvm.org/docs/HowToReleaseLLVM.html#tag-the-llvm-final-release). For example, the LLVM/Clang 11.0 commit is the [latest commit](176249bd6732a8044d457092ed932768724a6f06) on the [llvmorg-11.0.0](https://github.com/llvm/llvm-project/tree/llvmorg-11.0.0) tag.
+    git checkout -b updated-baseline llvmorg-11.0.0
 
 ## Run testing on your branched baseline branches.
 


sulekhark (commented):
LGTM.

clang/docs/checkedc/Update-to-latest-LLVM-sources.md
```diff
@@ -46,6 +53,7 @@ that first.
 You can now branch your baseline branches to create a new master branch:
 
     git checkout -b updated-master
+    git config --local merge.renamelimit 26000
```
mattmccutchen-cci, line 56: Instead of hard-coding the number, can you add a command to determine the number that should be used for a given merge?  I think `git diff` has shown me a warning message with the number of potential renames in that particular diff; if you then set that as the limit, you are guaranteed that the limit will not cause any renames to be missed.

clang/docs/checkedc/Update-to-latest-LLVM-sources.md
```diff
@@ -46,6 +53,7 @@ that first.
...
     git merge master
```
mattmccutchen-cci, line 57: Next time, when you go from one LLVM release branch to another, you are probably going to have to do an [ours-merge](https://github.com/microsoft/checkedc-clang/issues/947#issue-749005150).  I imagine you don't want to describe the details here since you haven't tested them yet, but perhaps at least link to my comment so you'll see it?

mattmccutchen-cci (changes requested):
Some suggestions.  I'm marking this review "request changes" in an effort to ensure it isn't accidentally overlooked; I mean no offense and will change the status to "comment" or "approve" if you state that you don't wish to address my suggestions.

clang/docs/checkedc/Update-to-latest-LLVM-sources.md
```diff
@@ -23,13 +23,20 @@ The third step is to merge your changes back into your baseline and master branc
...
 Then branch your baseline branch and merge changes into it:
 
     git checkout baseline
     git checkout -b updated-baseline
     git pull mirror master
+    git reset --hard <llvm-commit>
+
+`<llvm-commit>` is the commit on https://github.com/llvm/llvm-project that contains the desired version of the LLVM/Clang sources. For example, to update to LLVM/Clang version 11.0:
+
+    git reset --hard 176249bd6732a8044d457092ed932768724a6f06
```
mattmccutchen-cci, line 37: You can just use the tag name here, i.e., `git reset --hard llvmorg-11.0.0`, rather than manually copying and pasting the commit ID.

clang/docs/checkedc/Update-to-latest-LLVM-sources.md
```diff
@@ -23,13 +23,20 @@ The third step is to merge your changes back into your baseline and master branc
...
 
-    git remote add mirror https://github.com/llvm-mirror/llvm-project
+    git remote add mirror https://github.com/llvm/llvm-project
 
 Then branch your baseline branch and merge changes into it:
 
     git checkout baseline
     git checkout -b updated-baseline
     git pull mirror master
+    git reset --hard <llvm-commit>
```
mattmccutchen-cci, line 33: Rather than starting the `updated-baseline` branch at `baseline` only to hard-reset it to a different commit, you can start it at the commit you want; the effect will be the same.
```
git fetch mirror
git checkout -b updated-baseline <llvm-commit>
```

mattmccutchen-cci (changes requested):

clang/docs/checkedc/Update-to-latest-LLVM-sources.md
kkjeer: Could you please clarify what you mean by an ours-merge? Is this term referring to a git recursive merge with the `ours` option? I'm not sure we would want to do a `git merge recursive ours` for this scenario.

clang/docs/checkedc/Update-to-latest-LLVM-sources.md
kkjeer: Would the `git diff` command to check the necessary rename limit need to be run after attempting to run `git merge master`? Ideally the merge rename limit should be set before attempting to merge so that the merge works the first time.

clang/docs/checkedc/Update-to-latest-LLVM-sources.md
```diff
@@ -23,13 +23,17 @@ The third step is to merge your changes back into your baseline and master branc
 First create remote to the mirrored GitHub repo that contains the updated sources
 for LLVM/Clang. Go to your LLVM/Clang repo and do:
 
-    git remote add mirror https://github.com/llvm-mirror/llvm-project
+    git remote add mirror https://github.com/llvm/llvm-project
 
 Then branch your baseline branch and merge changes into it:
 
     git checkout baseline
```
mattmccutchen-cci, line 30: To be clear, you can delete the `git checkout baseline` command now since the later `git checkout` command switches to another branch before anything is done with `baseline`.

clang/docs/checkedc/Update-to-latest-LLVM-sources.md
mattmccutchen-cci: I think there is a way to check it before the merge, but I realized my second time around that it's probably more sensible to just set a number larger than anything you could ever need.  Was 26000 supposed to be such a number?  Maybe 99999999 or so would make the point clearer?

clang/docs/checkedc/Update-to-latest-LLVM-sources.md
mattmccutchen-cci: The essence of the problem is that the basic principle of an ordinary `git merge` is "take the union of the commits of all the parents".  In a hypothetical future upgrade to LLVM 12, we'd be starting with `updated-master` containing part of the LLVM mainline plus the LLVM 12 release branch up to the tag we use, and `master` containing a shorter part of the LLVM mainline plus the LLVM 11 release branch (up to the tag we used) and the Checked-C-specific commits.  So an ordinary `git merge` will try to retain everything up to our LLVM 12 tag and the Checked-C-specific commits, _plus_ the stuff on the LLVM 11 release branch.  What we really want is to retain everything up to the LLVM 12 tag and the Checked-C-specific commits but _not_ the stuff on the LLVM 11 release branch, which is no longer relevant to the LLVM 12 version we're upgrading to.

If we're lucky and everything on the LLVM 11 release branch is an _exact_ cherry-pick of something later on the mainline, the textual merge will be convergent and the difference will not matter.  But if anything goes wrong, we may get merge conflicts, and even if we resolve any and all merge conflicts, we may be left with some changes specific to the LLVM 11 release branch, so our LLVM baseline is not really equivalent to LLVM 12, and this may cause unforeseen problems later.

The workaround I proposed is to first construct a commit which I'll call X that has the same file content as the LLVM 12 tag we want, but pretend to Git that it also incorporates the portion of the LLVM 11 release branch up to the tag we used.  (This is a lie and may be somewhat confusing when we browse the history later, but it's a well-known technique that is the least evil to handle this scenario given Git's current fundamental limitations.)  We construct X by checking out the LLVM 12 tag and then merging the LLVM 11 tag with `git merge -s ours`.  Then, when we merge X (instead of the LLVM 12 tag) with our previous master, since Git thinks that X already contains the LLVM 11 release branch stuff, it will determine that the way to union everything is to apply the diff from the LLVM 11 tag to X to our master.  Since X has the same file content as the LLVM 12 tag, this will leave us with content equivalent to the LLVM 12 tag plus the Checked-C-specific commits, which is what we want.

Does that all make sense?

dtarditi: @mattmccutchen-cci: @kkjeer and I chatted.  Our goal is to follow the clang/LLVM branching strategy.   Our main branch should be based on the clang main branch.  That means our starting commit should have been taken from the clang/LLVM main branch, not a release branch.  Ideally it should be the commit in main used for creating the release branch.  In the past, we sometimes had problems with that exact commit having issues or build breaks, so we may be off by a few commits (clang/LLVM has an optimistic commit strategy).

 If we want to have a version of the Checked C compiler with all patches for a specific release, we would create a release branch of the Checked C clang/LLVM compiler.    It would follow the same pattern that we use overall: start with the actual clang/LLVM release branch, and then merge the Checked C changes into it.  Given that the Checked C clang/LLVM branch should be based off the same commit used for the actual clang/LLVM release branch, the merge should be smooth. There is a possibility that a cherry picked fix for the clang/LLVM release branch conflicts with a Checked C specific change, in which case we'd have to resolve that marge conflict.

mattmccutchen-cci: @dtarditi Sounds OK to me; then this PR should be updated accordingly.  If we do run into stability problems with the LLVM mainline serious enough to get in the way of our development, we can always switch the baseline of the Checked C `master` to an LLVM release branch later; IMO that would be saner than having to do a test merge of a Checked C branch with the LLVM release branch every time we want to validate a Checked C change.  Or if we can identify a small number of commits on the LLVM release branch that fix the problem affecting us, we could cherry-pick them and then revert the cherry-picks at the beginning of the next baseline upgrade process.

It would be nice if you had responded [here](https://github.com/microsoft/checkedc-clang/issues/947#issuecomment-736844803) before I took the time to write that wall of text, though I know you're busy.  In any event, now you know what it would take if you did want to use successive LLVM release branches as baselines for Checked C.

kkjeer:
Closing in favor of #1045.

PR #960, opened 2020-12-22
[Where Clause] Parse a Checked C where clause on a null statement [1/n]

mgrang:
Parse a Checked C where clause attached to a null statement. Such a where clause would have the following syntax:
      `_Where fact1 _And fact2`

A fact could either be a bounds fact or a relational operator fact. Facts occurring in a where clause are separated by a new keyword `_And`.

As a follow-up to this PR I will add support for where clauses occurring with declarations and assignments.


clang/lib/Parse/ParseStmt.cpp
```diff
@@ -2427,3 +2447,47 @@ bool Parser::ParseOpenCLUnrollHintAttribute(ParsedAttributes &Attrs) {
...
+
+  // Consume the "_Where" token.
+  if (ExpectAndConsume(tok::kw__Where))
+    return nullptr;
+
+  WhereClause *WClause = Actions.ActOnWhereClause(Loc);
+
+  while (true) {
+    if (ParseWhereClauseFact(WClause))
+      return nullptr;
```
sulekhark, line 2483: If the current WhereClauseFact is invalid in some way, must we choose to discard all the valid facts parsed previously, or must we issue a warning that we discard the current and subsequent facts of the where clause while continuing further with the valid facts parsed until this point?

clang/lib/Parse/ParseStmt.cpp
```diff
@@ -2427,3 +2447,47 @@ bool Parser::ParseOpenCLUnrollHintAttribute(ParsedAttributes &Attrs) {
...
+}
+
+WhereClause *Parser::ParseWhereClause() {
+  SourceLocation Loc = Tok.getLocation();
+
+  // Consume the "_Where" token.
+  if (ExpectAndConsume(tok::kw__Where))
+    return nullptr;
+
+  WhereClause *WClause = Actions.ActOnWhereClause(Loc);
```
sulekhark, line 2479: The null check on WClause is deep inside ActOnBoundsFact and ActOnRelopFact. Maybe it could move here?

clang/lib/Sema/SemaStmt.cpp
```diff
@@ -4418,3 +4418,52 @@ StmtResult Sema::ActOnCapturedRegionEnd(Stmt *S) {
...
+  LookupResult Lookup(*this, VarName, Loc, Sema::LookupOrdinaryName);
+  LookupParsedName(Lookup, CurScope, nullptr, true);
+  if (Lookup.empty())
+    return IsError;
+
+  VarDecl *VD = Lookup.getAsSingle<VarDecl>();
+  if (!VD)
+    return IsError;
+
+  BoundsFact *Fact = new (Context) BoundsFact(VD, Bounds, Loc);
```
sulekhark, line 4465: Not sure if we need a null check here and at line 4436.

clang/lib/Parse/ParseStmt.cpp
```diff
@@ -2427,3 +2447,47 @@ bool Parser::ParseOpenCLUnrollHintAttribute(ParsedAttributes &Attrs) {
...
+    ConsumeToken();
+
+    // Parse bounds expression.
+    ExprResult BoundsRes(ParseBoundsExpression());
+    return Actions.ActOnBoundsFact(WClause, VarName, BoundsRes,
+                                   getCurScope(), Loc);
+  }
+
+  ExprResult ExprRes(ParseExpression());
+  return Actions.ActOnRelopFact(WClause, ExprRes, Loc);
```
sulekhark, line 2469: Instead of passing WClause to functions ActOnBoundsFact and ActOnRelopFact, the functions could return a BoundsFact or a RelopFact (also, return null on error), which then gets added to WClause using the AddFact method of WClause. This will allow us to use ActOnBoundsFact and ActOnRelopFact in slightly different scenarios also.

clang/include/clang/Sema/Sema.h
```diff
@@ -4024,6 +4024,19 @@ class Sema {
...
+  // Checked C: Perform semantic analysis on a where clause.
+  WhereClause *ActOnWhereClause(SourceLocation Loc);
+
+  // Checked C: Perform semantic analysis on a where clause bounds fact.
+  WhereClauseFact *ActOnBoundsFact(IdentifierInfo *Id, Expr *E,
+                                   Scope *CurScope, SourceLocation Loc);
+
+  // Checked C: Perform semantic analysis on a where clause relop fact.
+  WhereClauseFact *ActOnRelopFact(Expr *E, SourceLocation Loc);
+
```
sulekhark, line 4036: Can ActOnBoundsFact return BoundsFact *, and ActOnRelopFact return RelopFact *?

clang/lib/Sema/SemaStmt.cpp
```diff
@@ -4418,3 +4418,65 @@ StmtResult Sema::ActOnCapturedRegionEnd(Stmt *S) {
...
+  BinaryOperator *BO = dyn_cast<BinaryOperator>(E);
+  if (!BO || !BO->isComparisonOp())
+    return nullptr;
+
+  Expr *LHS = IgnoreCasts(BO->getLHS());
+  Expr *RHS = IgnoreCasts(BO->getRHS());
+
+  // TODO: Use the preorder AST to more precisely validate the RelopExpr. This
+  // may involve constant folding the constants in the expression.
+
```
sulekhark, line 4462: This comment is no longer appropriate. It is better to remove it.

clang/test/CheckedC/parsing/valid-where-clause.c
```diff
@@ -0,0 +1,23 @@
...
+
+void valid_cases(_Nt_array_ptr<char> p, _Nt_array_ptr<char> q,
+                 int a, int b) {
+  _Where a < 0;
+  _Where a > 0;
+  _Where a <= 0;
+  _Where a >= 0;
+  _Where a == 0;
+  _Where a != 0;
+  _Where p : bounds(p, p + 1);
```
sulekhark, line 15: Can you add one test case here: _Where p : bounds(q, q+a);

sulekhark (approved):

clang/include/clang/Sema/Sema.h
```diff
@@ -4024,6 +4024,19 @@ class Sema {
...
+  // Ignore value preserving casts.
```
kkjeer, line 4037: This `IgnoreCasts` method also seems to ignore `LValueToRValue` casts, which I don't think we consider value-preserving.

Also, `Expr` has other cast-ignoring methods - should this be an `Expr` method rather than `Sema`?

clang/lib/Parse/ParseStmt.cpp
mgrang: We are currently discarding all facts in case there is an error while parsing any fact.

dtarditi (commented):
Overall, looks great.  I have some suggestions for this change.

mgrang:
Created issue https://github.com/microsoft/checkedc-clang/issues/967 to track updating the section on `where` clause in the Checked C spec.

clang/lib/Parse/ParseStmt.cpp
```diff
@@ -2427,3 +2447,75 @@ bool Parser::ParseOpenCLUnrollHintAttribute(ParsedAttributes &Attrs) {
...
+  WhereClause *WClause = Actions.ActOnWhereClause(Tok.getLocation());
+  if (!WClause) {
+    EmitDiag(Tok);
+    return nullptr;
+  }
+
+  while (true) {
+    WhereClauseFact *Fact = ParseWhereClauseFact();
+    if (!Fact) {
+      EmitDiag(Tok);
```
dtarditi, line 2497: It would be better to skip until an _And clause or a semicolon, whichever comes first.   Programmer time is valuable.  You want to give as many error message as possible.  You only want to bail out when you've hit a semicolon.

clang/lib/Sema/SemaStmt.cpp
```diff
@@ -4418,3 +4418,57 @@ StmtResult Sema::ActOnCapturedRegionEnd(Stmt *S) {
...
+  auto *BO = dyn_cast<BinaryOperator>(TmpE);
+  // Note: isComparisonOp is a function which checks for equality and
+  // relational operators.
+  if (!BO || !BO->isComparisonOp())
+    return nullptr;
+
+  if (!CheckIsNonModifying(BO->getLHS()) ||
+      !CheckIsNonModifying(BO->getRHS()))
+    return nullptr;
```
dtarditi, line 4471: Does CheckIsNonModifying emit an error message?

clang/lib/Sema/SemaStmt.cpp
```diff
@@ -4418,3 +4418,57 @@ StmtResult Sema::ActOnCapturedRegionEnd(Stmt *S) {
...
+
+  VarDecl *VD = Lookup.getAsSingle<VarDecl>();
+  if (!VD)
+    return nullptr;
+
+  return new (Context) BoundsFact(VD, Bounds, Loc);
+}
+
+EqualityOpFact *Sema::ActOnEqualityOpFact(Expr *E, SourceLocation Loc) {
+  // We define an equality-op fact in terms of equality-expressions as defined
```
dtarditi, line 4447: It would be helpful to add a comment "We are checking whether E is an equality expression".

clang/lib/Parse/ParseStmt.cpp
```diff
@@ -2427,3 +2447,75 @@ bool Parser::ParseOpenCLUnrollHintAttribute(ParsedAttributes &Attrs) {
...
+  if (Tok.is(tok::identifier) && NextToken().is(tok::colon)) {
+    IdentifierInfo *VarName = Tok.getIdentifierInfo();
+
+    // Consume the identifier.
+    ConsumeToken();
+    // Consume the ':' token.
+    ConsumeToken();
+
+    // Parse bounds expression.
+    ExprResult BoundsRes(ParseBoundsExpression());
```
dtarditi, line 2468: I am surprised by the error messages in the test cases where you get "expected expression" as an error message.  I would hope to get a "expected a bounds expression".   Perhaps this is a case of the test case containing only part of the error message.

dtarditi (commented):
I think the handling of error cases needs to be improved.  In general, we want error messages to be informative, so a programmers knows how to correct them.  We also want to strive for giving more errors when possible in one compile: a programmer's time is valuable. It is frustrating as a programmer to have to re-run the compiler multiple times because you get a single error per compile.

clang/lib/Parse/ParseStmt.cpp
```diff
@@ -2427,3 +2442,76 @@ bool Parser::ParseOpenCLUnrollHintAttribute(ParsedAttributes &Attrs) {
...
+    if (BoundsRes.isInvalid())
+      return nullptr;
+    return Actions.ActOnBoundsFact(VarName, BoundsRes.get(),
+                                   getCurScope(), Loc);
+  }
+
+  ExprResult ExprRes = Actions.CorrectDelayedTyposInExpr(ParseExpression());
```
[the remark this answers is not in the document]
mgrang: According to our discussions, `ParseExpression` is used to parse general expressions and is the only entry point for parsing equality expressions.

clang/lib/Parse/ParseStmt.cpp
```diff
@@ -2427,3 +2447,75 @@ bool Parser::ParseOpenCLUnrollHintAttribute(ParsedAttributes &Attrs) {
...
+  if (ExprRes.isInvalid())
+    return nullptr;
+  return Actions.ActOnEqualityOpFact(ExprRes.get(), Loc);
+}
+
+WhereClause *Parser::ParseWhereClause() {
+  // Consume the "_Where" token.
+  if (ExpectAndConsume(tok::kw__Where)) {
```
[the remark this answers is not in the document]
mgrang: I am using `ExpectAndConsume` here which will issue an error in case a `_Where` token is not seen.

I have removed `EmitDiag` since the parser and sema functions now handle errors themselves.

clang/lib/Sema/SemaStmt.cpp
mgrang: AFAIK and I checked, `CheckIsNonModifying` does not issue an error message.

clang/test/CheckedC/parsing/invalid-where-clause.c
```diff
@@ -0,0 +1,31 @@
...
+
+// Test parsing errors for incorrect where clause.
+
+void invalid_cases(_Nt_array_ptr<char> p, int a, int b) {
+  _Where ; // expected-error {{invalid where clause}} expected-error {{expected expression}}
+  _Where ;; // expected-error {{invalid where clause}} expected-error {{expected expression}}
+  _Where _Where; // expected-error {{invalid where clause}} expected-error {{expected expression}}
+  _Where _And; // expected-error {{expected expression}} expected-error {{invalid where clause}} expected-error {{expected expression}}
+  _Where _And _And _And; // expected-error {{expected expression}} expected-error {{invalid where clause}} expected-error {{expected expression}}
+  _Where a; // expected-error {{expected comparison operator in where clause equality expression}}
```
mgrang, line 11: expected comparison operator in where clause equality expression --> expected comparison operator in equality expression

sulekhark (approved):
LGTM.

clang/lib/Parse/ParseStmt.cpp
```diff
@@ -2427,3 +2447,65 @@ bool Parser::ParseOpenCLUnrollHintAttribute(ParsedAttributes &Attrs) {
...
+    // Consume the identifier.
+    SourceLocation IdLoc = ConsumeToken();
+    // Consume the ':' token.
+    ConsumeToken();
+
+    // Get the location of the start of bounds expr.
+    SourceLocation BoundsLoc = Tok.getLocation();
+
+    // Parse a bounds decl expression.
+    ExprResult BoundsRes(ParseBoundsExpression());
```
dtarditi, line 2469: Here is a suggestion for how to improve the error messages generated for a where clause when a bounds declaration or bounds expression is missing.  This can be done separately.

The Sema phase in clang tracks the kind of scope that we are in, using a stack of scopes.  You could create a "where" clause scope.  Then in ParseExpr.cpp, you can check whether we are in a where clause to emit a different error message.  See clang\include\sema\clang\Scope.h.   See Parser::EnterScope and Parser::ExitScope in Parser.h.

dtarditi (approved):
Looks great, thank you!

I  added a comment describing how to create a "where" clause scope, which would let the compiler generate a better diagnostic message when a bounds declaration or an expression is not given in a where clause.

PR #973, opened 2021-01-30
Skip free variable checking for member expressions

### LINKED ISSUE #969: Unexpected bounds error after PR #903
This example fails bounds checking after PR #903 with the error message `error: it is not possible to prove argument meets declared bounds for 2nd parameter`. 

```c
void foo(_Array_ptr<char> dest : count(n), _Array_ptr<char> src : count(n), int n);

struct bar {
  _Array_ptr<char> x : count(l);
  int l;
};

void test(_Array_ptr<char> a : count(c), int c) {
  struct bar b = {a, c};
  foo(a, b.x, c);
}
```
Before the PR was merged, the compiler gave a warning instead. `warning: cannot prove argument meets declared bounds for 2nd parameter`. Is this change intentional? If so, what would be the correct way to write the above so that is accepted?

kkjeer:
Fixes #969

The compiler currently does not track equality information for member expressions of the form `a.f`. This can result in false positive errors when checking bounds for free variables (see issue #969 for an example).

This PR modifies the free variable checking logic to skip checking two expressions `E1` and `E2` for free variables if one or both of `E1` or `E2` is or contains a member expression. This extends the check that was already done to skip free variable checking if `E1` or `E2` read memory via a pointer.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -1459,9 +1459,12 @@ namespace {
                                FreeVariablePosition Pos2,
                                EquivExprSets *EquivExprs,
                                FreeVariableListTy &FreeVars) {
-        // If E1 or E2 accesses memory via pointer, we skip because we cannot
+        // If E1 or E2 accesses memory via a pointer, we skip because we cannot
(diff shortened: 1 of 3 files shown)


sulekhark (approved):
LGTM.

PR #976, opened 2021-02-05
Free variables: get all variables in BoundsValueExprs

### LINKED ISSUE #974: Wrong error from #903 when passing _Assume_bounds_cast directly to function with itype
In this program `itype-assume-bounds-error.c`:
```c
void foo(_Ptr<int> p);
void foo_itype(int *p : itype(_Ptr<int>));

void bar() {
  int *p = 0;
  foo(_Assume_bounds_cast<_Ptr<int>>(p));
  foo_itype(_Assume_bounds_cast<_Ptr<int>>(p));

  _Ptr<int> q = _Assume_bounds_cast<_Ptr<int>>(p);
  foo(q);
  foo_itype(q);
}
```

After #903, `clang -c -o /dev/null itype-assume-bounds-error.c` gives the following error:
```
itype-assume-bounds-error.c:7:13: error: it is not possible to prove argument meets declared bounds for 1st parameter
  foo_itype(_Assume_bounds_cast<_Ptr<int>>(p));
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
itype-assume-bounds-error.c:7:13: note: the expected argument bounds use the variable 'p' and there is no relational information involving 'p' and any of the expressions used by the inferred bounds
itype-assume-bounds-error.c:7:13: note: (expanded) expected argument bounds are 'bounds((_Array_ptr<int>)_Assume_bounds_cast<_Ptr<int>>p(count(1)), (_Array_ptr<int>)_Assume_bounds_cast<_Ptr<int>>p(count(1)) + 1)'
itype-assume-bounds-error.c:7:13: note: (expanded) inferred bounds are 'bounds((_Array_ptr<int>)value of p, (_Array_ptr<int>)value of p + 1)'
  foo_itype(_Assume_bounds_cast<_Ptr<int>>(p));
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Before #903, this was a warning.

I'm no expert on the Checked C rules, but it looks like the cause of the error is that the `_Assume_bounds_cast` is present in the expression for the expected bounds but not the actual bounds.  The fact that the error doesn't occur when the casted value is assigned to another local variable before being passed to `foo_itype` leads me to believe this is a bug.  It's interesting that the problem does not occur with a function `foo` that takes the checked type outright rather than using an itype.

This is a simplified version of a failure we saw in [a recently added 3C regression test that has not yet been submitted to this repository](https://github.com/correctcomputation/checkedc-clang/blob/a17fd2de5e33595e1c1f045539f5aace15ba2abb/clang/test/3C/liberal_itypes_ptrptr.c#L147) when we tried to merge from this repository (correctcomputation/checkedc-clang#407).  I'm hoping that the root cause is the same and a fix to the case above will also fix our regression test.

kkjeer:
Fixes #974 

The RecursiveASTVisitor does not seem to traverse the child of a `BoundsExpr` (`E->getTemporaryBinding()`). If an expression uses a variable that is a child of a `BoundsExpr`, then CollectVariableSetHelper will not include it in the set of variables involved in the expression. This can result in spurious free variable errors when checking bounds.

This PR overrides CollectVariableSetHelper::VisitBoundsValueExpr to traverse the bounds value expression's temporary binding.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -605,14 +605,21 @@ namespace {
 
       return true;
     }
+
+    // Ensure that any variables within bounds value expressions are detected
+    // by explicitly traversing the child of a bounds value expression.
+    bool VisitBoundsValueExpr(BoundsValueExpr *E) {
+      TraverseStmt(E->getTemporaryBinding());
+      return true;
+    }
   };
 
   // Collect variables in E without duplication. If E is nullptr, return an
   // empty vector.
   EqualExprTy CollectVariableSet(Sema &SemaRef, Expr *E) {
-      CollectVariableSetHelper Helper(SemaRef);
-      Helper.TraverseStmt(E);
-      return Helper.GetVariableList();
+    CollectVariableSetHelper Helper(SemaRef);
+    Helper.TraverseStmt(E);
+    return Helper.GetVariableList();
   }
 }
 
--- a/clang/test/CheckedC/static-checking/free-variables.c
+++ b/clang/test/CheckedC/static-checking/free-variables.c
@@ -172,3 +172,17 @@ void g(void) {
                                               // expected-note {{inferred bounds are}}
 }
 
+void f_itype(int *p : itype(ptr<int>));
+void f_arrayptr(array_ptr<int> p : count(2));
+
+void h(void) {
+  int *p = 0;
+  f_itype(_Assume_bounds_cast<ptr<int>>(p)); // expected-warning {{cannot prove argument meets declared bounds for 1st parameter}} \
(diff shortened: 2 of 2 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -605,14 +605,21 @@ namespace {
 
       return true;
     }
+
+    // Ensure that any variables within bounds value expressions are detected
+    // by explicitly traversing the child of a bounds value expression.
+    bool VisitBoundsValueExpr(BoundsValueExpr *E) {
+      TraverseStmt(E->getTemporaryBinding());
```
kkjeer, line 612: Traversing the temporary binding of a `BoundsValueExpr` by default in `RecursiveASTVisitor` breaks non-modifying expression detection in inferred bounds. Example from checkedc/tests/typechecking/pointer_types.c:

```
ptr<int> t2 = h1();
```

error: inferred bounds 'bounds(value of h1(), value of h1() + 1)' contain a modifying expression; use a temporary instead

kkjeer:
Closing in favor of #978.

PR #980, opened 2021-02-11
A document evaluating approaches for supporting the implicit

sulekhark:
inclusion of checked header files.

### DIFF
--- a/clang/docs/checkedc/rfc/Include-Checked-Hdrs.md
+++ b/clang/docs/checkedc/rfc/Include-Checked-Hdrs.md
@@ -1,0 +1,128 @@
+# Explicit and Implicit Inclusion of Checked Header Files
+Author: Sulekha Kulkarni, Approver: David Tarditi, Date Approved:
+
+
+# OARP
+
+| Role          | Person |
+|-------------------|---------------|
+| Owner         | Sulekha Kulkarni  |
+| Approver      | David Tarditi     |
+| Reviewers     | Joseph Lloyd, Mandeep Grang, Katherine Kjeer |
+| Reviewers     | Mike Hicks and other folks at CCI |
+
+
+## Goal
+The goal of this document is to record the design considerations for allowing
+programs to include the checked counterparts of system header files either
+explicitly or implicitly during the conversion of these programs to Checked C.
+
+## Problem Summary
+The existing approach for including checked header files, which is to explicitly
+replace an included header file with its checked counterpart, is inconvenient
+for automation by 3C. So we need a way to implicitly include checked header
+files.
+
+## Conditions that a Solution must Satisfy
+ - The solution must not violate the "opt-in" philosophy that we advocate for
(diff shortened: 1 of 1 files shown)


sulekhark:
@mwhicks1 Could you please review this document and let us know you thoughts on this?

mwhicks1 (commented):
We are satisfied with this proposal; I believe it’s an improvement over the status quo, both for automation, and for manual porting/programming. I would like to summarize the position I took on Friday that suggests a slight alternative. See what you think.

The main question I asked during our call on Friday was: If a developer has chosen to use the Checked C `clang`, then has she not, at a minimal level, "opted in?" The next question is, what does this level of opt-in imply? My claim is that, at the least, it implies you should be using the Checked C headers. My reasoning is the following.

## Principle: A Consistent View

The intended design of the Checked C headers is that they can be included by both Checked C code and legacy C code. For the latter, there should be no changes whatsoever to the compilation of that legacy code vs. including the standard header. If that were not the case, there would be a problem for "mixed" files which do.

For example,
```
#include <string_checked.h>
_Checked void foo(nt_array_ptr<char> p,...) {
  strcpy(p,...);
}
void bar(char *q) {
  strcpy(q,...);
}
```
In this code, both the standard type and checked type of `strcpy` are being used, by C code and Checked C code. If the meaning or compilation of the call to `strcpy` from legacy code was different when called when using the prototype of `string_checked.h` vs. when called using `string.h` then you could be in trouble as your project moved forward, with "not yet opted in" source files using one interpretation (since they include `string.h`) and "opted in" files use another (since they include `string_checked.h`).

In sum, based on this reasoning, I think you want, as in normal C, “one header” that all files that will be linked together should use if they are accessing the same functionality. There should be no `*_checked.h` files in your code along with `*.h` versions of them; you want the single `*.h` instead, to ensure this single, consistent view across the whole codebase.

## Practice: Buggy Headers

On the other hand, the checked headers could have bugs on untested/unsupported systems; we have certainly seen this already. As such, as a practical matter, if you are converting your project to use Checked C, you might not want to use the checked headers at the very start, when dipping your toe in. 

That is, you might want to use the Checked C `clang`, to make sure the existing code compiles and works, with legacy headers. Assuming it does, the very next thing I think you'd be inclined to do is use the checked headers, for the whole project, but with no code changes. This is another level "sanity checking" before moving forwards. If things seem OK, then you would start converting files.

But I would not want "mixed headers" in my project, as a rule. I would want to ensure that all parts of the codebase have a consistent view of what the C and Checked C types should be. The use of legacy headers would be for testing/debugging/porting purposes, and only for a short period.

## Alternative Proposal: Opt out

As such, what about going with your proposed approach, but making it possible to “opt out” for debugging reasons? The opt-out could occur via setting the command line flag, as the dual of what you proposed here.

sulekhark:
We accept the OPT-OUT proposal by CCI.  This PR will be closed and a new PR will be opened to take this to a closure in order to ease merge after upgrade to llvm 11.

Issue #985, opened 2021-02-22
Fix failing tests after Clang 11 upgrade


kkjeer:
The following tests are currently failing on ADO after upgrading to Clang 11:

* [x] Clang :: Analysis/diagnostics/sarif-multi-diagnostic-test.c (Windows x64 and x86) - fixed by PR #987
* [x] Clang :: CheckedC/generic-function/generic-func-call-info.c (CI Windows Release x86) - fixed by PR #990
* [x] Clang :: CheckedC/verification/codegen-checked.c (CI Windows Release x86) - fixed by PR #989
* [x] Clang :: CheckedC/verification/codegen-unchecked.c (CI Windows Release x86) - fixed by PR #989

kkjeer:
PR #987 fixed the Clang :: Analysis/diagnostics/sarif-multi-diagnostic-test.c failure.

kkjeer:
PR #989 fixed the Clang :: CheckedC/verification/codegen-checked.c and Clang :: CheckedC/verification/codegen-unchecked.c failures.

PR #986, opened 2021-02-22
Disable sarif-multi-diagnostic-test on Windows

kkjeer:
See issue #985 

After upgrading to clang 11, the Clang :: Analysis/diagnostics/sarif-multi-diagnostic-test.c test fails on Windows. This PR disables the test on Windows while we continue to debug the failure in order to unblock our ADO builds.


kkjeer:
PR #987 fixes this test so it no longer needs to be disabled.

PR #988, opened 2021-02-23
Add clang 11 upgrade information to README

kkjeer:
Update the "source code update" section to reference the latest upgrade to clang 11 rather than the previous upgrade to clang 9.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ On Feb 19, 2021 we updated the checkedc-clang sources to upstream release_110,
 specifically [this](https://github.com/llvm/llvm-project/commit/2e10b7a39b930ef8d9c4362509d8835b221fbc0a) commit.
 
 On Feb 18, 2020 we updated the checkedc-clang sources to upstream release_90,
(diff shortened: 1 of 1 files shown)


sulekhark (approved):
LGTM. Thank you for all your efforts to make this happen!

mgrang:
I suggest we don't overwrite the existing source code update info. This would be helpful in preserving the history of our upgrades. We may decide to move the source upgrade info to its own file and simply have a pointer to it on the checkedc-clang wiki.

kkjeer:
> I suggest we don't overwrite the existing source code update info. This would be helpful in preserving the history of our upgrades. We may decide to move the source upgrade info to its own file and simply have a pointer to it on the checkedc-clang wiki.

I added the clang 9 upgrade information back to the README. On a related note, the link in the document for clang 9 (http://llvm.org/viewvc/llvm-project?view=revision&revision=366428) results in a 404 - is there a different link we can use?

sulekhark (approved):
LGTM.

mgrang (approved):

PR #989, opened 2021-02-24
Use clang_cc1 instead of clang in verification codegen tests

kkjeer:
Tests in clang/test/CheckedC/verification check the IR, so they should use %clang_cc1 instead of %clang in their RUN commands.

The codegen-checked.c and codegen-unchecked.c tests failed after the clang 11 upgrade on Windows X86 builds with no asserts. This PR fixes those test failures so the tests pass on Windows X86.

### DIFF
--- a/clang/test/CheckedC/verification/codegen-checked.c
+++ b/clang/test/CheckedC/verification/codegen-checked.c
@@ -2,7 +2,7 @@
 // There should be a __VERIFIER_error at each dynamic check failed block.
 //
 //
-// RUN: %clang -finject-verifier-calls -emit-llvm -S %s -o - | FileCheck %s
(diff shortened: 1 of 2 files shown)


sulekhark (approved):
LGTM. 

Comment from Mandeep: For tests where we want to check the IR it is not recommended/required to invoke the driver. Also, please see [this](https://llvm.org/docs/TestingGuide.html).

From the clang compiler manual: Emitting value names in LLVM IR increases the size and verbosity of the IR. By default, value names are only emitted in assertion-enabled builds of Clang.

Confirmed that this problem happens on linux also when assertions are disabled.

mgrang (approved):
LGTM.

PR #990, opened 2021-02-24
Guard only an assert with NDEBUG check

mgrang:
In a release without asserts build, we hit the error "function type mismatch".
This was because of the code to determine the type of a _For_any function from
the CallInfo was being protected by the #ifndef NDEBUG flag and hence did not
execute. The main purpose of that check is to guard the assert at the end of
that code. So we moved the check to just before the assert.

### DIFF
--- a/clang/lib/CodeGen/CGCall.cpp
+++ b/clang/lib/CodeGen/CGCall.cpp
@@ -4310,6 +4310,9 @@ RValue CodeGenFunction::EmitCall(const CGFunctionInfo &CallInfo,
     if (getLangOpts().CheckedC)
       IRFuncTy = cast<llvm::FunctionType>(TypeFromVal);
     else {
+    // NOTE: This may cause merge conflicts during checkedc-clang source
+    // upgrades. For more details refer
+    // https://github.com/microsoft/checkedc-clang/pull/990
 #ifndef NDEBUG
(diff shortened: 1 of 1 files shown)


clang/lib/CodeGen/CGCall.cpp
```diff
@@ -4310,10 +4309,12 @@ RValue CodeGenFunction::EmitCall(const CGFunctionInfo &CallInfo,
 
     if (getLangOpts().CheckedC)
       IRFuncTy = cast<llvm::FunctionType>(TypeFromVal);
-    else
+    else {
+#ifndef NDEBUG
```
kkjeer, line 4313: If this is going to be a consistent merge conflict for future upgrades, do we want to document that here?

dtarditi (commented):
Could you file an issue for the TODO in the comment preceding this change.  The original change was made as part of an internship project where there was limited time to make the fix.  It would be good to fix the underlying issue to avoid the need for the if-def.

The change looks fine to me, provided that @kkjeer's comment is addressed.

mgrang:
Filed https://github.com/microsoft/checkedc-clang/issues/991 to fix the underlying issue.

sulekhark (approved):
LGTM.

PR #995, opened 2021-02-27
[Where Clause] Don't allocate space for where clause on error [2/n]

mgrang:
We avoid allocating space for the where clause in case there is an error in one
of the facts. We collect the parsed facts in a local vector and add a fact to
the where clause only if there is no error.

### DIFF
--- a/clang/lib/Parse/ParseStmt.cpp
+++ b/clang/lib/Parse/ParseStmt.cpp
@@ -2651,22 +2651,30 @@ WhereClause *Parser::ParseWhereClause() {
     return nullptr;
   }
 
-  WhereClause *WClause = Actions.ActOnWhereClause(WhereLoc);
-  if (!WClause)
-    return nullptr;
-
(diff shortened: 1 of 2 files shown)


clang/lib/Parse/ParseStmt.cpp
```diff
@@ -2651,22 +2651,30 @@ WhereClause *Parser::ParseWhereClause() {
...
   do {
     WhereClauseFact *Fact = ParseWhereClauseFact();
-    if (!Fact)
+    if (Fact)
+      Facts.push_back(Fact);
+    else {
       IsError = true;
-    else
-      WClause->addFact(Fact);
+      delete Fact;
```
sulekhark, line 2665: Isn't `Fact` null here?

mgrang:
@sulekhark and I recognized that deferring the creation of the `WhereClause` object to after the facts are parsed does not save much in memory and in fact makes the code harder to read and reason about. So it's best we don't go ahead with this PR.

PR #997, opened 2021-03-02
[Where Clause] Introduce WhereClauseScope to emit better diagnostics [2/n]

mgrang:
We introduce a new scope for where clauses. Before parsing a where clause we
enter the `WhereClauseScope` and at the end of parsing we exit the scope. In `Sema`
we check if we are in the `WhereClauseScope` and emit the relevant diagnostics.

### DIFF
--- a/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -10912,6 +10912,9 @@ def err_probability_out_of_range : Error<
 def err_expected_comparison_op_in_equality_expr : Error<
   "expected comparison operator in equality expression">;
 
+def err_invalid_expr_in_where_clause : Error<
+  "invalid expr in where clause, expected bounds declaration or equality expression">;
+
 // Checked C diagnostic messages
 let CategoryName = "Checked C" in {
 
--- a/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -10913,7 +10913,7 @@ def err_expected_comparison_op_in_equality_expr : Error<
   "expected comparison operator in equality expression">;
 
 def err_invalid_expr_in_where_clause : Error<
-  "invalid expr in where clause, expected bounds declaration or equality expression">;
+  "invalid expression in where clause, expected bounds declaration or equality expression">;
 
 // Checked C diagnostic messages
 let CategoryName = "Checked C" in {
--- a/clang/lib/Sema/SemaStmt.cpp
+++ b/clang/lib/Sema/SemaStmt.cpp
@@ -4557,7 +4557,8 @@ EqualityOpFact *Sema::ActOnEqualityOpFact(Expr *E, SourceLocation ExprLoc) {
   auto *BO = dyn_cast_or_null<BinaryOperator>(TmpE);
 
   if (!BO) {
-    Diag(ExprLoc, diag::err_invalid_expr_in_where_clause);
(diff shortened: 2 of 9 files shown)


mgrang:
https://github.com/microsoft/checkedc-clang/pull/999 depends on this PR.

clang/test/CheckedC/parsing/invalid-where-clause.c
```diff
@@ -5,19 +5,19 @@
...
+  _Where ; // expected-error {{expected bounds declaration or equality expression in where clause}}
+  _Where ;;;;; // expected-error {{expected bounds declaration or equality expression in where clause}}
+  _Where _Where; // expected-error {{expected bounds declaration or equality expression in where clause}} expected-error {{expected bounds declaration or equality expression in where clause}}
+  _Where _And; // expected-error {{expected bounds declaration or equality expression in where clause}} expected-error {{expected bounds declaration or equality expression in where clause}}
+  _Where _And _And _And; // expected-error {{expected bounds declaration or equality expression in where clause}} expected-error {{expected bounds declaration or equality expression in where clause}} expected-error {{expected bounds declaration or equality expression in where clause}} expected-error {{expected bounds declaration or equality expression in where clause}}
   _Where a; // expected-error {{expected comparison operator in equality expression}}
```
sulekhark, line 13: Could this error message be improved too? Now that we have the WhereClause scope information, is it possible not to assume that `a` is the prefix of a potential equality expression?

mgrang: We enter `ParseBoundsExpression` only when we see an `identifier` followed by a `:`. Otherwise, we enter `ParseExpression`. According to this logic, we should assume that an `identifier` not followed by a `:` is the prefix of an equality expression. Hence, I think the error message is correct.

mgrang: This is now addressed in the latest change.

clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -10912,6 +10912,9 @@ def err_probability_out_of_range : Error<
 def err_expected_comparison_op_in_equality_expr : Error<
   "expected comparison operator in equality expression">;
 
+def err_invalid_expr_in_where_clause : Error<
+  "invalid expr in where clause, expected bounds declaration or equality expression">;
```
sulekhark, line 10916: Perhaps "expr" could be expanded to "expression"?

clang/lib/Sema/SemaStmt.cpp
```diff
@@ -4556,8 +4556,13 @@ EqualityOpFact *Sema::ActOnEqualityOpFact(Expr *E, SourceLocation ExprLoc) {
 
   auto *BO = dyn_cast_or_null<BinaryOperator>(TmpE);
 
+  if (!BO) {
```
sulekhark, line 4559: In addition to the condition `!BO`, it will be great if we can ensure that we are in a WhereClause scope here because the error message is only relevant in the context of the where clause.

mgrang: This function (`ActOnEqualityOpFact`) is only called when we are in a `WhereClauseScope` (ie: while parsing a where clause). So I am not sure how explicitly checking for `WhereClauseScope` would help?

sulekhark: It may be a good idea to keep ActOnEqualityOpFact general - it may be called (in the future) on "inferred facts" which may not always be embedded in a Where clause.

sulekhark (approved):
LGTM. Thank you!

PR #1000, opened 2021-03-03
[Where Clause] Support where clauses on ExprStmt [4/n]

mgrang:
We add support for where clauses on ExprStmts like:
  ```
  int a;
  a = 0 _Where a < 1 _And a != 2;
  ```

### DIFF
--- a/clang/lib/Parse/ParseStmt.cpp
+++ b/clang/lib/Parse/ParseStmt.cpp
@@ -2637,7 +2637,19 @@ WhereClauseFact *Parser::ParseWhereClauseFact() {
 
   // Parse an equality expression.
   SourceLocation ExprLoc = Tok.getLocation();
-  ExprResult ExprRes = Actions.CorrectDelayedTyposInExpr(ParseExpression());
+
+  // ParseExpression parses expressions including top-level commas. So
+  // declarations like the following are not parsed correctly.
+  // int a _Where a == 1, b;
+  // If fails because it parses "a == 1, b" as one expression.
+
+  // ParseAssignmentExpression parses an expression not including top-level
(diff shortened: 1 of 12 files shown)


clang/lib/Parse/ParseStmt.cpp
```diff
@@ -496,9 +496,28 @@ StmtResult Parser::ParseExprStatement(ParsedStmtContext StmtCtx) {
...
+    WClause = ParseWhereClause();
+    if (!WClause)
+      return StmtError();
+  }
+
   // Otherwise, eat the semicolon.
   ExpectAndConsumeSemi(diag::err_expected_semi_after_expr);
-  return handleExprStmt(Expr, StmtCtx);
+  StmtResult StmtRes = handleExprStmt(Expr, StmtCtx);
+  if (StmtRes.isInvalid() || !WClause)
```
sulekhark, line 510: The check `!WClause` may not be necessary - it has already been checked above.

mgrang: The `!WClause` check above is inside `if (getLangOpts().CheckedC && Tok.is(tok::kw__Where))`. If we are not in a where clause then we don't want the rest of the code below to execute. That's why I have this check here.

To make the logic cleaner, we can also have a dedicated variable which checks if where clause is present or not.

sulekhark: Yes, thank you for the clarification!

sulekhark (approved):
LGTM!

mgrang:
Closed in favor of https://github.com/microsoft/checkedc-clang/pull/1007

PR #1001, opened 2021-03-04
Lexicographic ordering for PreorderASTs

kkjeer:
This PR implements lexicographic ordering for PreorderASTs that uses the existing ordering in CanonBounds.cpp. Previously, PreorderAST had an `IsEqual` method that returned true if two ASTs were equal (using `Lexicographic::CompareExpr` to compare expressions in leaf nodes). This PR replaces `IsEqual` with a `Compare` method that returns a `Lexicographic::Result` - `LessThan`, `Equal`, or `GreaterThan` that continues to use `Lexicographic::CompareExpr` to compare expressions in leaf nodes.

This ordering will be used in lvalue generalization work to avoid an expensive linear search when computing the `AbstractSet` to which an lvalue expression belongs (see the [lvalue generalization design doc](https://github.com/microsoft/checkedc-clang/blob/master/clang/docs/checkedc/Bounds-Checking-for-LValue-Expressions.md)).

### DIFF
--- a/clang/lib/AST/PreorderAST.cpp
+++ b/clang/lib/AST/PreorderAST.cpp
@@ -427,8 +427,8 @@ Result PreorderAST::Compare(Node *N1, Node *N2) {
     if (O1->Opc > O2->Opc)
       return Result::GreaterThan;
 
-    unsigned ChildCount1 = O1->Children.size(),
-             ChildCount2 = O2->Children.size();
+    size_t ChildCount1 = O1->Children.size(),
+           ChildCount2 = O2->Children.size();
 
     // If the number of children of the two nodes mismatch.
     if (ChildCount1 < ChildCount2)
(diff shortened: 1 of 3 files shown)


clang/lib/AST/PreorderAST.cpp
```diff
@@ -403,51 +403,62 @@ bool PreorderAST::GetDerefOffset(Node *UpperNode, Node *DerefNode,
...
 
     // If the Opcodes mismatch.
-    if (O1->Opc != O2->Opc)
-      return false;
+    if (O1->Opc < O2->Opc)
+      return Result::LessThan;
+    if (O1->Opc > O2->Opc)
+      return Result::GreaterThan;
+
+    unsigned ChildCount1 = O1->Children.size(),
```
mgrang, line 430: `O1->Children` and `O2->Children` are `llvm::SmallVector`s and their sizes are of type `size_t`. So it's better to use `size_t` instead of `unsigned`.

mgrang (approved):
LGTM.

sulekhark (approved):
LGTM.

PR #1004, opened 2021-03-11
[Where Clause] Support where clauses on declarations [3/n]

mgrang:
We add support for where clauses:

1. For variable declarations, like:
```
int a = 1 _Where a > 0, b, c, d _Where d < 1;
```

2. For function parameters, like:
```
void f(int a _Where a < 0, int b, int c _Where c > 0);
void f(_Nt_Array_ptr<int> p _Where p : bounds(p, p + n), int n);
void f(int *p : itype(_Ptr<int>) _Where p == 0);
```

### DIFF
--- a/clang/lib/Parse/ParseDecl.cpp
+++ b/clang/lib/Parse/ParseDecl.cpp
@@ -1881,21 +1881,27 @@ void Parser::ExitQuantifiedTypeScope(DeclSpec &DS) {
   }
 }
 
-void Parser::ParseWhereClauseOnDecl(Decl *D) {
+bool Parser::ParseWhereClauseOnDecl(Decl *D) {
   if (!StartsWhereClause(Tok))
-    return;
+    return true;
 
-  if (!D)
-    llvm_unreachable("invalid where clause on empty Decl");
+  if (!D) {
+    Diag(Tok, diag::err_invalid_decl_where_clause);
+    return false;
+  }
 
   WhereClause *WClause = ParseWhereClause();
   if (!WClause)
-    return;
+    return false;
 
   if (auto *PD = dyn_cast<ParmVarDecl>(D))
     PD->setWhereClause(WClause);
   else if (auto *VD = dyn_cast<VarDecl>(D))
     VD->setWhereClause(WClause);
+  else
+    return false;
+
+  return true;
 }
 
 /// ParseDeclGroup - Having concluded that this is either a function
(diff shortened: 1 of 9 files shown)


clang/test/CheckedC/parsing/invalid-where-clause.c
```diff
@@ -36,3 +36,13 @@ void invalid_cases(_Nt_array_ptr<char> p, int a, int b) {
...
+
+void f1(int a _Where a, _Nt_array_ptr<int> p : count(0) _Where p :); // expected-error {{invalid expression in where clause, expected bounds declaration or equality expression}} expected-error {{expected bounds expression}}
+
+void f2(int *p : itype(_Ptr<int>) _Where p, int n); // expected-error {{invalid expression in where clause, expected bounds declaration or equality expression}}
+
+void f3(int a _Where a == 1, a == 1); // expected-error {{unknown type name 'a'}} expected-error {{expected ')'}} expected-note {{to match this '('}}
+
+void f4(int a _Where a == 1, b == 1); // expected-error {{unknown type name 'b'}} expected-error {{expected ')'}} expected-note {{to match this '('}}
+
+void f5(int a _Where a,a); // expected-error {{redefinition of parameter 'a'}} expected-error {{invalid expression in where clause, expected bounds declaration or equality expression}} expected-warning {{type specifier missing, defaults to 'int'}} expected-note {{previous declaration is here}}
```
mgrang, line 48: Add negative test cases for:
1. Unbalanced parentheses.
2. Where clause occurring w/o any parameters in a function decl.

kkjeer (approved):
LGTM

clang/lib/Parse/ParseDecl.cpp
```diff
@@ -4503,7 +4535,8 @@ void Parser::ParseStructUnionBody(SourceLocation RecordLoc,
     FieldDecl *FieldDecl = Pair.first;
     std::unique_ptr<CachedTokens> Tokens = std::move(Pair.second);
     BoundsAnnotations Annots;
-    if (DeferredParseBoundsExpression(std::move(Tokens), Annots, DeclaratorsInfo.D))
+    if (DeferredParseBoundsAnnotations(std::move(Tokens),
+                                       Annots, DeclaratorsInfo.D))
```
sulekhark, line 4539: In another place where `DeferredParseBoundsAnnotations` is called, `ActOnInvalidBoundsDecl` and `ActOnBoundsDecl` are called only when the first token in `Tokens` is not `_Where`. Is such a check not required here? If it is not required, maybe a comment here will be helpful?

clang/lib/Parse/ParseExpr.cpp
```diff
@@ -4027,14 +4107,15 @@ bool Parser::ConsumeAndStoreBoundsExpression(CachedTokens &Toks) {
 
 /// Given a list of tokens that have the same shape as a bounds
 /// expression, parse them to create a bounds expression.  Delete
-/// the list of tokens at the end.
-///
+/// the list of tokens at the end. Also parses where clause occurring on a
+/// decl.
```
sulekhark, line 4111: The last line of this comment looks like an after-thought. Could you please integrate it more with the body of the entire comment? Also, in the comment lines below, perhaps you could mention that the resulting `where` clause is attached to `ThisDecl`.

sulekhark (approved):
LGTM! Thank you! I have two minor comments - it will be great if you could address them!

PR #1005, opened 2021-03-18
Move expr utility functions from SemaBounds.cpp to new file ExprUtils.cpp

mgrang:
To make functions like `CreateBinaryOperator`, etc (which were defined in
`SemaBounds.cpp`) easier to call from other files we moved the class
`ExprCreatorUtil` to a new file called `ExprUtils.cpp`.  The two overloaded
versions of the function `CreateIntegerLiteral` have also been made members of
the `ExprCreatorUtil` class and moved to the same file.


sulekhark (approved):
LGTM! Thank you!

kkjeer (approved):
LGTM. Thanks!

PR #1006, opened 2021-03-18
Move funcs UpdateWidenedBounds and ResetKilledBounds to BoundsAnalysis [NFC]

mgrang:
We moved the functions `UpdateWidenedBounds` and `ResetKilledBounds` from
`SemaBounds.cpp` to `BoundsAnalysis.cpp` to create a cleaner separation between the
two files. `UpdateWidenedBounds` now returns a map of variables to their updated
bounds expressions, while `ResetKilledBounds` returns a set of variables killed
by the current statement. Inside `SemaBounds.cpp` we now simply update the
`CheckingState` with the updated bounds.

### DIFF
--- a/clang/include/clang/Sema/BoundsAnalysis.h
+++ b/clang/include/clang/Sema/BoundsAnalysis.h
@@ -17,6 +17,7 @@
 #define LLVM_CLANG_BOUNDS_ANALYSIS_H
 
 #include "clang/AST/CanonBounds.h"
+#include "clang/AST/ExprUtils.h"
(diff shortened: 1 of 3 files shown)


dtarditi (approved):
Could you add a comment describing the BoundsAnalysis class?   It'd be helpful to have a summary of what it odes.   It would be nice to move QueueSet some place else, but that can be done as a separate change later.

sulekhark (approved):
LGTM. Thank you!

kkjeer (approved):
LGTM. Thanks!

PR #1011, opened 2021-03-23
[LValue Generalization] Make AbstractSetManager a non-static class [3/n]

kkjeer:
This PR makes AbstractManager a non-static class. In the next stage of lvalue generalization, when using AbstractSets as the keys in CheckBoundsDeclarations and BoundsAnalysis, an AbstractSetManager instance will be created in CheckBoundsDeclarations::TraverseCFG and passed to BoundsAnalysis::WidenBounds and checking methods in CheckBoundsDeclarations.


sulekhark (approved):
LGTM! Thanks!

dtarditi (approved):
Looks good, thanks.

PR #1019, opened 2021-03-31
Fix for CompareDecl for strict weak ordering

kkjeer:
When a `Decl D1` occurs before `D2`, `CompareDecl` returns `Result::LessThan`. If `D1` occurs after `D2`, `CompareDecl` should return `Result::GreaterThan`. This helps ensure strict weak ordering on `Decls` so that, if `D1 < D2`, `D2 > D1`.


mgrang (approved):
LGTM.

sulekhark (approved):
LGTM.

PR #1022, opened 2021-04-05
Add a map from a VarDecl to its first use

kkjeer:
The AbstractSetManager method GetOrCreateAbstractSet for a VarDecl needs to compute a DeclRefExpr that is a use of the VarDecl. This DeclRefExpr is used to create a canonical form to search for an existing AbstractSet and, if necessary, is used as the representative expression when creating a new AbstractSet.

In order to avoid creating a new DeclRefExpr each time GetOrCreateAbstractSet(VarDecl *V) is called, we introduce a map called `VarUses` from a `VarDecl *` to the `DeclRefExpr *` (if any) that is the first use of the `VarDecl *` within a function body. This can help reduce the need to create a new DeclRefExpr in GetOrCreateAbstractSet.

Certain VarDecls may not have a DeclRefExpr use within a function. In that case, GetOrCreateAbstractSet will create a new DeclRefExpr and modify its VarUses member. This can occur for variable declarations with declared bounds that are not referenced within a function, but we still need to create AbstractSets for these variable declarations in order to store them in the ObservedBounds map.

### DIFF
--- a/clang/include/clang/Sema/Sema.h
+++ b/clang/include/clang/Sema/Sema.h
@@ -5819,7 +5819,7 @@ public:
   };
 
   // Map a VarDecl to its first use.
-  typedef llvm::DenseMap<const VarDecl *, DeclRefExpr *> VarDeclUsage;
+  using VarDeclUsage = llvm::DenseMap<const VarDecl *, DeclRefExpr *>;
 
   /// \brief Compute a mapping from statements that modify lvalues to
   /// in-scope bounds declarations that depend on those lvalues.
--- a/clang/include/clang/Sema/Sema.h
+++ b/clang/include/clang/Sema/Sema.h
@@ -5826,6 +5826,9 @@ public:
   /// FD is the function being declared and Body is the body of the
   /// function.   They are passed in separately because Body hasn't
   /// been attached to FD yet.
+  /// ComputeBoundsDependencies also computes a mapping from VarDecls with
+  /// bounds expressions to the DeclRefExpr (if any) that is the first use
+  /// of the VarDecl.
   void ComputeBoundsDependencies(ModifiedBoundsDependencies &Tracker,
                                  VarDeclUsage &VarUses,
                                  FunctionDecl *FD, Stmt *Body);
--- a/clang/include/clang/Sema/Sema.h
+++ b/clang/include/clang/Sema/Sema.h
@@ -5818,12 +5818,16 @@ public:
    DependentBounds Tracker;
   };
 
+  // Map a VarDecl to its first use.
+  typedef llvm::DenseMap<const VarDecl *, DeclRefExpr *> VarDeclUsage;
+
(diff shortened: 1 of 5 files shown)


clang/lib/AST/AbstractSet.cpp
```diff
@@ -32,7 +32,25 @@ const AbstractSet *AbstractSetManager::GetOrCreateAbstractSet(Expr *E) {
...
+  // The parameter declaration `unused` does not have a DeclRefExpr in the
+  // VarUses map, but the statement i = 0 invalidates the observed bounds
+  // of V.
+  DeclRefExpr *VarUse = nullptr;
+  auto It = VarUses.find(V);
+  if (It != VarUses.end()) {
+    VarUse = It->second;
+  } else {
+    VarDecl *D = const_cast<VarDecl *>(V);
+    VarUse = ExprCreatorUtil::CreateVarUse(S, D);
```
sulekhark, line 52: The created `VarUse` may become leaked memory when the `VarUses` map goes out of scope (when the function in which `VarUses` is a local variable, exits). Perhaps we need to free up all the created `VarUse` elements in the  `Clear` method of `AbstractSetManager` class.

kkjeer:
An empty VarUses map is created in the following Sema methods. This VarUses map is not populated with first uses of any VarDecls, which would be problematic if GetOrCreateAbstractSet is ever called on a VarDecl. If GetOrCreateAbstractSet is called on a VarDecl when the VarUses map is empty, then GetOrCreateAbstractSet would have to create a DeclRefExpr that is a use of the VarDecl. We want to avoid creating these DeclRefExprs where possible.

1. CreateCountForArrayType
2. MakeAssignmentImplicitCastExplicit
3. CheckTopLevelBoundsDecls
4. ExpandBoundsToRange
5. GetLValueDeclaredBounds

Of these methods, only CheckTopLevelBoundsDecls may call GetOrCreateAbstractSet, by calling CheckBoundsDeclarations::CheckVarDecl. However, in CheckVarDecl, GetOrCreateAbstractSet is only called on a DeclRefExpr. CheckVarDecl creates a TargetDeclRef that is used to create an rvalue expression for the VarDecl. This rvalue expression is included in the EquivExprs set of sets of equivalent expressions to record equality between the declared variable and the initializer (if the VarDecl has an initializer). The TargetDeclRef is used to get or create the AbstractSet for the VarDecl.

clang/include/clang/Sema/Sema.h
```diff
@@ -5818,12 +5818,16 @@ class Sema final {
    DependentBounds Tracker;
   };
 
+  // Map a VarDecl to its first use.
+  typedef llvm::DenseMap<const VarDecl *, DeclRefExpr *> VarDeclUsage;
```
mgrang, line 5822: In C++, we can use `using` instead of `typedef`.

sulekhark (approved):
LGTM!

PR #1024, opened 2021-04-12
Change the argument to VariableOccurrenceCount from DeclRefExpr to ValueDecl

kkjeer:
This PR enables use cases where VariableOccurrenceCount can be called on a Decl rather than an Expr. Given a `DeclRefExpr *E`, it is simple to retrieve the `ValueDecl *` associated with `E`. However, given a `ValueDecl *V`, it is not easy to retrieve a `DeclRefExpr *E` that is a use of `V`.


sulekhark (approved):
LGTM. Thank you!

Issue #1026, opened 2021-04-13
Enabling the LifetimeEnds CFGElement crashes the compiler during CFG construction


sulekhark:
Test case:

extern int bar(char *s, int n);

// test case from CodeGen/lifetime2.c
void backpatched_goto2() {
  {
    char x;
  label:
    bar(&x, 1);
  }
  goto label;
}
Command line:  ./bin/clang -cc1 -analyze -analyzer-checker=debug.DumpCFG -analyzer-config cfg-lifetime=true,cfg-temporary-dtors=false,cfg-rich-constructors=false -analyzer-config cfg-implicit-dtors=false ../DEBUG_lifetime/r1.cpp

Crash Backtrace:

Starting program: /home/sulekha/Work/checkedc-clang/build/bin/clang -cc1 -analyze -analyzer-checker=debug.DumpCFG -analyzer-config cfg-lifetime=true,cfg-temporary-dtors=false,cfg-rich-constructors=false -analyzer-config cfg-implicit-dtors=false ../DEBUG_lifetime/r1.cpp
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
clang: /home/sulekha/Work/checkedc-clang/src/clang/lib/Analysis/CFG.cpp:333: int {anonymous}::LocalScope::const_iterator::distance({anonymous}::LocalScope::const_iterator): Assertion `F != const_iterator() && "L iterator is not reachable from F iterator."' failed.

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) where
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7a57859 in __GI_abort () at abort.c:79
#2  0x00007ffff7a57729 in __assert_fail_base (fmt=0x7ffff7bed588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=0x555561fbc050 "F != const_iterator() && \"L iterator is not reachable from F iterator.\"",
    file=0x555561fbbd70 "/home/sulekha/Work/checkedc-clang/src/clang/lib/Analysis/CFG.cpp", line=333, function=<optimized out>) at assert.c:92
[...]

PR #1028, opened 2021-04-13
Fix for #1026: Enabling LifetimeEnds CFGElement crashes the compiler during CFG construction

sulekhark:
The crash happens in some scenarios where a `goto` statement has to be backpatched with `CFGLifetimeEnds` elements. During CFG construction, the statements enclosed within a compound statement are iterated through in reverse (i.e., last statement is processed first). In the case where the target of the `goto` statement occurs earlier than the `goto`, the  `goto` statement is processed before its target labeled statement is processed. Now, the variables that go out of scope when the `goto` is executed depend on the position of the target of the `goto` statement. Therefore, the `LifetimeEnds` `CFGElements` for these variables, in the case mentioned above, have to be backpatched into the basic block containing the `goto` only after the target of the `goto` has been processed.

The bug is in determining the set of variables that go out of scope, while performing the backpatching. Consider the following block structure:
```
 {
      <variable A declared>
      {
            <variable B declared>
            label:
               ......
      }
      <variable C declared>
      {
             <variable D declared>
             goto label
      }
      <variable E declared>
 }     
``` 
The variables that go out of scope when the `goto` is executed are: `D` and `C`. The above block structure represents the general case that is not handled during backpatching. The cases that are handled are when the scope of the `goto` target is an ancestor of the scope that contains the `goto` statement in the tree of scopes within the function.


clang/lib/Analysis/CFG.cpp
```diff
@@ -2085,9 +2085,24 @@ void CFGBuilder::prependAutomaticObjLifetimeWithTerminator(
...
+  //
+  // To go from B to E, one first goes up the scopes from B
+  // to PofB, then sideways in one scope from PofB to PofE and then down
+  // the scopes from PofE to E.
+  // The lifetime of all objects between B and PofE end.
+  //
+  // Get E's ancestor that has the same LocalScope as one of B's ancestors.
+  LocalScope::const_iterator PofE = E.shared_parent(B);
+  int Dist = B.distance(PofE);
+  if (Dist <= 0)
```
mgrang, line 2100: Can the distance be less than 0?

sulekhark: While `Dist` cannot be less than 0 if the `distance` function is called as intended, it is possible to provide arguments to it such that the distance between two LocalScope iterators is less than zero. For example, this can happen if the control comes here for a "forward" jumping `goto` statement (which should not ideally happen). Moreover, this kind of check is there in two other places too and they have used a `<=` check (see lines 1770 - 1777).

mgrang (approved):
LGTM.

kkjeer (approved):

dtarditi (commented):
@sulekhark and I discussed this change.  Overall, looks good.  I suggested using full variables names instead of abbreviations like `PofB`.   Also, this is not a parent, but an ancestor of the scope. 

I thought the comment was somewhat confusing because it doesn't explain is actually happening. We find a common shared scope and end the lifetime of any variables defined in the scopes nested within the shared scope that enclose the goto statement.  If you goto a scope A from a scope B that lexically scope A, the shared scope is scope B, so no variable definitions end.  I understand the comment is taken from another place in the file; the original comment is confusing.

dtarditi (approved):
Looks good, thanks.

Issue #1010, opened 2021-03-22
Move QueueSet out of BoundsAnalysis.h


mgrang:
The class [QueueSet](https://github.com/microsoft/checkedc-clang/blob/master/clang/include/clang/Sema/BoundsAnalysis.h#L25) was added to `BoundsAnalysis.h` to enable processing unique elements in the `FIFO` order. However, this class is generic enough and can be used in other places. So it should be moved out of `BoundsAnalysis.h` (maybe either to its own file in `clang/include` or to `include/llvm/ADT`).

This was suggested in a review [comment ](https://github.com/microsoft/checkedc-clang/pull/1006#pullrequestreview-616835521)by David.

mgrang:
PR https://github.com/microsoft/checkedc-clang/pull/1029

PR #1029, opened 2021-04-13
Move QueueSet out of BoundsAnalysis.h

mgrang:
The class `QueueSet` was added to `BoundsAnalysis.h` to enable processing unique
elements in the FIFO order. However, this class is generic enough and can be
used in other places. So we are moving it out of `BoundsAnalysis.h`.

This fixes https://github.com/microsoft/checkedc-clang/issues/1010


sulekhark (approved):
LGTM!

kkjeer (approved):
LGTM

PR #1030, opened 2021-04-14
[LValue Generalization] Canonicalize MemberExprs [6/n]

kkjeer:
This PR modifies PreorderAST so that it can create canonical forms for member expressions such as `a->f`, `(*a).f`, `a[0].f`, etc.

Three new `Node` kinds are introduced: `MemberNode` for canonicalizing member expressions, `UnaryOperatorNode` for canonicalizing unary operators and array subscripts such as `*p`, `*(p + i)`, `p[i]`, etc., and `ImplicitCastNode` for canonicalizing implicit cast expressions such as `LValueToRValue(e)`.

This is part of the ongoing work to generalize bounds checking to lvalue expressions. In order to create an `AbstractSet` for an lvalue expression `e` to use as the key in `ObservedBounds` to track the inferred bounds of the value produced by `e`, we need to use `PreorderAST` to create a canonical form for `e`.

### DIFF
--- a/clang/include/clang/AST/PreorderAST.h
+++ b/clang/include/clang/AST/PreorderAST.h
@@ -51,8 +51,10 @@ namespace clang {
   class BinaryOperatorNode : public Node {
   public:
     BinaryOperator::Opcode Opc;
-    // Note: A BinaryOperatorNode has a list of children because the preorder
-    // AST is an n-ary tree.
+    // A BinaryOperatorNode representing a commutative and associative binary
+    // operation may have more than two children because of coalescing.
+    // Ex: a + (b + c) will be represented by one BinaryOperatorNode for +
+    // with three children nodes for a, b and c after coalescing.
     llvm::SmallVector<Node *, 2> Children;
 
     BinaryOperatorNode(BinaryOperator::Opcode Opc, Node *Parent) :
@@ -142,7 +144,7 @@ namespace clang {
     void Create(Expr *E, Node *Parent = nullptr);
 
     // Create a BinaryOperatorNode with an addition operator and two children
-    // (E and 0), and add the created BinaryOperatorNode to the Parent node.
+    // (E and 0), and attach the created BinaryOperatorNode to the Parent node.
     // @param[in] E is the expression that is one of the two children of
     // the created BinaryOperatorNode (the other child is 0).
     // @param[in] Parent is the parent of the created BinaryOperatorNode.
@@ -171,7 +173,8 @@ namespace clang {
     // to control when to stop recursive coalescing.
     void Coalesce(Node *N, bool &Changed);
 
-    // Sort the children expressions in a non-leaf Node of the AST.
+    // Recursively descend the PreorderAST to sort the children of all
+    // BinaryOperatorNodes if the binary operator is commutative.
     // @param[in] N is current node of the AST. Initial value is Root.
     void Sort(Node *N);
 
--- a/clang/include/clang/AST/PreorderAST.h
+++ b/clang/include/clang/AST/PreorderAST.h
@@ -32,6 +32,7 @@ namespace clang {
     enum class NodeKind {
       OperatorNode,
       UnaryOperatorNode,
+      MemberNode,
       ImplicitCastNode,
       LeafExprNode
     };
@@ -75,6 +76,21 @@ namespace clang {
 
     static bool classof(const Node *N) {
       return N->Kind == NodeKind::UnaryOperatorNode;
+    }
+  };
+
+  class MemberNode : public Node {
+  public:
+    Node *Base = nullptr;
+    ValueDecl *Field = nullptr;
+    bool IsArrow;
+
+    MemberNode(ValueDecl *Field, bool IsArrow, Node *Parent) :
+      Node(NodeKind::MemberNode, Parent),
+      Field(Field), IsArrow(IsArrow) {}
+
+    static bool classof(const Node *N) {
+      return N->Kind == NodeKind::MemberNode;
     }
   };
 
--- a/clang/include/clang/AST/PreorderAST.h
+++ b/clang/include/clang/AST/PreorderAST.h
@@ -23,14 +23,13 @@
 
 namespace clang {
   using Result = Lexicographic::Result;
-  class OperatorNode;
 
   class Node {
   public:
     // Nodes with two different kinds are sorted according to the order in
     // which their kinds appear in this enum.
     enum class NodeKind {
-      OperatorNode,
+      BinaryOperatorNode,
       UnaryOperatorNode,
       MemberNode,
       ImplicitCastNode,
@@ -44,19 +43,19 @@ namespace clang {
       Kind(Kind), Parent(Parent) {}
   };
 
-  class OperatorNode : public Node {
+  class BinaryOperatorNode : public Node {
   public:
     BinaryOperator::Opcode Opc;
-    // Note: An OperatorNode has a list of children because the preorder AST is
-    // an n-ary tree.
+    // Note: A BinaryOperatorNode has a list of children because the preorder
+    // AST is an n-ary tree.
     llvm::SmallVector<Node *, 2> Children;
 
-    OperatorNode(BinaryOperator::Opcode Opc, Node *Parent) :
-      Node(NodeKind::OperatorNode, Parent),
+    BinaryOperatorNode(BinaryOperator::Opcode Opc, Node *Parent) :
+      Node(NodeKind::BinaryOperatorNode, Parent),
       Opc(Opc) {}
 
     static bool classof(const Node *N) {
-      return N->Kind == NodeKind::OperatorNode;
+      return N->Kind == NodeKind::BinaryOperatorNode;
     }
 
     // Is the operator commutative and associative?
@@ -137,11 +136,11 @@ namespace clang {
     // @param[in] Parent is the parent of the new node.
     void Create(Expr *E, Node *Parent = nullptr);
 
-    // Create an OperatorNode with an addition operator and two children
-    // (E and 0), and add the created OperatorNode to the Parent node.
+    // Create a BinaryOperatorNode with an addition operator and two children
+    // (E and 0), and add the created BinaryOperatorNode to the Parent node.
     // @param[in] E is the expression that is one of the two children of
-    // the created OperatorNode (the other child is 0).
-    // @param[in] Parent is the parent of the created OperatorNode.
+    // the created BinaryOperatorNode (the other child is 0).
+    // @param[in] Parent is the parent of the created BinaryOperatorNode.
     void AddZero(Expr *E, Node *Parent);
 
     // Add a new node to the AST.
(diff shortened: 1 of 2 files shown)


clang/include/clang/AST/PreorderAST.h
```diff
@@ -27,12 +27,20 @@ namespace clang {
 
   class Node {
   public:
-    enum class NodeKind { OperatorNode, LeafExprNode };
+    // Nodes with two different kinds are sorted according to the order in
+    // which their kinds appear in this enum.
+    enum class NodeKind {
+      OperatorNode,
```
mgrang, line 33: Maybe we should rename `OperatorNode` to `BinaryOperatorNode` since we now have a separate `UnaryOperatorNode`.

clang/include/clang/AST/PreorderAST.h
```diff
@@ -57,11 +65,54 @@ namespace clang {
...
+      return N->Kind == NodeKind::ImplicitCastNode;
+    }
+  };
+
   class LeafExprNode : public Node {
   public:
     Expr *E;
 
-    LeafExprNode(Expr *E, OperatorNode *Parent) :
+    LeafExprNode(Expr *E, Node *Parent) :
```
mgrang, line 115: Since the parent is now of type `Node` it means even a `LeafExprNode` can be a parent. But we should not allow parent to be a `LeafExprNode`. Maybe an assert and a check should be added.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -145,34 +153,200 @@ void PreorderAST::Create(Expr *E, OperatorNode *Parent) {
...
+      auto *M = dyn_cast<MemberNode>(N);
+      Coalesce(M->Base, Changed);
+      break;
+    }
+    case Node::NodeKind::ImplicitCastNode: {
+      auto *I = dyn_cast<ImplicitCastNode>(N);
+      Coalesce(I->Child, Changed);
+      break;
+    }
+    default:
```
mgrang, line 296: Having `default` as the first case in a `switch` allows better code readability as the switch may become longer when new cases are added to it in future.

Also do we even need a default case here?

clang/lib/AST/PreorderAST.cpp
```diff
@@ -188,53 +362,94 @@ bool PreorderAST::CompareNodes(const Node *N1, const Node *N2) {
...
+      auto *M = dyn_cast<MemberNode>(N);
+      Sort(M->Base);
+      break;
+    }
+    case Node::NodeKind::ImplicitCastNode: {
+      auto *I = dyn_cast<ImplicitCastNode>(N);
+      Sort(I->Child);
+      break;
+    }
+    default:
```
mgrang, line 408: Same comments on the `default` case here.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -188,53 +362,94 @@ bool PreorderAST::CompareNodes(const Node *N1, const Node *N2) {
...
+      auto *I = dyn_cast<ImplicitCastNode>(N);
+      ConstantFold(I->Child, Changed);
+      break;
+    }
+    default:
+      break;
+  }
+}
+
+void PreorderAST::ConstantFoldOperator(OperatorNode *O, bool &Changed) {
```
mgrang, line 446: We do not need to change the name of this function. It can be called as `ConstantFold` and C++ function overloading would figure out that this function needs to be called based on the type of the first argument (which would be `Node *` in one function and `OperatorNode *` in the other)

mgrang (commented):
As a general observation, I see that now we need to have `switch-cases` in all the functions which act on the PreorderAST like `ConstantFold`, `Coalesce`, etc. Maybe we should move the class-specific implementations of these functions into their respective classes. Then we could simply invoke `Object->ConstantFold()`, `Object->Coalesce()`, etc. This would avoid `switch-cases` and thus result in simpler code.

This can be done in a follow-up PR.

clang/lib/AST/PreorderAST.cpp
kkjeer: The default case avoids a warning that would be emitted if the switch did not handle the `LeafExprNode` case.

clang/lib/AST/PreorderAST.cpp
kkjeer: I tried renaming `ConstantFoldOperator` to `ConstantFold` and unfortunately constant folding no longer worked. It seems that `ConstantFold(BinaryOperatorNode *B)` will only be called on a `Node * N` if `N` is declared as a `BinaryOperatorNode *`. For example, I tried changing the type of `PreorderAST.Root` from `Node *` to `BinaryOperatorNode *` and then `ConstantFold(Root)` called the correct function; however, this does not work for any descendants of a `BinaryOperatorNode`. For example, the expression `*(*(p + 1 + 2) + 3 + 4)` should be constant folded to `*(*(p + 3) + 7)`, but it is not.

kkjeer: Discussed offline - the need to have two different `ConstantFold` and `ConstantFoldOperator` methods should no longer be a consideration after the follow-up work in issue #1032 to move methods such as `ConstantFold` to their respective `Node` classes. For now, the methods `ConstantFold` and `ConstantFoldOperator` can keep their respective names.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -36,56 +36,57 @@ void PreorderAST::AddNode(Node *N, Node *Parent) {
...
+  auto *BParent = dyn_cast_or_null<BinaryOperatorNode>(B->Parent);
+  if (!BParent)
     return;
 
   // Remove the current node from the list of children of its parent.
-  for (auto I = OParent->Children.begin(),
-            E = OParent->Children.end(); I != E; ++I) {
-    if (*I == O) {
-      OParent->Children.erase(I);
+  for (auto I = BParent->Children.begin(),
```
mgrang, line 81: Because `BParent` is mutated in the loop we need to compute its end each time through the loop. Currently we are not doing so because we compute `E = BParent->Children.end()` in the loop initializer. So the correct initializer should be as follows:
`for (auto I = BParent->Children.begin(); I != BParent->Children.end(); ++I)`

Could you also add a comment here indicating that the reason we need to evaluate loop end each time is because the container is mutated in the loop?

For more details see https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop

clang/include/clang/AST/PreorderAST.h
```diff
@@ -23,14 +23,14 @@
 
 namespace clang {
   using Result = Lexicographic::Result;
-  class OperatorNode;
+  class LeafExprNode;
```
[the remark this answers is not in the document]
kkjeer: It's used for the assert in the `Node` constructor if `Parent` is a `LeafExprNode`.

mgrang (approved):
LGTM.

sulekhark (commented):
In addition to the inlined review comments, I have a few general comments:
1. I noticed one issue that seems to have been present before the work on this PR started: the use of `CanCoalesceNode` is sub-optimal. Every call to `CoalesceNode` is preceded by a call to `CanCoalesceNode` but `CanCoalesceNode` is called again as soon as `CoalesceNode` is entered.  The logic of `CanCoalesceNode` is also a bit clunky - I can explain offline in more detail.
2. Most importantly, we need to discuss the following design/implementation issue: It is difficult to go from the canonical form of a subexpression (of a "top-level" expression) to the `AbstractSet` that may record useful information (like aliasing) for the subexpression. While creating a canonical `PreorderAST` for each "top-level" expression, a fresh set of `Nodes` are allocated to construct the canonical forms for subexpressions. We cannot compare the pointer to the subtree (in the preorder AST) corresponding to the subexpression, with the root nodes of the canonical forms of other `AbstractSets`. For example, let `p` be a pointer to an array of structures (i.e. `p` is of type `_Array_ptr<struct S>`). Consider the expressions `*(p+i)` and  `(*(p+i)).f` where `f` is a field in `struct S`. Further suppose that `p+i` is aliased to `q`. Therefore, `*(p+i)` and `*q` will belong to the same `AbstractSet`, say `A`. We may need this information while processing  expression `(*(p+i)).f` - to recognize that `(*(p+i)).f` and `(*q).f` must belong to the same `AbstractSet`. But, under the present design/implementation, it may be very expensive to figure out that the subexpression `*(p+i)` of  `(*(p+i)).f` is actually the same as the canonical form of `AbstractSet A`. We need to discuss this point.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -139,40 +157,210 @@ void PreorderAST::Create(Expr *E, OperatorNode *Parent) {
...
+                                             AE->getExprLoc(), FPOptionsOverride());
+    auto *N = new UnaryOperatorNode(UnaryOperatorKind::UO_Deref, Parent);
+    AddNode(N, Parent);
+    // Even though e1 + e2 is already a binary operator, the child of the
+    // UnaryOperatorNode should be e1 + e2 + 0. This enables expressions such
+    // as p[i + -(1 + 2)] to be constant folded. In order for a
+    // BinaryOperatorNode to be constant folded, it must have at least two
+    // LeafExprNode children whose expressions are integer constants. For
+    // example, i + -(1 + 2) + 0 will be constant folded to i + -3, but
+    // i + -(1 + 2) will not be constant folded.
```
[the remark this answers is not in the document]
kkjeer: Integer constant expressions of the form `+e` and `-e` are `LeafExprNodes` rather than `UnaryOperatorNodes`. The reason for this is so that they can be correctly constant folded. The only children of a `BinaryOperatorNode` that are constant folded are `LeafExprNodes` whose `Exprs` are integer constant expressions. So, for example, the expression `+ p -(1 + 2) 4` will be constant folded to `+ p 1` only if `-(1 + 2)` is a `LeafExprNode` with the `Expr` `-(1 + 2)`. If `-(1 + 2)` is a `UnaryOperatorNode`, this expression will not be constant folded. I don't think this is currently reflected in the design doc - I will update the doc so this is more clear.

The children within a `BinaryOperatorNode` will only be constant folded if the `BinaryOperatorNode` has at least two `LeafExprNode` children whose `Exprs` are integer constant expressions. For example, the `BinaryOperatorNode` + `p -(1 + 2)` will not be constant folded since it only has one integer constant child node, but `+ p -(1 + 2) 0` will be constant folded to `+ p -3`.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -139,40 +157,210 @@ void PreorderAST::Create(Expr *E, OperatorNode *Parent) {
       // to RHS.
     }
 
-    auto *N = new OperatorNode(BinOp, Parent);
+    auto *N = new BinaryOperatorNode(BinOp, Parent);
```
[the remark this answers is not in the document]
kkjeer: The only places where `0` is added to a `BinaryOperatorNode` are where the `AddZero` method is called, currently in the following places:
1. If the `Root` is currently null, so `+ e 0` is the root of the `PreorderAST`
2. To create the `Base` node of an arrow `MemberNode`, so `e->f` is canonicalized to `(e + 0)->f`
3. To create the `Child` node of a dereference `UnaryOperatorNode`, so `*e` is canonicalized to `*(e + 0)`
4. To create the `Child` node of a dereference `UnaryOperatorNode` that is created from an `ArraySubscriptExpr`, so `e1[e2]` is canonicalized to `*(e1 + e2 + 0)`

The intent was for the examples in the design doc to only add `0` to a `BinaryOperatorNode` in these cases. I'll double check the examples in the doc.

sulekhark (approved):
LGTM! Thank you!

mgrang (approved):
LGTM.

Issue #1017, opened 2021-03-30
Improper compiler error when checked variable is out of scope


2000jedi:
# Problem

The following code should compile

```c
int main() {
  int arr _Checked[5];
  int len = 5;
  {
    _Array_ptr<int> dest : count(len) = arr;
  }
  len = 4;
  return 0;
}
```

But instead, the 0.8 (release) version of the compiler generates the following error:

> error: inferred bounds for 'dest' are unknown after statement
> len = 4;

Note that the variable `dest` is already out of scope.

This should not be an error.

sulekhark:
Intersecting the set of in-scope variables across the predecessors of a basic block excludes most variables that are not in-scope for the basic block. However, an exception is when a variable goes out of scope within the basic block. When this happens, the bug noted above manifests.

Fix:
Enabling the AddLifetime flag during CFG construction produces the CFGLifetimeEnds CFGElement for every variable that goes out of scope. When we encounter this CFGElement, we remove the variable that goes out of scope from the ObservedBounds map if it is a checked pointer, and also delete all available expressions that use this out-of-scope variable.

sulekhark:
Fixed by PR #1031.

PR #1031, opened 2021-04-14
Fix for #1017: Compiler issues an invalid-bounds error message for an out-of-scope variable.

sulekhark:
Intersecting the set of in-scope variables across the predecessors of a basic block excludes most variables that are not in-scope for the basic block. However, an exception is when a variable goes out of scope within the basic block. When this happens, the bug in #1017 manifests. 

Fix:
Enabling the `AddLifetime` flag during CFG construction produces the `CFGLifetimeEnds` `CFGElement` for every variable that goes out of scope. When we encounter this `CFGElement`, we remove the variable that goes out of scope from the `ObservedBounds` map if it is a checked pointer, and also delete all available expressions that use this out-of-scope variable.


clang/lib/Sema/BoundsAnalysis.cpp
```diff
@@ -687,13 +687,11 @@ T BoundsAnalysis::Intersect(T &A, T &B) const {
     return Ret;
   }
 
-  for (auto I = Ret.begin(), E = Ret.end(); I != E;) {
+  for (auto I = Ret.begin(); I != Ret.end();) {
```
mgrang, line 690: LLVM coding guidelines recommend not to write iterator-based loops in this form, in general. This is because `Ret.end()` would be evaluated each time through the loop. However, since the container is mutated here we need to write the loop in this form.

Could you please add a comment here indicating that the reason we need to evaluate `Ret.end()` each time is because the container is mutated in the loop?

For more details see https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2726,6 +2726,35 @@ namespace {
...
+     if (V->hasBoundsExpr()) {
+       const AbstractSet *A = AbstractSetMgr.GetOrCreateAbstractSet(V);
+       auto I = State.ObservedBounds.find(A);
+       if (I != State.ObservedBounds.end())
+         State.ObservedBounds.erase(A);
+     }
+
+     EquivExprSets CrntEquivExprs(State.EquivExprs);
+     State.EquivExprs.clear();
+     for (auto I = CrntEquivExprs.begin(); I != CrntEquivExprs.end(); ++I) {
```
mgrang, line 2745: Since `CrntEquivExprs` is not mutated inside the loop, a more efficient way to write this loop is as follows:
`for (auto I = CrntEquivExprs.begin(), E = CrntEquivExprs.end(); I != E; ++I)`

This avoids repeated evaluation of `CrntEquivExprs.end()` each time through the loop.

See https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -2845,6 +2874,14 @@ namespace {
             // variable should be in effect until the very end of traversing S.
             ResetKilledBounds(BA, Block, S, BlockState);
          }
+         else if (Elem.getKind() == CFGElement::LifetimeEnds) {
+            // Every variable going out of scope is indicated by a LifetimeEnds
+            // CFGElement. When a variable goes out of scope, ObservedBounds and
+            // EquivExprs in the CheckingState have to be updated.
+            CFGLifetimeEnds LE = Elem.castAs<CFGLifetimeEnds>();
+            VarDecl *V = const_cast<VarDecl *>(LE.getVarDecl());
```
mgrang, line 2882: Can `LE.getVarDecl()` be null?

clang/lib/Sema/BoundsAnalysis.cpp
sulekhark: Done.

clang/lib/Sema/SemaBounds.cpp
sulekhark: Done.

clang/lib/Sema/SemaBounds.cpp
sulekhark: It shouldn't be null but I have added check to ensure this.

mgrang: If `LE.getVarDecl()` is null `const_cast<VarDecl *>(LE.getVarDecl())` will crash. It may be better to use `dyn_cast_or_null` instead of `const_cast`.

mgrang: > If `LE.getVarDecl()` is null `const_cast<VarDecl *>(LE.getVarDecl())` will crash. It may be better to use `dyn_cast_or_null` instead of `const_cast`.

In an offline discussion Sulekha pointed to this link (https://en.cppreference.com/w/cpp/language/const_cast) which indicates that "a null pointer value would be converted to a null pointer value of `new_type`". So this code should work fine and there is no need to explicitly check if `LE.getVarDecl()` is null before invoking `const_cast`.

mgrang (approved):
LGTM.

kkjeer (approved):
LGTM

dtarditi (commented):
@sulekhark, what is the effect on compile-time of this change?

In the future, we might want to optimize how deletion is done for equivalent expressions, given that the deletion operation is being done every time a variable lifetime ends.  I'm not requesting that we optimize it right now.  I'm curious about the effect on compile-time of this change.

sulekhark:
@dtarditi  Compile time increases by 2.35% . 

Here are the details:
**master**:            compile time summed up over all clang and checkedc test cases (averaged over six runs): 6133 s
**bug_fix_1017**:  compile time summed up over all clang and checkedc test cases (averaged over six runs): 6277 s
                                                                                                                 Overall increase in compile time : 144 s
                                                                                                                                    Percentage increase : 2.35%

dtarditi (approved):
Thanks for measuring the effect on compile-time and opening an issue to track looking at that later.   A 2% compile-time increase is something that we can live with for now, so I approve the change.

PR #1033, opened 2021-04-15
Remove unused code to silence warning

mgrang:
In `SemaBounds.cpp`, we assign to variable `IT` but never use it. This results in
the following compile-time warning:
```
  warning: variable ‘IT’ set but not used [-Wunused-but-set-variable]
  InteropTypeExpr *IT = nullptr;
```


kkjeer (approved):
LGTM. Thanks for fixing this!

sulekhark (approved):
LGTM.

PR #1036, opened 2021-04-19
Add CheckedCPrepass

kkjeer:
This PR implements a prepass that traverses a function and gathers information that is used by Checked C analyses.

Currently, the pass gathers `VarUses`, which maps a `VarDecl` that is used in a bounds declaration or has declared bounds to the `DeclRefExpr` that is the first use of the `VarDecl`. This information is used by bounds checking, which passes it to `AbstractSetManager` in order to get or create an `AbstractSet` for a `VarDecl`.

In the future, this pass can be modified to gather additional information such as variables that are used in bounds, which can be used by the bounds widening analysis.

### DIFF
--- a/clang/include/clang/AST/AbstractSet.h
+++ b/clang/include/clang/AST/AbstractSet.h
@@ -81,10 +81,10 @@ namespace clang {
     // DeclRefExpr for V is used to get or create the AbstractSet for V.
     // Otherwise, a use of V is constructed and added to VarUses. This created
     // use of V is currently not released. It should be rare that a use of V
-    // needs to be created, since this should only if occur if:
-    // 1. V does not occur in any declared bounds expressions and V is unused
-    //    in the body of a function, or:
-    // 2. V does not have a declared bounds expression.
+    // needs to be created, since this should only occur if:
+    // V does not occur in any declared bounds expressions, and:
+    //  1. V is unused within the body of the function, or:
+    //  2. V does not have a declared bounds expression.
     // In order to get or create the AbstractSet for V, any use of V would
     // be sufficient. We choose the first use of V.
     VarUsageTy &VarUses;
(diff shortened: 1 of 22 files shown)


clang/include/clang/AST/AbstractSet.h
```diff
@@ -75,16 +76,18 @@ namespace clang {
...
     Sema &S;
 
-    // VarUses maps a VarDecl with a bounds expression to the DeclRefExpr
-    // (if any) that is the first use of the VarDecl. If a VarDecl V has
-    // an entry in VarUses, the DeclRefExpr for V is used to get or create
-    // the AbstractSet for V. Otherwise, a use of V is constructed and
-    // added to VarUses. This created use of V is currently not released.
-    // It should be rare that a use of V needs to be created, since this
-    // should only occur if V is unused within the body of a function.
+    // VarUses maps a VarDecl to the DeclRefExpr (if any) that is the first
```
mgrang, line 79: that is the first first use of the VarDecl --> that is the first use of the VarDecl

mgrang (commented):
I think the name `CheckedCPrepass` sounds too general as it indicates a compiler pre-pass as opposed to an analyses pre-pass. Maybe `CheckedCAnalysesPrePass` is a more apt name.

sulekhark:
How about `CheckedCSemaPrepass`?

clang/lib/Sema/CheckedCAnalysesPrepass.cpp
```diff
@@ -0,0 +1,61 @@
+#include "clang/Sema/CheckedCAnalysesPrepass.h"
```
mgrang, line 1: This file seems to be missing the header containing the license info.

mgrang (approved):
Thanks for addressing the review comments. LGTM!

PR #1039, opened 2021-04-22
Move Node-specific behavior into Node subclasses

### LINKED ISSUE #1032: Remove switch statements in PreorderAST methods
PR #1030 introduces three new `Node` subclasses to `PreorderAST`: `UnaryOperatorNode`, `MemberNode`, and `ImplicitCastNode`. These node kinds all have class-specific behavior for `PreorderAST` methods such as `Compare`, `CompareNodes`, `Coalesce`, `Sort`, and `ConstantFold`, and these methods currently use `switch` statements based on the kind of the `Node *` argument. These methods should be moved into their respective `Node` classes so that `PreorderAST` can simply call `N->Sort()`, etc. for a `Node *N`.

kkjeer:
Fixes #1032 

This PR refactors PreorderAST to move Coalesce, Sort, ConstantFold, Compare, Cleanup, and PrettyPrint behavior into Node subclass methods.

### DIFF
--- a/clang/include/clang/AST/PreorderAST.h
+++ b/clang/include/clang/AST/PreorderAST.h
@@ -47,9 +47,35 @@ namespace clang {
                  "Parent node cannot be a LeafExprNode");
       }
 
+    // Recursively coalesce BinaryOperatorNodes having the same commutative
+    // and associative operator.
+    // @param[in] this is the current node of the AST.
+    // @param[in] Changed indicates whether a node was coalesced. We need this
+    // to control when to stop recursive coalescing.
+    // @param[in] Error indicates whether an error occurred during coalescing.
     virtual void Coalesce(bool &Changed, bool &Error) { }
+
+    // Recursively descend a Node to sort the children of all
+    // BinaryOperatorNodes if the binary operator is commutative.
+    // @param[in] this is the current node of the AST.
+    // @param[in] Lex is used to lexicographically compare Exprs and Decls
+    // that occur within nodes.
     virtual void Sort(Lexicographic Lex) { }
+
+    // Constant fold integer expressions.
+    // @param[in] this is the current node of the AST.
+    // @param[in] Changed indicates whether constant folding was done. We need
+    // this to control when to stop recursive constant folding.
+    // @param[in] Error indicates whether an error occurred during constant
+    // folding.
+    // @param[in] Ctx is used to create constant expressions.
     virtual void ConstantFold(bool &Changed, bool &Error, ASTContext &Ctx) { }
+
+    // Compare nodes according to their kind.
+    // @param[in] this is the current node of the AST.
+    // @param[in] Other is the node to compare to this.
+    // @return Returns a Lexicographic::Result indicating the comparison
+    // between this and Other according to their node kinds.
     Result CompareKinds(const Node *Other) const {
       if (Kind < Other->Kind)
         return Result::LessThan;
@@ -57,9 +83,20 @@ namespace clang {
         return Result::GreaterThan;
       return Result::Equal;
     }
+
+    // Compare two nodes lexicographically.
+    // @param[in] this is the current node of the AST.
+    // @param[in] Other the node to compare to this.
+    // @param[in] Lex is used to lexicographically compare Exprs and Decls
+    // that occur within nodes.
+    // @return Returns a Lexicographic::Result indicating the comparison
+    // between this and Other.
     virtual Result Compare(const Node *Other, Lexicographic Lex) const {
       return CompareKinds(Other);
     }
+
+    // Cleanup the memory consumed by this node.
+    // @param[in] this is the current node of the AST.
     virtual void Cleanup() {
       delete this;
     }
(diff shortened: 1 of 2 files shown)


clang/include/clang/AST/PreorderAST.h
```diff
@@ -46,6 +46,64 @@ namespace clang {
...
+      return CompareKinds(Other);
+    }
+
+    // Print the node.
+    // @param[in] this is the current node of the AST.
+    virtual void PrettyPrint(llvm::raw_ostream &OS, ASTContext &Ctx) const { }
+
+    // Cleanup the memory consumed by this node.
+    // @param[in] this is the current node of the AST.
+    virtual void Cleanup() {
```
mgrang, line 104: The Node class must have a virtual destructor otherwise when you invoke delete on the derived class object the base class destructor won't be called. I also see a compile-time warning for this:

`PreorderAST.cpp:410:14: warning: deleting object of polymorphic class type ‘clang::Node’ which has non-virtual destructor might cause undefined behavior [-Wdelete-non-virtual-dtor]
  410 |     delete(*I);`

clang/include/clang/AST/PreorderAST.h
```diff
@@ -46,6 +46,64 @@ namespace clang {
...
                  "Parent node cannot be a LeafExprNode");
       }
+
+    // Recursively coalesce BinaryOperatorNodes having the same commutative
+    // and associative operator.
+    // @param[in] this is the current node of the AST.
+    // @param[in] Changed indicates whether a node was coalesced. We need this
+    // to control when to stop recursive coalescing.
+    // @param[in] Error indicates whether an error occurred during coalescing.
+    virtual void Coalesce(bool &Changed, bool &Error) { }
```
mgrang, line 56: It's good to make any function which does not have an implementation in the base class a pure virtual function meaning that its implementation is present only in the derived class.

For example:
`virtual void Coalesce(bool &Changed, bool &Error) = 0;`

clang/lib/AST/PreorderAST.cpp
```diff
@@ -256,8 +256,12 @@ void BinaryOperatorNode::Coalesce(bool &Changed, bool &Error) {
     return;
 
   // Coalesce the children first.
-  for (auto *Child : Children)
+  // Since Children is modified within the loop, we need to evaluate
+  // the loop end on each iteration.
+  for (size_t I = 0; I != Children.size(); ++I) {
+    auto *Child = Children[I];
```
sulekhark, line 262: There is still an incorrectness here - Consider this scenario:
In the third iteration of the loop, `I = 2`. `Child` will be `Children[2]`. Let the `coalesce` function called on `Child` (which is now the same as `Children[2]`) erase `Child`. Because of this erasure, `Children[2]` will now contain what was previously `Children[3]`. Then the control comes out of the `Coalesce`  function and `I` gets incremented to 3. In the next iteration of the loop, `Children[3]` will be processed while `Children[2]` never got processed.

kkjeer: Discussed offline. Summary:

Not every child of a `BinaryOperatorNode` `B` will be processed in a call to `B->Coalesce()`. Any children of `B` that were not processed during a call to `B->Coalesce()` will be processed during future calls to `B->Coalesce()` in the `PreorderAST::Normalize` method, which calls `Root->Coalesce()`, `Root->Sort()`, and `Root->ConstantFold()` in a `while` loop while a `Changed` flag is true.

In the future, we may want to rewrite the `for (size_t I = 0; I != Children.size(); ++I)` loop on line 259 so that all children of `B` are processed in a single call to `B->Coalesce()`. This would be done in separate PR(s).

clang/lib/AST/PreorderAST.cpp
```diff
@@ -278,229 +233,133 @@ void PreorderAST::AddZero(Expr *E, Node *Parent) {
...
-      Coalesce(M->Base, Changed);
-      break;
-    }
-    case Node::NodeKind::ImplicitCastNode: {
-      auto *I = dyn_cast<ImplicitCastNode>(N);
-      Coalesce(I->Child, Changed);
+  // Remove the current node from the list of children of its parent.
+  // Since BParent is modified within the loop, we need to evaluate
+  // the loop end on each iteration.
+  for (auto I = BParent->Children.begin(); I != BParent->Children.end(); ++I) {
```
mgrang, line 278: Actually, as soon as the loop is modified we break out of the loop. So we don't really need to evaluate `BParent->Children.end()` on each iteration. This can be changed to:
`for (auto I = BParent->Children.begin(), E != BParent->Children.end(); I != E; ++I)`

sulekhark (approved):
LGTM!

clang/include/clang/AST/PreorderAST.h
```diff
@@ -93,19 +93,15 @@ namespace clang {
...
     // @param[in] this is the current node of the AST.
-    virtual void PrettyPrint(llvm::raw_ostream &OS, ASTContext &Ctx) const { }
+    virtual void PrettyPrint(llvm::raw_ostream &OS, ASTContext &Ctx) const = 0;
 
     // Cleanup the memory consumed by this node.
     // @param[in] this is the current node of the AST.
-    virtual void Cleanup() {
-      delete this;
-    }
+    virtual void Cleanup() = 0;
```
mgrang, line 104: `Cleanup` should not be a pure virtual function otherwise the base class members like `Parent` won't be deleted. Please restore the previous definition of `Cleanup` in `Node`.

mgrang (approved):
LGTM. Thanks!

PR #1046, opened 2021-05-05
Compare initializer list exprs by address

kkjeer:
For the lvalue generalization work, we need to create AbstractSets for member expressions, which may include initializer lists (e.g. in an assignment such as `((struct S){ 0 }).f = 0;`. The AbstractSetManager maintains a sorted set of PreorderASTs, so PreorderASTs need to be compared. This involves calling `Lexicographic::CompareExpr`. So if an initializer list expression is used to create a PreorderAST, we need to be able to lexicographically compare initializer list expressions.

This PR adds a `CompareAddress` method to CanonBounds.cpp as a fallback method of comparing two expressions which otherwise cannot be easily lexicographically compared. We compare `InitListExprs` and `ImplicitValueInitExprs` using `CompareAddress`.

### DIFF
--- a/clang/lib/AST/CanonBounds.cpp
+++ b/clang/lib/AST/CanonBounds.cpp
@@ -142,6 +142,15 @@ Lexicographic::CompareScope(const DeclContext *DC1, const DeclContext *DC2) cons
 }
 
 Result
+Lexicographic::CompareAddress(const Expr *E1, const Expr *E2) const {
+  if (E1 == E2)
+    return Result::Equal;
+  if (E1 < E2)
+    return Result::LessThan;
+  return Result::GreaterThan;
+}
+
+Result
 Lexicographic::CompareDecl(const CapturedDecl *CD1Arg,
                            const CapturedDecl *CD2Arg) const {
   const CapturedDecl *CD1 = dyn_cast<CapturedDecl>(CD1Arg->getCanonicalDecl());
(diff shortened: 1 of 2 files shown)


clang/lib/AST/CanonBounds.cpp
```diff
@@ -141,6 +141,15 @@ Lexicographic::CompareScope(const DeclContext *DC1, const DeclContext *DC2) cons
   }
 }
 
+Result
+Lexicographic::CompareAddress(const Expr *E1, const Expr *E2) const {
```
mgrang, line 145: I am not sure I fully understand the logic behind comparing addresses. AFAIK, comparing addresses is as non-deterministic (even for one run of the compiler) as randomly return Less/Greater/Equal from this function.

For initializer lists is it possible to compare each member of the list instead of comparing addresses?

sulekhark (approved):
LGTM after an offline discussion on this PR.

clang/lib/AST/CanonBounds.cpp
kkjeer: I believe comparing addresses should be deterministic across a single compiler run (but not across multiple runs).

Two initializer list expressions `A` and `B` should be considered equal if and only if they are the exact same expression as determined by pointer equality. For example, in the following function:
```
void f() {
  ((struct S){ 0 }).f = 0; // A is (struct S){ 0 }
  ((struct S){ 0 }).f = 0; // B is (struct S){ 0 }
}
```
Comparing the initializer list expressions `A` and `B` should not return `Result::Equal`.

We may compare the subexpressions of initializer lists in the future in order to implement a more sophisticated lexicographic ordering `A < B` or `A > B` for initializer lists `A` and `B`. @sulekhark and I discussed offline: there may not be a sufficient need to order `A` and `B` based on their subexpressions. We need a deterministic (at least for one compiler run) ordering of initializer list expressions in order to sort a set of PreorderASTs, but this ordering shouldn't need to compare the subexpressions.

mgrang (approved):
LGTM.

PR #1048, opened 2021-05-11
[LValue Generalization] Synthesize AbstractSets for MemberExprs [8/n]

kkjeer:
When a `MemberExpr *` `M` is modified by an assignment, we need to create an `AbstractSet *` `A` for each member expression whose target bounds use the value of `M`.

For example:
```
struct S {
  int len;
  _Array_ptr<int> p : count(len);
};

void f(struct S *s) {
  s->len = 0;
}
```

When `s->len` is modified by the assignment `s->len = 0`, we create an `AbstractSet` for `s->p`, since the field `p` uses the value of the field `len`.


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -5623,6 +5712,84 @@ namespace {
...
+      if (!Field)
+        return;
+
+      RecordDecl *Parent = Field->getParent();
+      if (!Parent)
+        return;
+
+      // Iterate over each sibling field F of Field to check whether the
+      // target bounds of `Base->F` or `Base.F` use the value of M.
+      for (auto It = Parent->field_begin(), E = Parent->field_end(); It != E; ++It) {
```
kkjeer, line 5776: Once PR #1047 is merged, we can iterate over the fields in `BoundsSiblingFields[Field]` instead of all the fields in `Parent`. `BoundsSiblingFields[Field]` is the set of fields in `Parent` in whose declared bounds `Field` appears. If a field `F` in `Parent` is not in `BoundsSiblingFields[Field]`, then the target bounds of `Base->F` (or `Base.F`) do not use the value of `M`, so we do not need to create a `MemberExpr *` for `Base` and `F`, concretize or expand its declared bounds, or create an `AbstractSet *` for `Base` and `F`.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -5623,6 +5712,84 @@ namespace {
...
+        if (!F->hasBoundsExpr())
+          continue;
+        MemberExpr *BaseF =
+          ExprCreatorUtil::CreateMemberExpr(S, Base, F, ME->isArrow());
+        BoundsExpr *Bounds = MemberExprTargetBounds(BaseF, CSS);
+        int Count = MemberOccurrenceCount(S, M, Bounds);
+        if (Count > 0) {
+          const AbstractSet *A = AbstractSetMgr.GetOrCreateAbstractSet(BaseF);
+          AbstractSets.insert(A);
+        }
```
sulekhark, line 5787: It will be great if we can add two profiling counters here: one for how many times lines 5780 - 5783 executed, and another one for how many times lines 5785 executed. These counters can be dumped after the compilation of the compilation unit.

sulekhark (commented):
Can we please measure the impact of this PR on compilation time?

clang/lib/Sema/SemaBounds.cpp
kkjeer: Are these counters something that should be added to a test file, e.g. synthesized-members.c (in which case it would be more straightforward to print them at the end of checking a function, similar to the existing dump for the synthesized abstract sets), or should they be more similar to the stats that are currently printed at the end of the compilation unit, e.g. in [S.getASTContext().PrintStats())](https://github.com/microsoft/checkedc-clang/blob/master/clang/lib/Parse/ParseAST.cpp#L184)?

sulekhark: My thoughts: Similar to the stats that are currently printed at the end of the compilation unit e.g. in `S.getASTContext().PrintStats()`.

kkjeer:
Impact of this PR on compile time (as measured by LNT benchmarks):

Master compile time: 1595.67s
PR compile time: 1594.87s

Difference: -0.8s (-0.05%)

sulekhark (commented):
Looks very good! I have a last comment:
Could you please add a test case (or, modify an existing one) to make sure that the dump of the profiling counters works as expected?

sulekhark (approved):
LGTM! Thank you!

mgrang (approved):
LGTM.

PR #1049, opened 2021-05-12
Free variables: fix check for integer constant expressions

kkjeer:
When checking for free variables in bounds expressions, a variable `v` should not be considered free if it is equivalent to some integer constant expression.

The previous check `isa<IntegerLiteral>(E)` was not sufficient to determine that an expression `E` equivalent to the value of `v` is an integer constant expression. This check will fail for expressions such `(unsigned int)0`, `1 + 2`, etc. This PR uses `isIntegerConstantExpression(E)` instead.


sulekhark (approved):
LGTM!

Issue #1043, opened 2021-05-03
Clean up constant folding and coalescing methods in PreorderAST


kkjeer:
The following issues should be addressed with `BinaryOperatorNode::ConstantFold` and `BinaryOperatorNode::CanCoalesce`:

**1. Test case**
Another test case that needs to be tested: `*((p & 0xFFFFFFF0) + j + 2)`. Such patterns may occur in real-life code. I see many checks like `IsAssociativeAndCommutative` being applied to binary operators in multiple places - we need to review/test what are the places in `PreorderAST.cpp` where they are really required.

**2. CanCoalesce cleanup**
The function `BinaryOperatorNode::CanCoalesce` needs to be cleaned up.

**3. IsOpCommutativeAndAssociative**
It is not true that we can perform constant folding only if the operator is commutative and associative. Instead we should say the following:

Performing coalescing for the `+` and the `*` operators (which are commutative and associative) creates more opportunities for performing constant folding on the operands of the coalesced operator. Therefore, we explicitly perform constant folding for the operands of the (coalesced) `+` and `*` operators while piggy-backing on `isIntegerConstantExpr` to perform constant folding in all other cases.

**4. Idx increment placement**
Can this increment move to the bottom of the while loop?
Reason for this comment: The increment/decrement of the loop control variable must happen uniformly either at the beginning of the body of the loop or at the end. In lines 371 thro' 377 above, the increment of `Idx` is happening after `Children[<current value of Idx>]` is processed. Therefore, even in the else case (when the condition on line 370 does not hold true, the increment of `Idx` must happen after `Children[<current value of Idx>]` is processed.

Note on this issue:
The increment was placed here due to the `continue` on lines 365, 374, and 388. If any of these cases occur and `Idx` is not incremented, it results in an infinite loop.

**5. Minimum number of constants needed to constant fold**
It is not true that we need at least two constants to fold constants - a single constant can be trivially "folded". The presence of this check may cause redundant calls to `AddZero` just to ensure that constant folding happens - this issue had come up in one of the earlier offline review discussions.

**6. Creating new LeafExprNodes to constant fold**
Here, it is safer to call `PreorderAST::CreateUnaryOperator(ConstFoldedExpr, this)` instead of doing `new LeafExprNode(ConstFoldedExpr, this)`.
Note: "here" refers to `Children.push_back(new LeafExprNode(ConstFoldedExpr, this));` in `BinaryOperatorNode::ConstantFold`.

PR #1053, opened 2021-05-18
PreorderAST: fold one constant

kkjeer:
This PR addresses the following point from issue #1043:

**5. Minimum number of constants needed to constant fold**
It is not true that we need at least two constants to fold constants - a single constant can be trivially "folded". The presence of this check may cause redundant calls to AddZero just to ensure that constant folding happens - this issue had come up in one of the earlier offline review discussions.


mgrang (approved):
LGTM.

sulekhark (approved):
LGTM! Thanks!

PR #1056, opened 2021-05-20
[BoundsVars] Store separate lower and upper bounds variables

mgrang:
We separate out BoundsVars into BoundsVarsLower and BoundsVarsUpper that track
the variables occurring in the lower and upper bounds expressions,
respectively. This is needed for the bounds widenening analysis where we want
to determine the variables that can potentially be widened in a given
dereference expression.

### DIFF
--- a/clang/lib/Sema/CheckedCAnalysesPrepass.cpp
+++ b/clang/lib/Sema/CheckedCAnalysesPrepass.cpp
@@ -82,17 +82,17 @@ class PrepassHelper : public RecursiveASTVisitor<PrepassHelper> {
       return T->getAsRecordDecl();
     }
 
-    // AddPtrWithBounds adds PtrWithBounds to the set of variables in whose
-    // bounds Var occurs.
-    template<class M, class P, class V>
-    void AddPtrWithBounds(M &Map, const P *PtrWithBounds, const V *Var) {
-      auto It = Map.find(Var);
+    // AddPtrWithBounds adds PtrWithBounds to the set of variables or fields in
+    // whose bounds VarOrField occurs.
+    template<class T, class U>
+    void AddPtrWithBounds(T &Map, const U *PtrWithBounds, const U *VarOrField) {
+      auto It = Map.find(VarOrField);
       if (It != Map.end())
         It->second.insert(PtrWithBounds);
       else {
-        typename M::mapped_type Ptrs;
+        typename T::mapped_type Ptrs;
         Ptrs.insert(PtrWithBounds);
-        Map[Var] = Ptrs;
+        Map[VarOrField] = Ptrs;
       }
     }
 
--- a/clang/lib/Sema/CheckedCAnalysesPrepass.cpp
+++ b/clang/lib/Sema/CheckedCAnalysesPrepass.cpp
@@ -50,6 +50,18 @@ class PrepassHelper : public RecursiveASTVisitor<PrepassHelper> {
     // };
     FieldDecl *FieldWithBounds = nullptr;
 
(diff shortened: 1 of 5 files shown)


mgrang:
Note: After review and approval, this PR is meant to be merged into the `bounds_widening_wip` from the `bounds_widening_dev` branch which is my development branch.
We will merge all bounds widening related PRs into the `bounds_widening_wip` branch. Once all PRs in this branch are approved and tested on ADO we will merge this branch into the master.

clang/lib/Sema/CheckedCAnalysesPrepass.cpp
```diff
@@ -70,16 +82,17 @@ class PrepassHelper : public RecursiveASTVisitor<PrepassHelper> {
       return T->getAsRecordDecl();
     }
 
-    // AddBoundsSiblingField adds FieldWithBounds to the set of fields in
-    // whose declared bounds F occurs.
-    void AddBoundsSiblingField(const FieldDecl *F) {
-      auto It = Info.BoundsSiblingFields.find(F);
-      if (It != Info.BoundsSiblingFields.end())
-        It->second.insert(FieldWithBounds);
+    // AddPtrWithBounds adds PtrWithBounds to the set of variables in whose
```
kkjeer, line 85: Nit: could this comment say something like "...to the set of variables or fields" or "...to the set of decls" since this method can also be used to update `BoundsSiblingFields`?

clang/lib/Sema/CheckedCAnalysesPrepass.cpp
```diff
@@ -70,16 +82,17 @@ class PrepassHelper : public RecursiveASTVisitor<PrepassHelper> {
...
+    // bounds Var occurs.
+    template<class M, class P, class V>
+    void AddPtrWithBounds(M &Map, const P *PtrWithBounds, const V *Var) {
```
kkjeer, line 88: Similarly, could `Var` be renamed since this argument isn't necessarily a variable?

kkjeer, line 88: Do any calls to `AddPtrWithBounds` use different types for `PtrWithBounds` and `Var`?

clang/lib/Sema/CheckedCAnalysesPrepass.cpp
mgrang: Yes, I though about adding "variables or fields" to the comment. But currently it updates variables and fields but in future it may be invoked on something else. So I wanted didn't want to keep extending the comment. That's the reason I simply went with variables (because I think it is general enough).

clang/lib/Sema/CheckedCAnalysesPrepass.cpp
mgrang: Same reply as above. And maybe I can add a note saying "Var may either be a variable or a field".

mgrang: I am not sure I understood your question. All calls to `AddPtrWithBounds` use different types for `PtrWithBounds` and `Var`.

kkjeer: I thought that, for any given call to `AddPtrWithBounds`, `PtrWithBounds` and `Var` would either both be of type `VarDecl *` or both be of type `FieldDecl *`. My question is, do we need separate template type arguments `class P` and `class V` for `AddPtrWithBounds`?

sulekhark (approved):
LGTM! Thanks!

PR #1057, opened 2021-05-24
[SemaBounds] Add a wrapper around BaseRange::InRange

mgrang:
Added a wrapper around CheckBoundsDeclaration::BaseRange::InRange. This
provides an easy way to invoke this function from outside the class. Given
two bounds expressions B1 and B2 it determines if B2 is a subrange of B1.
This function would be used by the bounds widening analysis.


sulekhark (approved):
LGTM!

PR #1058, opened 2021-05-24
[LValue Generalization] Update bounds context for member expressions [9/n]

kkjeer:
This PR updates the `ObservedBounds` context for the synthesized `AbstractSets` for member expressions after an assignment.

This PR depends on several smaller changes. This PR will be a draft until those smaller changes are merged in separate, independent PRs.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -4625,10 +4625,11 @@ namespace {
         std::pair<BoundsExpr *, Expr *> Lost = LostLValueIt->second;
         BoundsExpr *InitialObservedBounds = Lost.first;
         Expr *LostLValue = Lost.second;
-        // TODO: currently, we only add a pair <B, E> to LostLValues if
-        // E is a DeclRefExpr, so we can still emit the note_lost_variable
-        // diagnostic message. In the future, we should change this message
-        // so that it applies to all kinds of lvalue expressions.
+        // TODO: currently, we only validate bounds for an AbstractSet A in
+        // ObservedBounds if the representative of A is a DeclRefExpr, so
+        // we can still emit the note_lost_variable diagnostic message.
+        // In the future, we should change this message so that it applies
+        // to all kinds of lvalue expressions.
         S.Diag(LostLValue->getBeginLoc(), diag::note_lost_variable)
           << LostLValue << InitialObservedBounds << V << LostLValue->getSourceRange();
       }
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -721,7 +721,7 @@ namespace {
   Expr *ReplaceLValue(Sema &SemaRef, Expr *E, Expr *LValue,
                       Expr *OriginalValue,
                       CheckedScopeSpecifier CSS) {
-    // Don't transform E if does not use the value of LValue.
+    // Don't transform E if it does not use the value of LValue.
     if (!FindLValue(SemaRef, LValue, E))
       return E;
 
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -532,25 +532,46 @@ namespace {
 }
 
 namespace {
-  class FindMemberHelper : public RecursiveASTVisitor<FindMemberHelper> {
+  class FindLValueHelper : public RecursiveASTVisitor<FindLValueHelper> {
     private:
       Sema &SemaRef;
       Lexicographic Lex;
-      MemberExpr *M;
+      Expr *LValue;
       bool Found;
 
     public:
-      FindMemberHelper(Sema &SemaRef, MemberExpr *M) :
+      FindLValueHelper(Sema &SemaRef, Expr *LValue) :
         SemaRef(SemaRef),
         Lex(Lexicographic(SemaRef.Context, nullptr)),
-        M(M),
+        LValue(LValue),
         Found(false) {}
 
       bool IsFound() { return Found; }
 
+      bool VisitDeclRefExpr(DeclRefExpr *E) {
+        DeclRefExpr *V = dyn_cast_or_null<DeclRefExpr>(LValue);
+        if (!V)
+          return true;
+        if (Lex.CompareExpr(V, E) == Lexicographic::Result::Equal)
+          Found = true;
+        return true;
+      }
+
       bool VisitMemberExpr(MemberExpr *E) {
+        MemberExpr *M = dyn_cast_or_null<MemberExpr>(LValue);
+        if (!M)
+          return true;
         if (Lex.CompareExprSemantically(E, M))
           Found = true;
+        return true;
+      }
+
+      // Do not traverse the child of a BoundsValueExpr.
+      // Expressions within a BoundsValueExpr should not be considered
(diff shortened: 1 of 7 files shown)


sulekhark (commented):
At this point in the implementation of this feature , it will be great if one or two test cases that are complete meaningful functions representing realistic scenarios that users of Checked C may encounter, are added. This will give us an idea of what additional support we need to provide in order to make this feature usable by programmers.
One thing (as is recorded in a TODO in the code) is to support inverting operations on member expressions whenever an operation is invertible.

clang/test/CheckedC/inferred-bounds/bounds-context-members.c
```diff
@@ -0,0 +1,282 @@
...
+  // CHECK-NEXT:     DeclRefExpr {{.*}} 's'
+  // CHECK-NEXT: Bounds:
+  // CHECK-NEXT: NullaryBoundsExpr {{.*}} Any
+  // CHECK-NEXT: }
+}
+
+// Set the inferred bounds of the LHS to the inferred bounds of the RHS
+void updated_source_bounds2(struct S *s) {
+  // Observed bounds context after assignment: { s[4].q => bounds(s[4].r, s[4].r + s[4].i) }
+  s[4].q = s[4].r;
```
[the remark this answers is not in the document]
kkjeer: The concrete, expanded bounds of `s[3].r` are `bounds(s[3].r, s[3].r + s[3].i)`, so the observed bounds context will contain `{ s[4].q => bounds(s[3].r, s[3].r + s[3].i) }` after the assignment `s[4].q = s[3].r`. I've added this test case.

sulekhark (approved):
LGTM! Thank you!

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -532,39 +532,60 @@ namespace {
...
+        LValue(LValue),
         Found(false) {}
 
       bool IsFound() { return Found; }
 
+      bool VisitDeclRefExpr(DeclRefExpr *E) {
+        DeclRefExpr *V = dyn_cast_or_null<DeclRefExpr>(LValue);
+        if (!V)
+          return true;
+        if (Lex.CompareExpr(V, E) == Lexicographic::Result::Equal)
```
mgrang, line 555: I am assuming we don't need to call `CompareExprSemantically` here?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -622,26 +643,44 @@ namespace {
 }
 
 namespace {
-  class ReplaceVariableHelper : public TreeTransform<ReplaceVariableHelper> {
-    typedef TreeTransform<ReplaceVariableHelper> BaseTransform;
+  class ReplaceLValueHelper : public TreeTransform<ReplaceLValueHelper> {
+    typedef TreeTransform<ReplaceLValueHelper> BaseTransform;
```
mgrang, line 647: While you are at it you can consider replacing `typedef` with `using`. However, this can be done in a separate PR and won't hold up this PR.

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -675,22 +714,27 @@ namespace {
...
-    // Don't transform e if it does not use the value of v.
-    if (!VariableOccurrenceCount(SemaRef, V, E))
+  // If an original value is provided, ReplaceLValue returns an expression
+  // that replaces all uses of the lvalue expression LValue in E with the
+  // original value.  If no original value is provided and E uses LValue,
+  // ReplaceLValue returns nullptr.
+  Expr *ReplaceLValue(Sema &SemaRef, Expr *E, Expr *LValue,
+                      Expr *OriginalValue,
+                      CheckedScopeSpecifier CSS) {
+    // Don't transform E if does not use the value of LValue.
```
mgrang, line 724: `Don't transform E if does not` --> `Don't transform E if it does not`

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -675,22 +714,27 @@ namespace {
...
+    if (!FindLValue(SemaRef, LValue, E))
       return E;
 
+    // If E uses the value of LValue, but no original value is provided,
+    // we know the result is null without needing to transform E.
+    if (!OriginalValue)
```
mgrang, line 730: Can the check for `OriginalValue` move before the `FindLValue` check because this check is much cheaper than `FindLValue`?

clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4713,168 +4760,248 @@ namespace {
...
+    // occurrences of LValue replaced with OriginalValue.
+    // If OriginalValue is null, each expression in EquivExprs that uses the
+    // value of LValue is removed from EquivExprs.
+    void UpdateEquivExprsAfterAssignment(Expr *LValue, Expr *OriginalValue,
+                                         CheckedScopeSpecifier CSS,
+                                         CheckingState &State) {
+      const EquivExprSets PrevEquivExprs = State.EquivExprs;
       State.EquivExprs.clear();
-      for (auto I = PrevState.EquivExprs.begin(); I != PrevState.EquivExprs.end(); ++I) {
+      for (auto I = PrevEquivExprs.begin(); I != PrevEquivExprs.end(); ++I) {
```
mgrang, line 4934: Since `PrevEquivExprs` is not modified inside the loop we don't need to re-evaluate `PrevEquivExprs.end()` on each loop iteration. For better readability, this loop can be written in a range-based format instead of using iterators. Same for the `InnerList` loop below.

Note: This is only a suggestion and does not need to be done in the current PR.

clang/lib/Sema/SemaBounds.cpp
kkjeer: We need to know whether `E` uses the value of `LValue` regardless of whether `OriginalValue` is null. If `OriginalValue` is null and `FindLValue` returns true, we can return null without transforming `E`. If `OriginalValue` is non-null, we still need to call `FindLValue` so we can determine whether we can just return `E` without calling `ReplaceLValue`.

clang/lib/Sema/SemaBounds.cpp
kkjeer: I don't believe we need to call `CompareExprSemantically` in order to compare two `DeclRefExprs`. We shouldn't need to create `PreorderASTs` for `DeclRefExprs` to compare them since the comparison will just involve unnecessary sorting of the `PreoderAST` nodes `LeafExprNode(0)` and `LeafExprNode(v)` for a `DeclRefExpr` `v`.

clang/lib/Sema/SemaBounds.cpp
kkjeer: This `typedef` is consistent with other `TreeTransform` classes in SemaBounds.cpp. We can replace all `typedef` with `using` in a separate PR.

kkjeer: Created issue #1075 to track this.

clang/lib/Sema/SemaBounds.cpp
kkjeer: There are several instances of this in SemaBounds.cpp. Created issue #1077 to track this cleanup work item.

mgrang (approved):
LGTM.

PR #1061, opened 2021-05-24
PreorderAST: ensure that ConstFoldedVal and CurrConstVal have the same bit width

kkjeer:
This PR fixes an assert that would happen in `APInt::+=` in `sadd_ov` and `smul_ov` when arguments had different bit widths. When we expand the set of expressions that we canonicalize (create PreorderASTs for), we can encounter binary operators with subexpressions with different bit widths, so we can hit this assert unless we ensure that the arguments have the same bit width.

This is one of several small changes to support #1058.

### DIFF
--- a/clang/lib/AST/PreorderAST.cpp
+++ b/clang/lib/AST/PreorderAST.cpp
@@ -407,7 +407,7 @@ bool BinaryOperatorNode::ConstantFold(bool &Error, ASTContext &Ctx) {
       // Ensure that ConstFoldedVal and CurrConstVal have the same bit width.
       if (ConstFoldedVal.getBitWidth() < CurrConstVal.getBitWidth())
         ConstFoldedVal = ConstFoldedVal.extOrTrunc(CurrConstVal.getBitWidth());
-      if (CurrConstVal.getBitWidth() < ConstFoldedVal.getBitWidth())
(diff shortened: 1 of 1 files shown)


sulekhark (approved):
LGTM! Thanks!

clang/lib/AST/PreorderAST.cpp
```diff
@@ -404,6 +404,12 @@ bool BinaryOperatorNode::ConstantFold(bool &Error, ASTContext &Ctx) {
       ConstFoldedVal = CurrConstVal;
 
     } else {
+      // Ensure that ConstFoldedVal and CurrConstVal have the same bit width.
+      if (ConstFoldedVal.getBitWidth() < CurrConstVal.getBitWidth())
+        ConstFoldedVal = ConstFoldedVal.extOrTrunc(CurrConstVal.getBitWidth());
+      if (CurrConstVal.getBitWidth() < ConstFoldedVal.getBitWidth())
```
mgrang, line 410: This should be an `else if` because either `ConstFoldedVal.getBitWidth() < CurrConstVal.getBitWidth()` or `CurrConstVal.getBitWidth() < ConstFoldedVal.getBitWidth()`, but not both.

mgrang (approved):
LGTM. Thanks.

PR #1062, opened 2021-05-24
Return nullptr from ExpandToRange and GetOriginalValue if arguments are null

kkjeer:
This PR is one of several small changes to support #1058.


sulekhark (approved):
LGTM! Thanks!

mgrang (approved):
LGTM.

Issue #1078, opened 2021-06-01
Remove PreorderAST::GetDerefOffset after the updated implementation of the bounds widening analysis merges.


mgrang:
`PreorderAST::GetDerefOffset` should be removed after the updated implementation of the bounds widening analysis merges. This method will be replaced by `PreorderAST::GetExprIntDiff`.

PR #1079, opened 2021-06-01
[PreorderAST] Add method to compute integer diff between two expressions

mgrang:
Added a method called PreorderAST::GetExprIntDiff that computes the integer
difference between two input expressions and returns true if the expressions
are comparable. If the expressions are not comparable then it returns false.

This method will be used in the updated bounds widening analysis. This method
is meant to replace PreorderAST::GetDerefOffset after the updated bounds
widening analysis merges.

See https://github.com/microsoft/checkedc-clang/issues/1078

### DIFF
--- a/clang/lib/AST/PreorderAST.cpp
+++ b/clang/lib/AST/PreorderAST.cpp
@@ -514,6 +514,9 @@ bool PreorderAST::GetDerefOffset(Node *UpperNode, Node *DerefNode,
   if (B1->Children.size() != B2->Children.size())
     return false;
 
+  llvm::APSInt Zero(Ctx.getTargetInfo().getIntWidth(), 0);
+  Offset = Zero;
+
   // Check if the children are equivalent.
   for (size_t I = 0; I != B1->Children.size(); ++I) {
     auto *Child1 = B1->Children[I];
@@ -549,7 +552,6 @@ bool PreorderAST::GetDerefOffset(Node *UpperNode, Node *DerefNode,
     // This guards us from a case where the constants were not folded for
     // some reason. In theory this should never happen. But we are adding this
     // check just in case.
-    llvm::APSInt Zero(Ctx.getTargetInfo().getIntWidth(), 0);
     if (llvm::APSInt::compareValues(Offset, Zero) != 0)
       return false;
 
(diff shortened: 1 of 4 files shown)


kkjeer (approved):
LGTM. One minor comment on a typo.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -560,6 +564,81 @@ bool PreorderAST::GetDerefOffset(Node *UpperNode, Node *DerefNode,
...
+    llvm::APSInt IntegerPart2;
+    if (!L2->E->isIntegerConstantExpr(IntegerPart2, Ctx))
+      return false;
+
+    // Offset should always be of the form (ptr + offset). So we check for
+    // addition.
+    // Note: We have already converted (ptr - offset) to (ptr + -offset). So
+    // it is okay to only check for addition.
+    if (B1->Opc != BO_Add)
+      return false;
```
sulekhark, line 622: This check could perhaps happen earlier in the function before iterating over the children nodes.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -560,6 +564,81 @@ bool PreorderAST::GetDerefOffset(Node *UpperNode, Node *DerefNode,
...
+
+    llvm::APSInt IntegerPart2;
+    if (!L2->E->isIntegerConstantExpr(IntegerPart2, Ctx))
+      return false;
+
+    // This guards us from a case where the constants were not folded for
+    // some reason. In theory this should never happen. But we are adding this
+    // check just in case.
+    llvm::APSInt Zero(Ctx.getTargetInfo().getIntWidth(), 0);
```
sulekhark, line 627: The creation of the `Zero` object can also move outside the loop.

clang/lib/AST/PreorderAST.cpp
```diff
@@ -560,6 +564,81 @@ bool PreorderAST::GetDerefOffset(Node *UpperNode, Node *DerefNode,
...
+  if (B1->Children.size() != B2->Children.size())
+      return false;
+
+  // Check if the children are equivalent.
+  for (size_t I = 0; I != B1->Children.size(); ++I) {
+    auto *Child1 = B1->Children[I];
+    auto *Child2 = B2->Children[I];
+
+    if (Child1->Compare(Child2, Lex) == Result::Equal)
+      continue;
```
sulekhark, line 605: The output parameter `Offset` is not assigned any valid value in the case when `E1` and `E2` are exactly equal, and the loop that iterates through the children nodes "continues" (at line 605) in each iteration.

sulekhark (commented):
Sorry, I realized that I missed one or two minor things during my previous review.

sulekhark (approved):
LGTM, thank you!

Issue #1085, opened 2021-06-07
A function parameter having a Where clause causes the compiler to issue an incorrect error message at a callsite


sulekhark:
The test case below causes the compiler to issue an incorrect error message:
```
#pragma CHECKED_SCOPE on

void iface(unsigned long n _Where n > 0) {
    return;
}

void iface_test1(void) {
    char buf _Nt_checked[50];
    iface(50);
}
```

Compilation error: 
>  ../../build/bin/clang -c ext11.c
ext11.c:9:9: error: argument has unknown bounds, bounds expected because the 1st parameter has bounds
  iface(50);
        ^~
1 error generated.

PR #1086, opened 2021-06-07
Fix for issue #1085

sulekhark:
The fix is to correct a slip in the function `StartsWhereClause`.

Because of the bug in this function, the code in `ParseParameterDeclarationClause` associated a `BoundsAnnotations` object with the first parameter of the function `iface` in the test case of issue #1085. This `BoundsAnnotations` object set the `Bounds` field to a (non-null) `NullaryBoundsExpr` and the `InteropType` field to null. The non-null `Bounds` field in the `BoundsAnnotations` associated with the first parameter caused the `CheckCallExpr` function to expect the first argument to function `iface` to have bounds.


mgrang (approved):
LGTM. Thanks for fixing this!

kkjeer (approved):
LGTM!

PR #1087, opened 2021-06-08
CreateStandardForm: check for multiplication operators

kkjeer:
This PR addresses an issue with the `CreateStandardForm` method.

Consider an expression of the form `p + (v OP c)`, where:
* `p` is a pointer to a `char`
* `v` is an integer-typed variable expression
* `c` is an integer constant

The out parameter `ConstantPart` should be set to `c` if and only if `OP` is multiplication. Otherwise, `ConstantPart` should be set to 1. For example, for the expression `p + v * sizeof(int)`, `ConstantPart` should be set to `sizeof(int)`. For `p + v + 2`, `ConstantPart` should not be set to 1, not 2.

This change fixes an issue that would occur with bounds checking for `char` pointers: with the previous behavior, `bounds(p, p + i - 1` imply `bounds(p, p + i)` for an `_Array_ptr<char>` or `_Nt_array_ptr<char>` `p`. `CreateStandardForm` set `VariablePart` to `i` and `ConstantPart` to 1 for both `p + i` and `p + i - 1`. With this change, `CreateStandardForm` sets `ConstantPart` to 1 for both `p + i` and `p + i - 1`, but it sets `VariablePart` to `i` and `i - 1` for `p + i` and `p + i - 1`, respectively.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -1951,6 +1951,8 @@ namespace {
         return false;
       if (Base->getType()->getPointeeOrArrayElementType()->isCharType()) {
         if (BinaryOperator *BO = dyn_cast<BinaryOperator>(Offset->IgnoreParens())) {
+          // Check to see if Offset is already of the form e * i or i * e. In this case,
+          // ConstantPart = i, VariablePart = e.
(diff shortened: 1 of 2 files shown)


sulekhark (approved):
LGTM! Thank you!

dtarditi:
@kkjeer, thanks for fixing this bug.  The changes look good to me as well and make the function consistent with the original intent.

PR #1093, opened 2021-06-15
[BoundsWidening] Implement the bounds widening dataflow analysis

mgrang:
This is an updated implementation of the dataflow analysis for bounds widening.
In addition to the existing support for bounds widening on a conditional
dereferencing a null-terminated array at its upper bound this implementation
also supports bounds widening in presence of _Where clauses. For example:
```
  _Nt_array_ptr<char> p : bounds(p, p + 1);
  int x = strlen(p) _Where p : bounds(p, p + x); // Bounds of p widened to bounds(p, p + x).
```

### DIFF
--- a/clang/include/clang/Sema/BoundsWideningAnalysis.h
+++ b/clang/include/clang/Sema/BoundsWideningAnalysis.h
@@ -314,9 +314,7 @@ namespace clang {
 
     // Run the dataflow analysis to widen bounds for null-terminated arrays.
     // @param[in] FD is the current function.
-    // @param[in] NestedStmts is a set of top-level statements that are
-    // nested in another top-level statement.
-    void WidenBounds(FunctionDecl *FD, StmtSetTy NestedStmts);
+    void WidenBounds(FunctionDecl *FD);
 
     // Pretty print the widened bounds for all null-terminated arrays in the
     // current function.
--- a/clang/include/clang/Sema/BoundsWideningAnalysis.h
+++ b/clang/include/clang/Sema/BoundsWideningAnalysis.h
@@ -321,6 +321,22 @@ namespace clang {
     // @param[in] FD is the current function.
     void DumpWidenedBounds(FunctionDecl *FD);
 
+    // Get the Out set for the statement. This set represents the bounds
+    // widened after the statement.
+    // Note: This method can be called from outside this class to get the
+    // widened bounds after a statement.
+    // @param[in] B is the current block.
+    // @param[in] CurrStmt is the current statement.
+    BoundsMapTy GetStmtOut(const CFGBlock *B, const Stmt *CurrStmt) const;
+
+    // Get the In set for the statement. This set represents the bounds
+    // widened before the statement.
+    // Note: This method can be called from outside this class to get the
+    // widened bounds before a statement.
+    // @param[in] B is the current block.
+    // @param[in] CurrStmt is the current statement.
+    BoundsMapTy GetStmtIn(const CFGBlock *B, const Stmt *CurrStmt) const;
+
   private:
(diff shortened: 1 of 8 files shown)


clang/include/clang/Sema/BoundsWideningAnalysis.h
```diff
@@ -15,55 +15,442 @@
...
+    // @param[in] EB is the current ElevatedCFGBlock.
+    // @param[out] EB is added to WorkList if the Out set of EB changes.
+    void ComputeOutSet(ElevatedCFGBlock *EB, WorkListTy &WorkList);
+
+    // Initialize the In and Out sets for the block.
+    // @param[in] FD is the current function.
+    // @param[in] EB is the current ElevatedCFGBlock.
+    void InitBlockInOutSets(FunctionDecl *FD, ElevatedCFGBlock *EB);
+
+    // Prune the Out set of the pred block according to various conditions.
```
sulekhark, line 384: Here, it will be great to have an explanation about why `PruneOutSet` is required because creating this function is an implementation decision. As I understand, the implementation of  the intersection operation as defined in the `BoundsWideningAnalysis` is done in two steps:
1) the function `PruneOutSet` pre-processes the `BoundsMaps` that we wish to intersect, and
2) the regular intersection operation is performed.

clang/lib/Sema/BoundsWideningAnalysis.cpp
```diff
@@ -14,4 +14,1213 @@
...
+  }
+
+  // Compute the In and Out sets for blocks. This is the fixedpoint computation
+  // for the dataflow analysis.
+  while (!WorkList.empty()) {
+    ElevatedCFGBlock *EB = WorkList.next();
+    WorkList.remove(EB);
+
+    ComputeInSet(EB);
+    ComputeOutSet(EB, WorkList);
```
sulekhark, line 69: If possible, the code for adding the successors of `EB` into the `WorkList` can be moved here to make the fixpoint computation clearer. For example, `ComputeOutSet` can return a list (or set) of successor basic blocks that need to be added to `WorkList` (this list can be empty if the out set is unchanged).
If this is hard, then there should be a clear comment here explaining this.

sulekhark (commented):
I have added two minor review comments. In addition, it will be great if we can discuss the following in the planned review meeting:
1) What is the plan for the functions `IsSubRange`, `GetDerefExpr`, and `AddOffsetToExpr` going forward? I think that the implementation of these functionalities must not be the responsibility of the `BoundsWideningAnalysis`.
2) Review of the replacement for `VarsToWiden`.
3) Review of `PruneOutSets`
4) Review/Discussion of test cases that require the computation of the In/Out dataflow sets for each basic block more than once (i.e. it takes more than one pass over the basic blocks to reach fixpoint).
5) Review of the support for richer termination conditions.

clang/lib/Sema/BoundsWideningAnalysis.cpp
```diff
@@ -14,4 +14,1213 @@
...
+  // In set calculation.
+  EB->TermCondDerefExpr = DerefExpr;
+
+  // Get all variables in the expression that are pointers to null-terminated
+  // arrays. For example: On a dereference expression like "*(p + i + j + 1)"
+  // GetNtPtrsInExpr() will return {p} if p is a pointer to a null-terminated
+  // array.
+
+  VarSetTy NtPtrsInExpr;
+  BWUtil.GetNtPtrsInExpr(DerefExpr, NtPtrsInExpr);
```
mgrang, line 603: This function should return a single NtPtr because a deref expr can only contain at most one null-terminated pointer.

mgrang:
Superseded by https://github.com/microsoft/checkedc-clang/pull/1122

PR #1095, opened 2021-06-17
Simple constant folding for bounds upper offsets

kkjeer:
This PR introduces a temporary solution for an issue with bounds validation: the bounds checker should be able to prove that `bounds(p, p + ((len + 1) - 1))` imply `bounds(p, p + len)`.

The solution involves performing constant folding for a limited set of expressions. If the upper offset of a range is of the form `(e +/- a) +/ b`, where `a` and `b` are integer constants and `e` is an expression, we extract the variable part as `e` and the constant part as `a + b` (or `a + -b`, `-a + b`, or `-a + -b`).

This is intended to be a temporary fix until we can solve issues with constant folding, commutativity, and associativity in bounds expressions in a more general way.

This fix addresses +1/-1 situations that can arise from passing arguments to functions with declared parameter bounds (see [checkedc/450](https://github.com/microsoft/checkedc/issues/450)). However, it does not fix +1/-1 situations that can occur as a result of bounds widening, e.g.

```
  if (*(p + len)) {
    ++len;
  }
}
```

The bounds checker is unable to create a base range for the inferred bounds `bounds(p, (p + (len - 1)) + 1)`. We need further work to address associativity issues with bounds expressions.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -1366,6 +1366,100 @@ namespace {
         LowerOffsetVariable(LowerOffsetVariable), UpperOffsetVariable(UpperOffsetVariable) {
       }
 
+    private:
+      // If E is of the form e +/- c, where c is a constant, GetRHSConstant
+      // returns true and sets the out parameter Constant.
+      // If E is of the form e + c, Constant will be set to c.
+      // If E is of the form e - c, Constant will be set to -c.
+      bool GetRHSConstant(BinaryOperator *E, llvm::APSInt &Constant) {
+        if (!E->isAdditiveOp())
+          return false;
+        if (!E->getRHS()->isIntegerConstantExpr(Constant, S.Context))
+          return false;
+
+        bool Overflow;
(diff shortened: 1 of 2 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -1366,6 +1366,177 @@ namespace {
...
+        // Normalize the operation by negating the offset if necessary.
+        if (E->getOpcode() == BO_Sub) {
+          uint64_t PointerWidth = S.Context.getTargetInfo().getPointerWidth(0);
+          Constant = llvm::APSInt(PointerWidth, false).ssub_ov(Constant, Overflow);
+          if (Overflow)
+            return false;
+        }
+        llvm::APSInt ElemSize;
+        if (!BoundsUtil::getReferentSizeInChars(S.Context, Base->getType(), ElemSize))
+          return false;;
```
mgrang, line 1413: Nit: Extra semicolon.

sulekhark (approved):
LGTM! Thank you!

mgrang (approved):
LGTM.

PR #1098, opened 2021-06-17
Add null statements to the CFG

sulekhark:
`_Where` clauses, which are processed in a dataflow analysis like `BoundsWideningAnalysis` (and will soon be processed in `AvailableFactsAnalysis`) could be attached to null statements. Therefore, it is necessary for the null statements to be present among the statements of a basic block.

Null statements are added to the CFG on a flag, and this flag is set to true only for the CFG built in `SemaBounds.cpp`. These null statements are currently not being dumped during an AST dump or a CFG dump.

### DIFF
--- a/clang/include/clang/Analysis/CFG.h
+++ b/clang/include/clang/Analysis/CFG.h
@@ -1252,6 +1252,7 @@ public:
     bool MarkElidedCXXConstructors = false;
     bool AddVirtualBaseBranches = false;
     bool OmitImplicitValueInitializers = false;
+    bool AddNullStmt = false;
 
     BuildOptions() = default;
 
--- a/clang/lib/Analysis/CFG.cpp
+++ b/clang/lib/Analysis/CFG.cpp
@@ -540,6 +540,7 @@ public:
 
 private:
   // Visitors to walk an AST and construct the CFG.
+  CFGBlock *VisitNullStmt(Stmt *S);
   CFGBlock *VisitInitListExpr(InitListExpr *ILE, AddStmtChoice asc);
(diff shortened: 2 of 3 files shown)


kkjeer (approved):
LGTM.

mgrang (approved):
The change looks good. I just have a couple of questions:

1. I guess without printing the AST there is no real way to test this PR? If that is the case, we can add tests when implement dumper for null statements.

2. Do any existing unit tests need to be updated? The call to `autoCreateBlock()` will create a new block and any existing tests that have a null statement (standalone semicolon) and that check basic block numbers may need to be updated because of the introduction of a new block. It's fine if there are no such tests though.

sulekhark:
> The change looks good. I just have a couple of questions:
> 
> 1. I guess without printing the AST/CFG there is no real way to test this PR? If that is the case, we can add tests when implement dumper for null statements.
       
 Yes. For the purpose of this PR, I tested with gdb. The processing of stand-alone  `_Where` clauses (attached to null statements) in bounds widening analysis and available facts analysis will test this. I did not want to hold up this fix so that both bounds widening and available facts analyses can support stand-alone `_Where` clauses.
 
> 2. Do any existing unit tests need to be updated? The call to `autoCreateBlock()` will create a new block and any existing tests that have a null statement (standalone semicolon) and that check basic block numbers may need to be updated because of the introduction of a new block. It's fine if there are no such tests though.
       
 I have run all the tests (the four `DevTests` on the ADO machines - the test results are above - clicking on `show all checks` will give the details. None of the existing tests need to be updated.

PR #1111, opened 2021-06-26
Move ExpandToRange methods into BoundsUtil

kkjeer:
This PR moves methods to expand bounds to a range bounds expression into the `BoundsUtil` class.

### DIFF
--- a/clang/include/clang/Sema/BoundsUtils.h
+++ b/clang/include/clang/Sema/BoundsUtils.h
@@ -37,6 +37,11 @@ public:
   // be able to infer bounds for.
   static BoundsExpr *CreateBoundsAlwaysUnknown(Sema &S);
 
+  // Given an array type with constant dimension size, CreateBoundsForArrayType
+  // returns a count expression with that size.
+  static BoundsExpr *CreateBoundsForArrayType(Sema &S, QualType T,
+                                              bool IncludeNullTerminator = false);
+
   // If Bounds uses the value of LValue and an original value is provided,
   // ReplaceLValueInBounds will return a bounds expression where the uses
   // of LValue are replaced with the original value.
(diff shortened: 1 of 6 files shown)


mgrang:
`ExpandToRange` and `ExpandBoundsToRange` methods also normalize the bounds. Is it possible to simply call all these methods that normalize bounds as `NormalizeBounds`? That will greatly simplify the users of these methods. Otherwise, each time we would need to figure out whether to call `NormalizeBounds`, `ExpandToRange` or `ExpandBoundsToRange`.

sulekhark:
Each time `ExpandBoundsToRange` is called, it creates an `Expr *` `Base`. Therefore, it may not be a good idea to directly call `ExpandBoundsToRange` (particularly, it should not be called multiple number of times for the same `VarDecl`). The function `ExpandToRange` needs to be supplied `Base` as an argument. So, this function cannot be called directly. Katherine, please correct me if I am wrong.

In my opinion, `NormalizeBounds` should be the function called by clients because it calls `ExpandBoundsToRange` once and stores the expanded bounds as  normalized bounds in a `VarDecl`. Now, in the case of a bounds expression specified in a `_Where` clause, we don't have a "place" to store the normalized bounds. 
One solution could be to add a new field to `BoundsDeclFact` that will store the normalized bounds fact. Then, we can define a function called `BoundsExpr *Sema::NormalizeBounds(const BoundsDeclFact *BDF)`  or something similar. The key ideas are:
1) Clients should always call `NormalizeBounds` only (with the appropriate arguments) because `NormalizeBounds` will take care of calling `ExpandBoundsToRange` only once and storing the result.
2) To make (1) above possible, we need to create an appropriate field where the normalized bounds associated with the bounds expression in a `BoundsDeclFact` can be stored.

My opinion is that this PR is good for now and we can discuss what I am proposing above and take it up once we finalize the design.

sulekhark (approved):
LGTM! Thank you!

Issue #1107, opened 2021-06-24
Inconsistent warning when using pointer to incomplete structure type


john-h-kastner:
In this example, a warning is emitted on the first function call, but not on the second. This is odd for a couple of reasons. First, the bounds it can't prove are trivially correct, and will be checked correctly in other situations (without incomplete types).  Second, the function call with bound `count(1)` seems more problematic than the first, since it depends on knowing the size of an incomplete type. 

```c
#pragma CHECKED_SCOPE on
struct vsf_sysutil_statbuf;
void vsf_sysutil_free0(_Array_ptr<struct vsf_sysutil_statbuf> : count(0));
void vsf_sysutil_free1(_Array_ptr<struct vsf_sysutil_statbuf> : count(1));
void test(void) {
  _Ptr<struct vsf_sysutil_statbuf> p_statbuf = 0;
  vsf_sysutil_free0(p_statbuf);
  vsf_sysutil_free1(p_statbuf);
}
```

```
struct.c:7:21: warning: cannot prove argument meets declared bounds for 1st parameter [-Wcheck-bounds-decls-checked-scope]
  vsf_sysutil_free0(p_statbuf);
                    ^~~~~~~~~
struct.c:7:21: note: (expanded) expected argument bounds are 'bounds((_Array_ptr<struct vsf_sysutil_statbuf>)p_statbuf, (_Array_ptr<struct vsf_sysutil_statbuf>)p_statbuf + 0)'
struct.c:7:21: note: (expanded) inferred bounds are 'bounds((_Array_ptr<struct vsf_sysutil_statbuf>)p_statbuf, (_Array_ptr<struct vsf_sysutil_statbuf>)p_statbuf + 1)'
  vsf_sysutil_free0(p_statbuf);
                    ^~~~~~~~~
1 warning generated.
```

kkjeer:
Hi @john-h-kastner, this is currently a limitation in the compiler's ability to reason about bounds expressions. We plan to address this in the near term.

[...]

PR #1117, opened 2021-07-06
Handle incomplete types in BaseRange

kkjeer:
See issue #1107 

This PR modifies `BaseRange` to be able to compare constant-sized ranges whose base expressions have the same incomplete type.

Consider source bounds `bounds(p, p + a)` and target bounds `bounds(q, q + b)`, where `a` and `b` are constants. If `p` and `q` have types `T1` and `T2`, where `T1` and `T2` are incomplete types, then we cannot normalize `a` and `b` by multiplying them by the size of the referent types of `T1` and `T2`. However, we can prove that `bounds(p, p + a)` imply `bounds(q, q + b)` if `T1 == T2` and `a >= b`.


clang/test/CheckedC/static-checking/bounds-decl-checking.c
```diff
@@ -286,7 +288,37 @@ int f30(_Ptr<struct S> p) {
...
+
+int f33(_Ptr<struct S> p) {
+  test_f31(p);
+  return 0;
+}
+
+int f34(_Array_ptr<struct S> p : bounds(p, p), _Array_ptr<struct S> q : count(0)) { // expected-note {{(expanded) declared bounds are 'bounds(p, p)'}}
+  // The upper bound q + 0 is an incomplete constant sized upper offset,
+  // but the upper bound p is not an incomplete constant sized upper offset.
+  p = q; // expected-warning {{cannot prove declared bounds for 'p' are valid after assignment}} \
```
sulekhark, line 310: I think the expected warning on line 310 in this test case is not really correct. The lower/upper bounds of `p` are not marked as `Lower/UpperConstantIncomplete` because of the way these fields are set. Currently, these fields are set only when the call to `GetReferentSizeInChars` fails. In the case of the bounds expression `(p, p)`, I guess the control is going to the `exit` label in `SplitIntoBaseAndOffset` where the lower and upper bounds are split as `p + 0` without calling `GetReferentSizeInChars`, and therefore the incompleteness of the type of `p` is not recorded.

I suggest the following:
Try checking if the type of `Base` is incomplete in the code under the `exit` label in `SplitIntoBaseAndOffset`. If the type is incomplete, return `Kind::IncompleteConstantSized`.  If this gets further complicated or causes more issues/failures, file an issue for this test case and adding a TODO here with a reference to the filed issue. 

Ideally, a `BaseRange` object must be marked as having an incomplete type if the type of the base (in this case `p`) is incomplete.

sulekhark (approved):
LGTM! Thank you!

mgrang (approved):
LGTM.

Issue #1120, opened 2021-07-07
An nt_checked array with an empty initializer list should be an error


mgrang:
Consider the following declaration of an `nt_checked` array:
`char p nt_checked[] = {}`

According to the Checked C spec section 2.4:
1. `nt_checked` declares an array whose last element is a null terminator. The size of the array includes the null terminator element.
2. An `nt_checked` array with size `d` converts to an `nt_array_ptr` with a count of `d - 1` elements.

So it should be illegal to declare an `nt_checked` array with an empty initializer list.

mgrang:
Fixed in https://github.com/microsoft/checkedc-clang/pull/1121

PR #1121, opened 2021-07-07
An nt_checked array with an empty initializer list should be an error

mgrang:
Consider the following declaration of an nt_checked array:
char p nt_checked[] = {}

According to the Checked C spec section 2.4:

1. nt_checked declares an array whose last element is a null terminator. The size
   of the array includes the null terminator element.

2. An nt_checked array with size d converts to an nt_array_ptr with a count of d -
   1 elements.

So it should be illegal to declare an nt_checked array with an empty initializer list.

This fixes issue #1120.


sulekhark (approved):
LGTM! Thank you!

kkjeer (approved):
LGTM

PR #1127, opened 2021-07-13
Invertibility for unchecked pointers

### LINKED ISSUE #1123: Allow invertibility for unchecked pointers
We should calculate original values for unchecked pointers in invertible expressions. For example:

```
#pragma CHECKED_SCOPE off
void f(char * p : bounds(p, p + 1)) {
  p = p + 1;
}
```

Since the body of `f` is in a unchecked scope, `p` has an unchecked pointer type `char *`. The original value of `p` in the assignment `p = p + 1` should be `p - 1`.

kkjeer:
Fixes #1123 

This PR allows pointer arithmetic expressions with unchecked pointer type to be invertible. This means that unchecked pointers with bounds-safe interfaces in unchecked scopes can have inverses.

For example:

```
void f(int *p : count(1)) _Unchecked {
  p = p + 1;
}
```

In the body of `f`, `p` has the unchecked pointer type `int *` since it is in an unchecked scope. Before this PR, `p` would have no inverse in the expression `p + 1`. After this PR, `p` has the inverse `p - 1`.

### DIFF
--- a/clang/test/CheckedC/inferred-bounds/equiv-expr-sets.c
+++ b/clang/test/CheckedC/inferred-bounds/equiv-expr-sets.c
@@ -1055,7 +1055,7 @@ void original_value11(array_ptr<int> p, array_ptr<int> q) {
   // CHECK-NEXT: }
 }
 
-// UnaryOperator and ArraySubscriptExpr: no inverse (unchecked pointer arithmetic)
+// UnaryOperator and ArraySubscriptExpr: unchecked pointer arithmetic inverse
 void original_value12(int *p, int *q) {
   // Updated EquivExprs: { { p + 1, q } }
   q = p + 1;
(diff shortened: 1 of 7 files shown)


clang/test/CheckedC/inferred-bounds/equiv-expr-sets.c
```diff
@@ -1436,21 +1447,21 @@ void original_value19(array_ptr<int> *x, array_ptr<int> *i) {
   // CHECK-NEXT: }
   // CHECK-NEXT: }
 
-  // Original value of i in i - 1: x
-  // Updated EquivExprs: { { x - 1, i } }, Updated SameValue: { i + 1 }
+  // Original value of i in i - 1: i + 1
+  // Updated EquivExprs: { { i + 1, i } }, Updated SameValue: { i + 1 }
```
sulekhark, line 1451: Looks like there is a typo here: the updated equivalent expressions must be `{i+1, x}`.

sulekhark (approved):
LGTM! Thank you! There's just one minor typo.

mgrang (approved):
LGTM.

PR #1131, opened 2021-07-20
[BoundsWidening] Space improvements for bounds widening

mgrang:
We get rid of UnionGen and UnionKill sets to save memory consumed by the
compiler. We also do not need to compute the Gen and Kill sets for blocks thus
resulting in further space savings. Instead we now compute and use (but don't
store) the Out sets of each statement in the fixpoint loop.

For each block we only store the Out set of the last statement of the block
which is used to compute the Out set of the block.


sulekhark (approved):
LGTM! Thank you!

PR #1133, opened 2021-07-22
Update NormalizeUtil::TransformAssocLeft (fix Windows X86 failure)

kkjeer:
On Windows X86, we cannot convert `-1` to signed pointer width. This means that we cannot constant fold an expression like `((p + len) + -1) + 1`. However, we can constant fold `((p + len) - 1) + 1`. This means that we should not transform subtraction operators `X - Y` to `X + -Y` in the pointer-typed subexpression `(p + len) - 1`.

When we associate a pointer-typed subexpression like `p + (len - 1)` to the left, we should not first transform it to `p + (len + -1)`. Instead, `NormalizeUtil::TransformAssocLeft` should return `(p + len) - 1`. This PR updates `TransformAssocLeft` to take an expression of the form `E1 + (E2 +/- E3)` and return `(E1 + E2) +/- E3)`, where `E1` has pointer type and `E2` and `E3` have integer types.

### DIFF
--- a/clang/include/clang/AST/NormalizeUtils.h
+++ b/clang/include/clang/AST/NormalizeUtils.h
@@ -36,8 +36,8 @@ public:
   // This transformation will also be applied to E1 and E2.
   static Expr *TransformAdditiveOp(Sema &S, Expr *E);
 
-  // Input form:  E1 + (E2 + E3)
-  // Output form: (E1 + E2) + E3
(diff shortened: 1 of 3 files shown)


sulekhark (approved):
LGTM! Thank you!

mgrang (approved):
LGTM.

PR #1134, opened 2021-07-22
Support normalizing supplied bounds

mgrang:
If bounds expression B is provided normalize it to a range bounds
expression. In this case the normalized bounds are *not* attached to the
VarDecl D and will be recomputed each time NormalizeBounds is called with a
non-null bounds expression.
Else if bounds expression B is not provided then if the VarDecl D has a
byte_count or count bounds expression, NormalizeBounds expands it to a
range bounds expression.  The expanded range bounds are attached to the
VarDecl D to avoid recomputing the normalized bounds for D.

### DIFF
--- a/clang/include/clang/Sema/Sema.h
+++ b/clang/include/clang/Sema/Sema.h
@@ -5726,11 +5726,15 @@ public:
   // will always fail.
   void WarnDynamicCheckAlwaysFails(const Expr *Condition);
 
-  // If the VarDecl D has a byte_count or count bounds expression,
-  // NormalizeBounds expands it to a range bounds expression.  The expanded
-  // range bounds are attached to the VarDecl D to avoid recomputing the
-  // normalized bounds for D.
-  BoundsExpr *NormalizeBounds(const VarDecl *D);
+  // If bounds expression B is provided normalize it to a range bounds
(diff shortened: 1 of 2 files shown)


mgrang:
Note: This is needed for handling bounds widening in loops using invertibility of statements.

sulekhark (commented):
The function `NormalizeBounds` is expected to "normalize" a count or a byte_count bounds expression that is associated with a variable or a `BoundsDeclFact` only once and store the normalized bounds expression. This PR proposes to use the function `NormalizeBounds` in a different way. In an offline discussion, we revisited the need for this PR, and discussed alternate ways to resolve the issue that this PR addresses. So, this PR can be closed.

mgrang:
Since we deal with `RangeBoundsExpr` in the bounds widening analysis we pass a `RangeBoundsExpr` to `ReplaceLValueInBounds` and expect back a `RangeBoundsExpr`. So we do not explicitly need to "normalize" the returned bounds expression to a `RangeBoundsExpr`. We can simply `dyn_cast` the returned value to a `RangeBoundsExpr`. We will guard against a returned value which is not a `RangeBoundsExpr` by adding an `llvm_unreachable`. So we no longer need this PR.

PR #1135, opened 2021-07-22
Get variable and constant from upper bounds

kkjeer:
This PR adds a `GetVariableAndConstant` method to `NormalizeUtil` in order to compare two upper bound expressions. This enables the bounds checker to validate bounds that have been widened multiple times.

For example:
```
  if (*(p + len)) {
    if (*(p + len + 1)) {
      // Widened bounds before increment: bounds(p, ((p + len) + 1) + 1)
      // Inferred bounds after increment: bounds(p, (((p + (len - 1)) + 1)) + 1)
     ++len;
    }
  }
}
```

`GetVariableAndConstant` separates an input expression `E` into a pointer-typed expression `E1` and an integer constant `C` for as long as possible. It adds up each constant found this way and, if the result does not overflow, sets the out parameter `Constant` to the result of adding up all the constants and sets the out parameter `Variable` to the final pointer-typed subexpression of `E`.


sulekhark (approved):
LGTM! Thank you!

mgrang (approved):
LGTM.

PR #1137, opened 2021-07-26
[BoundsWidening] Use invertibility to support bounds widening in loops

mgrang:
An expression that modifies an LValue is said to be invertible w.r.t. the
LValue if we can write an expression in terms of the original value of the
LValue before the modification. For example, the expression x + 1 is invertible
w.r.t x because we can write this expression in terms of the original value of
x which is (x - 1) + 1.

In this PR, we use invertibility of statements to support bounds widening in
loops. More specifically, if a statement modifies a variable that occurs in the
bounds expression of a null-terminated array then instead of killing its bounds
at that statement we use invertibility of the statement to try to write the
widened bounds in terms of the original value of the variable.

### DIFF
--- a/clang/lib/Sema/BoundsWideningAnalysis.cpp
+++ b/clang/lib/Sema/BoundsWideningAnalysis.cpp
@@ -99,7 +99,7 @@ BoundsMapTy BoundsWideningAnalysis::GetOutOfLastStmt(
     // If this is the last statement of the current block, then at this point
     // InOfCurrStmt contains the Out set of the second last statement of the
     // block.  This is equal to the In set for the last statement of this
-    // block. So we set InOfLastStmt to StmtOut.
+    // block. So we set InOfLastStmt to InOfCurrStmt.
     if (CurrStmt == EB->LastStmt)
       EB->InOfLastStmt = InOfCurrStmt;
 
@@ -153,7 +153,7 @@ BoundsMapTy BoundsWideningAnalysis::GetOutOfLastStmt(
       // a bounds expression which is strictly wider than the declared upper
       // bound.
       // So we will proceed only if AdjustedRangeBounds is wider than
-      // StmtOut[V] which contains the delcared bounds of V at this point. For
+      // StmtOut[V] which contains the declared bounds of V at this point. For
       // example:
 
       // Let DB = Declared bounds, AB = Adjusted bounds.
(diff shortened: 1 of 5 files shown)


clang/lib/Sema/BoundsWideningAnalysis.cpp
```diff
@@ -511,15 +584,20 @@ BoundsMapTy BoundsWideningAnalysis::GetStmtOut(const CFGBlock *B,
...
   ElevatedCFGBlock *EB = BlockIt->second;
 
-  if (CurrStmt) {
-    auto Diff = BWUtil.Difference(EB->OutOfPrevStmt, EB->StmtKill[CurrStmt]);
-    auto StmtOut = BWUtil.Union(Diff, EB->StmtGen[CurrStmt]);
-    EB->OutOfPrevStmt = StmtOut;
-    return StmtOut;
+  if (!CurrStmt) {
+    EB->OutOfPrevStmt = EB->In;
+    return EB->In;
```
sulekhark, line 589: Does this mean that a client has to call `GetStmtOut` with a null value for `CurrStmt` in order to initialize `EB->OutOfPrevStmt` (because it seems like `EB->OutOfPrevStmt` needs to have a valid value for this function to return sensible values). Should we be returning `BoundsMapTy()` if `CurrStmt` is null and initializing `EB->OutOfPrevStmt` to `EB->In` if `EB->OutOfPrevStmt` is null?

clang/lib/Sema/BoundsWideningAnalysis.cpp
```diff
@@ -700,18 +783,125 @@ void BoundsWideningAnalysis::GetVarsAndBoundsForModifiedVars(
...
     BoundsExpr *NormalizedBounds = SemaRef.NormalizeBounds(V);
     VarsAndBounds[V] = dyn_cast_or_null<RangeBoundsExpr>(NormalizedBounds);
   }
+
+  // If the modification of a variable by the current statement affects the
+  // bounds of a null-terminated array, then check invertibility of the
+  // statement. If the statement is invertible then store the statement, the
+  // modified LValue, the original LValue and the set of null-terminated arrays
+  // whose bounds are affected by the statement. We will use this info in the
+  // computation of the Out sets of blocks.
```
sulekhark, line 804: Nit: ...computation of the Out sets of blocks -> ...computation of the Out sets of statements.

clang/test/CheckedC/inferred-bounds/bounds-context.c
```diff
@@ -300,16 +300,14 @@ void assign1(array_ptr<int> arr : count(1)) { // expected-note {{(expanded) decl
 // Assignment to a variable used in other variables' bounds
 void assign2(
   array_ptr<int> a : count(len - 1), // expected-note {{(expanded) declared bounds are 'bounds(a, a + len - 1)'}}
-  char b nt_checked[0] : count(len), // expected-note {{(expanded) declared bounds are 'bounds(b, b + len)'}}
+  char b nt_checked[0] : count(len),
```
sulekhark, line 303: This change on line 303 and lines 311 and 312 do not seem to be caused by the implementation of invertibility. It will be great if you could give some context for this change.

mgrang: These changes are a direct result of invertibility. `b` is a null-terminated array which uses `len` in its bounds and `len` is modified. Without invertibility we would have killed the bounds of `b` and issued the warning/note messages. But with invertibility we can "adjust" the bounds of `b` to `bounds(b, b + len - 3)`.

mgrang: Filed https://github.com/microsoft/checkedc-clang/issues/1138 to track the invalid declaration of `nt_checked` array of size 0.

sulekhark (approved):
LGTM! Thank you!

PR #1139, opened 2021-07-27
[BoundsWidening] Determine checked scope specifier per statement

mgrang:
We use the CheckedCAnalysesPrepass.cpp to gather the checked scopes for
statements. We store a map of statements to their checked scope specifiers. An
entry in this map is only made for the following statements:
1. For the first non-compound statement of a compound statement.
2. For the first statement that follows a compound statement.

We then use this info in the bounds widening analysis to determine the checked
scope specifiers for each statement.

### DIFF
--- a/clang/lib/Sema/CheckedCAnalysesPrepass.cpp
+++ b/clang/lib/Sema/CheckedCAnalysesPrepass.cpp
@@ -277,6 +277,41 @@ class PrepassHelper : public RecursiveASTVisitor<PrepassHelper> {
       return ProcessWhereClause(S->getWhereClause());
     }
 
+    bool VisitCompoundStmt(CompoundStmt *CS) {
+      if (!CS)
+        return true;
+
+      // Get the checked scope specifier for the current compound statement.
+      CheckedScopeSpecifier CSS = CS->getCheckedSpecifier();
+
+      bool FirstStmt = true;
+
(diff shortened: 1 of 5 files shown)


clang/lib/Sema/CheckedCAnalysesPrepass.cpp
```diff
@@ -277,6 +277,41 @@ class PrepassHelper : public RecursiveASTVisitor<PrepassHelper> {
...
+	if (FirstStmt) {
+          Info.CheckedScopeMap[CurrStmt] = CSS;
+          FirstStmt = false;
+
+	// Else if the previous statement was a compound statement store the
+	// checked scope specifier of the current statement.
+	} else if (isa<CompoundStmt>(*(I - 1))) {
+          Info.CheckedScopeMap[CurrStmt] = CSS;
+        }
+      }
```
sulekhark, line 310: 1) Labeled statements should also have an entry in `CheckedScopeMap` because basic blocks may start with labeled statements. We need to be able to update the CSS value correctly at the beginning of a basic block as we are assured of ordered lookup only within a basic block (i.e. the first statement of a basic block may be accessed out of statement-order).
3) It is important for this feature to have test cases.
4) It looks like there is an indentation issue in lines 294 - 310.

sulekhark (approved):
LGTM! Thank you!

PR #1141, opened 2021-07-30
Revert "[BoundsWidening] Determine checked scope specifier per statement (#1139)"

mgrang:
This reverts commit 980321d2378918268e23dc758ca55efc76a5018b.

We are working on a cleaner solution to determining checked scopes per statement.

PR #1142, opened 2021-07-31
Determine checked scopes per statement

mgrang:
We introduce a 2-bit field called CheckedScopeSpecifier in the Stmt class.
During parsing when a compound statement is created we iterate the elements
(statements) of the compound statement and set the checked scope specifier for
each element to the checked scope specifier of the compound statement.

We can get the checked scope specifier for a statement by calling the
getCheckedScopeSpecifier method on the statement.


mgrang:
Depends on https://github.com/microsoft/checkedc-clang/pull/1141

clang/include/clang/AST/Stmt.h
```diff
@@ -102,8 +102,10 @@ class alignas(void *) Stmt {
 
     /// The statement class.
     unsigned sClass : 8;
+    /// The checked scope specifier for the statement.
+    unsigned CSS : 2;
```
sulekhark, line 106: The field `CSS` is added as a sibling field of  `sClass` in the class `StmtBitfields` - i.e. it is present for **every** "Stmt". The field `sClass` is initialized in the constructor for `Stmt` at line 1200  (`in AST/Stmt.h` on `master` branch). It is also possible to initialize the field `CSS` in the same constructor because `GetCheckedScopeInfo` (which reads `CheckingKind` that stores the currently applicable value of the checked scope specification) is a `Sema` function. This will avoid looping over the statements in the body of a compound statement, which may be unavoidable if the `CSS` field is set in `ActOnCompoundStmt`.

@dtarditi It will be great to know your thoughts on this.

kkjeer (approved):
LGTM.

PR #1149, opened 2021-08-09
[BoundsWidening] Handle complex conditionals in bounds widening

mgrang:
Support bounds widening in presence of complex conditionals like:
  "if (*p != 0)", "if ((c = *p) == 'a')", etc.

### DIFF
--- a/clang/include/clang/Sema/BoundsWideningAnalysis.h
+++ b/clang/include/clang/Sema/BoundsWideningAnalysis.h
@@ -71,29 +71,12 @@ namespace clang {
     // if (*(p + i) == 0) ==> DerefExpr = p + i
     Expr *DerefExpr;
 
-    // Whether the value that an expression is compared against is null. For
-    // example:
-    // if (*p == 0)   ==> IsComparedValNull = True
-    // if (*p == 'a') ==> IsComparedValNull = False
-    bool IsComparedValNull;
-
-    // Whether the operator of the conditional is positive. For example:
-    // if (*p == 0) ==> IsCheckPositive = True
-    // if (*p != 0) ==> IsCheckPositive = False
-    bool IsCheckPositive;
-
-    // Whether the terminating condition is an expression involving a binary
-    // operator. For example:
-    // if (*p == 0) ==> HasBinaryOp = True
-    // if (!*p)     ==> HasBinaryOp = False
-    bool HasBinaryOp;
-
-    // Whether the terminating condition is an expression involving a unary not
-    // operator. For example:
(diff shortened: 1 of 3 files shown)


clang/include/clang/Sema/BoundsWideningAnalysis.h
```diff
@@ -63,6 +63,39 @@ namespace clang {
...
+    bool HasBinaryOp;
+
+    // Whether the terminating condition is an expression involving a unary not
+    // operator. For example:
+    // if (!*p)        ==> HasUnaryNot = True
+    // if (!(*p == 0)) ==> HasUnaryNot = True
+    // if (*p == 0)    ==> HasUnaryNot = False
+    bool HasUnaryNot;
+  };
+
```
sulekhark, line 98: 1) The fields `HasBinaryOp` and `HasUnaryNot` are book-keeping variables required only to set the other fields when an instance of `TermCondInfoTy` is initialized. Ideally, they need not persist as fields in this structure.
2) The fields `IsComparedValNull` and `IsCheckPositive` can be fused into one field, say for example `IsCheckNull` (effectively folding the logic in the function `DoesTermCondCheckNull` into function `FillTermcondInfo`). In other words, this structure needs to have only two fields: the dereferenced expression, and a boolean field that indicates if the terminating condition is checking the nullness or non-nullness of the (potentially) upperbound element.
3) Doing 1) and 2) above will also probably simplify the logic in function `FillTermCondInfo`: if the (fused) field `IsCheckNull` is set **after** the recursive call to function `FillTermCondInfo`, I suspect that the book-keeping variables may not be required even as local variables.

clang/lib/Sema/BoundsWideningAnalysis.cpp
```diff
@@ -785,8 +806,9 @@ void BoundsWideningAnalysis::GetVarsAndBoundsInPtrDeref(
 
     RangeBoundsExpr *WidenedBounds =
       new (Ctx) RangeBoundsExpr(R->getLowerExpr(),
-                                BWUtil.AddOffsetToExpr(DerefExpr, 1),
-                                SourceLocation(), SourceLocation());
+        BWUtil.AddOffsetToExpr(EB->TermCondInfo.DerefExpr, 1),
+        SourceLocation(), SourceLocation());
```
arbipher, line 810: I am not sure if it would be more meaningful to use the `SourceLocation` of the widen expr here.
I face the similar question.

sulekhark (approved):
LGTM! Thank you!

PR #1154, opened 2021-08-12
Verify if an expr contains errors before checking invertibility 

mgrang:
The community has introduced a new annotation called "contains-errors" on AST
nodes that contain semantic errors. As a result, after the upgrade of Checked C
sources to LLVM 12 we need to check if an expr contains errors before operating
on the expr. One such place is in InverseUtil::IsInvertible where we need to
check if the input modifying expr contains errors.

### DIFF
--- a/clang/lib/AST/ExprUtils.cpp
+++ b/clang/lib/AST/ExprUtils.cpp
@@ -469,7 +469,7 @@ void ExprUtil::EnsureEqualBitWidths(llvm::APSInt &A, llvm::APSInt &B) {
 }
 
 bool InverseUtil::IsInvertible(Sema &S, Expr *LValue, Expr *E) {
-  if (!E)
+  if (!E || E->containsErrors())
     return false;
 
   E = E->IgnoreParens();


clang/lib/AST/ExprUtils.cpp
```diff
@@ -469,7 +469,7 @@ void ExprUtil::EnsureEqualBitWidths(llvm::APSInt &A, llvm::APSInt &B) {
 }
 
 bool InverseUtil::IsInvertible(Sema &S, Expr *LValue, Expr *E) {
-  if (!E)
+  if (!E || E->containsErrors())
```
sulekhark, line 472: Maybe we should add the same check in `InverseUtil::Inverse(...)`?

mgrang: Should we do this on a case-by-case basis?

sulekhark: Yes. But we could think of functions `IsInvertible` and `Inverse` as the same "case". Though `Inverse` is called only when `IsInvertible` returns true, it may be safe to add the check in the function `Inverse` also.

sulekhark (approved):
LGTM! Thank you!

kkjeer (approved):
LGTM! Thanks!

PR #1162, opened 2021-08-18
Don't record temporary equality between expression such as x and x + 1 in TargetSrcEquality

kkjeer:
This PR fixes a bug with the way equivalent expressions were being recorded in `RecordEqualityWithTarget`.

In an assignment such as `x++`, `x += 1`, `x = x + 1`, etc., `SameValue` is empty after calling `UpdateSameValueAfterAssignment` (since the RHS `x + 1` of the assignment uses the value of the LHS `x`). This meant that `SrcAllowedInEquivExprs` was `false` in `RecordEqualityWithTarget`, so the mapping `x => x + 1` was added to `TargetSrcEquality`. The information in `TargetSrcEquality` is added to `EquivExprs` in `ValidateBoundsContext`, so `EquivExprs` would contain a set that contained both `x` and `x + 1` only while checking bounds after the current top-level statement.

This PR adds an `AllowTempEquality` argument to `RecordEqualityWithTarget` that controls whether the mapping `Target => Src` is permitted to be added to `TargetSrcEquality`. `UpdateSameValueAfterAssignment` now returns `true` if `State.SameValue` was unchanged by the assignment (if the RHS of the assignment uses the value of the LHS, then at least one expression will be removed from `State.SameValue` if `State.SameValue` was initially nonempty). If `State.SameValue` was unchanged by the assignment, then temporary equality is allowed to be recorded between `Target` and `Src`.

### DIFF
--- a/clang/lib/Sema/SemaBounds.cpp
+++ b/clang/lib/Sema/SemaBounds.cpp
@@ -4703,13 +4703,14 @@ namespace {
                                         CheckedScopeSpecifier CSS,
                                         CheckingState &State) {
       // Adjust SameValue to account for any uses of LValue in PrevSameValue.
-      // If the original value uses the value of V, then any expressions that
-      // use the value of V should be removed from SameValue.
+      // If the original value uses the value of LValue, then any expressions
+      // that use the value of LValue should be removed from SameValue.
       // For example, in the assignment i = i + 2, where the original value
       // of i is i - 2, the expression i + 2 in SameValue should be removed
       // rather than replaced with (i - 2) + 2.
-      // Otherwise, SameValue would contain (i - 2) + 2 and i, which is a
-      // tautology.
+      // Otherwise, EquivExprs would eventually contain (i - 2) + 2 and i,
+      // which is a tautology.
+      bool ModifiedSameValue = false;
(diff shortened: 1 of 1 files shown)


clang/lib/Sema/SemaBounds.cpp
```diff
@@ -4710,25 +4721,38 @@ namespace {
         if (AdjustedE && !EqualExprsContainsExpr(State.SameValue, AdjustedE))
           State.SameValue.push_back(AdjustedE);
       }
+      return State.SameValue.size() == PrevSameValue.size();
```
sulekhark, line 4724: Is it sufficient to check that `State.SameValue.size() == PrevSameValue.size()` to be sure that no expressions in `SameValue` use the value of `LValue`? What if the number of elements in `SameValue` remain the same but one or more of these elements are modified?

kkjeer: If the original value uses the value of `LValue`, then the value used to replace uses of `LValue` in `SameValue` will be null, which means that any expressions in `SameValue` that use the value of `LValue` will be removed from `SameValue`.

I think this could be clearer in the code; I'll work on cleaning up the solution.

sulekhark (approved):
LGTM! Thank you!

PR #1163, opened 2021-08-18
Treat address-of array subscripts the same way as address-of dereferences

### LINKED ISSUE #1148: Inconsistent behavior with `str` and `&str[0]`
For this code
```
#include <stdio.h>
void foo(_Nt_array_ptr<char> str) {
  sprintf(str,"%s","hello");     // clang complains about incompatible type of str
  sprintf(&str[0],"%s","hello"); // no complaints
}
```
The `stdio_checked.h` file defines `sprintf`'s first argument to be a `char *`. As such, passing `str` to it is rejected as incompatible (passing a checked type where an unchecked one was expected). But the second line is accepted, even though it is passing literally the same address to `sprintf` as the first. (The first line is accepted if you add an unsafe cast, e.g., `sprintf((char *)str,"%s","hello");`)

kkjeer:
Fixes #1148

This PR modifies the type checker so that, if an expression `e` has type `T`, then `&e[idx]` and `&idx[e]` also have type `T`. This is similar to the current behavior where, if `e` has type `T`, then `&*e` also has type `T`.

From the C spec section 6.5.3.2:

> Similarly, if the operand is the result of a [] operator, neither the & operator nor the unary * that is implied by the [] is evaluated and the result is as if the & operator were removed and the [] operator were changed to a + operator.

This is similar to the rules for `&*e`:

> If the operand is the result of a unary * operator, neither that operator nor the & operator is evaluated and the result is as if both were omitted, except that the constraints on the operators still apply and the result is not an lvalue.

### DIFF
--- a/clang/lib/Sema/SemaExpr.cpp
+++ b/clang/lib/Sema/SemaExpr.cpp
@@ -14203,6 +14203,11 @@ QualType Sema::CheckAddressOfOperand(ExprResult &OrigOp, SourceLocation OpLoc) {
 
   if (getLangOpts().C99) {
     // Implement C99-only parts of addressof rules.
+    // This logic should be near the end of this method (as opposed to the
+    // C99-only rules for addressof dereferences above) so that the necessary
+    // checks can be made for addressof array subscript expressions (taking
+    // the address of a register variable, taking the address of a vector
+    // element, etc.).
     if (ArraySubscriptExpr *arrSubscript = dyn_cast<ArraySubscriptExpr>(op)) {
       // Per C99 6.5.3.2, the address of an array subscript always returns
       // a valid result (assuming the array subscript expression is valid).
--- a/clang/lib/Sema/SemaExpr.cpp
+++ b/clang/lib/Sema/SemaExpr.cpp
@@ -14201,18 +14201,13 @@ QualType Sema::CheckAddressOfOperand(ExprResult &OrigOp, SourceLocation OpLoc) {
 
   CheckAddressOfPackedMember(op);
 
-  if (getLangOpts().C99) {
-    // Implement C99-only parts of addressof rules.
-    // This logic should be near the end of this method (as opposed to the
-    // C99-only rules for addressof dereferences above) so that the necessary
-    // checks can be made for addressof array subscript expressions (taking
-    // the address of a register variable, taking the address of a vector
-    // element, etc.).
-    if (ArraySubscriptExpr *arrSubscript = dyn_cast<ArraySubscriptExpr>(op)) {
-      // Per C99 6.5.3.2, the address of an array subscript always returns
-      // a valid result (assuming the array subscript expression is valid).
+  // For Checked C, &e1[e2] should be a pointer to T if e1 or e2 is a pointer
+  // to T, regardless of checked scope. This avoids the unexpected result of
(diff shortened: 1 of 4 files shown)


clang/lib/Sema/SemaExpr.cpp
```diff
@@ -14201,6 +14199,15 @@ QualType Sema::CheckAddressOfOperand(ExprResult &OrigOp, SourceLocation OpLoc) {
 
   CheckAddressOfPackedMember(op);
 
+  if (getLangOpts().C99) {
```
dtarditi, line 14202: I'm curious why this code is not placed where the comment was removed.  Why place it here?

Is this a change that is specific to the Checked C version of clang?  Or should it be propagated upstream?  Put another way, is this a bug in clang?  Or is there something specific about Checked C that requires this change?

dtarditi (commented):
It's great to see this change.  I have a question about whether this is Checked C specific or something that should be upstreamed.

clang/lib/Sema/SemaExpr.cpp
kkjeer: This code was placed here (rather than where the comment was removed) so that it occurs after the check on [line 14127](https://github.com/microsoft/checkedc-clang/blob/master/clang/lib/Sema/SemaExpr.cpp#L14127) for taking the address of an lvalue. If this code is placed where the comment is removed, then certain tests (e.g. Sema/complex-imag.c) fail due to missing expected errors (they expect errors "cannot take the address of an rvalue of type <T>" to be emitted).

I can add a comment explaining why this code is placed where it is.

To the best of my knowledge, this change isn't Checked C specific.

kkjeer: If this code is moved to where the comment was removed, then the Sema/expr-address-of.c test fails. This test expects the following error:

```
void foo() {
  register int x[10];
  &x[10];              // expected-error {{address of register variable requested}}
}
```

The "address of register variable requested" error is emitted in the call to `diagnoseAddressOfInvalidType` on [line 14187](https://github.com/microsoft/checkedc-clang/blob/master/clang/lib/Sema/SemaExpr.cpp#L14187). If the code in this PR is added before line 14187, then this expected error will not be emitted.

sulekhark (approved):
LGTM! Thank you!

clang/lib/Sema/SemaExpr.cpp
dtarditi: I looked at the code you are adding and I don't think it is needed for the regular C path.   Taking the address of a subscript expression `e1[e2]`, where e1 or e2 is a pointer to T, will result in a pointer to T being created with the existing code.  The subscript expression will have type T and the final statement will create a pointer to T.    

I believe the code is really only needed for the Checked C path.  It might be better to put this under a Checked C flag.   You could then explain in the comment that the code avoids the unexpected result of &e1[e2] having a different kind of pointer type than the pointer type that is being subscripted.  This can happen in unchecked scopes where the `&` operator is applied to a subscript expression involving a checked pointer.

dtarditi (commented):
I looked at the code and think the change is only needed for Checked C.  I think it would be helpful clarify that.

dtarditi (approved):
LGTM.  Thanks!

PR #1166, opened 2021-08-23
Update bounds checking notes

kkjeer:
This PR makes a minor modification to the notes emitted during bounds checking:

Only one note is emitted per statement per lvalue expression for an expression with unknown bounds being assigned to an lvalue expression.
For example:
```
void f(_Array_ptr<int> p : count(1), _Array_ptr<int> q : bounds(unknown) {
  p = q, p++;
}
```
Both assignments `p = q` and `p++` assign expressions (`q` and `p + 1` respectively) with unknown bounds to `p`. A note will only be emitted for the first assignment. This helps avoid confusion in the note messages - it may be unclear to the user why the expression `p++` would result in the note "assigned expression p + 1 with unknown bounds to p".


sulekhark (approved):
LGTM! Thank you!

PR #1174, opened 2021-08-30
Support variadic function calls in checked scope

mgrang:
We add support for calling variadic functions in checked scope. These are
functions like printf, scanf, etc that take a format string and have a variable
number of arguments. We implement checking of arguments to these functions.
Following is a list of some important checks that we implement in checked scope
for these functions:

- check that the argument corresponding to the %s format specifier is a
  null-terminated array.
- all warnings emitted by the -Wformat family of flags have been converted to
  errors in checked scope.

### DIFF
--- a/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -11507,7 +11507,7 @@ def err_bounds_type_annotation_lost_checking : Error<
     "variable arguments function cannot be made in a checked scope">;
 
   def err_checked_scope_no_variadic_func_for_expression : Error<
-    "cannot use a variable arguments function in a checked scope or function">;
+    "cannot use this variable arguments function in a checked scope or function">;
 
   def err_checked_scope_no_assume_bounds_casting : Error<
     "_Assume_bounds_cast not allowed in a checked scope or function">;
--- a/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -11520,6 +11520,12 @@ def err_bounds_type_annotation_lost_checking : Error<
   "%select{'_Unchecked'|'_Checked _Bounds_only|'_Checked'}0 "
   "can only appear on functions">;
 
+  def err_checked_scope_invalid_format_specifier_argument : Error<
+    "in a checked scope %0 format specifier requires %1 argument">;
+
+  def err_checked_scope_scanf_width : Error<
+    "in a checked scope width is not allowed with format specifier in scanf">;
+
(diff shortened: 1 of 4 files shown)


sulekhark (approved):
LGTM! Thank you!

kkjeer (approved):
LGTM. Thanks!

mattmccutchen-cci:
I realize I may have waited a little too late to post this, but I tested this PR and found several holes in the new checking.  If you save the code below as `printf-checking-bugs.c` and then run `clang -o printf-checking-bugs printf-checking-bugs.c` and `./printf-checking-bugs percent_n`, etc., you can see the SEGV.  If you don't want to address the problems in this PR, I can copy the examples to a new issue.

```c
#pragma CHECKED_SCOPE on

#include <string.h>
#include <stdio.h>
#include <unistd.h>
#include <fcntl.h>

#define TEST(_name) else if (strcmp(test_name, #_name) == 0)

int main(int argc, _Nt_array_ptr<_Nt_array_ptr<char>> argv : count(argc)) {
  _Nt_array_ptr<char> test_name = argv[1];
  if (!test_name) {
    fprintf(stderr, "No test name specified\n");
    return 1;
  }

  if (0) {}
  TEST(percent_n) {
    // Missing check that %n argument is a _Ptr.
    int arr _Checked[1];
    printf("hello\n%n", arr + 123456789);
  }
  TEST(scanf_scalar) {
    // Missing check that _any_ scalar scanf argument is a _Ptr.
    int arr _Checked[1];
    sscanf("42", "%d", arr + 123456789);
  }
  TEST(printf_s_count) {
    // Missing check that printf %s argument has at least count(0).
    char buf _Nt_checked[1];
    printf("%s", buf + 123456789);
  }
  TEST(scanf_p) {
    // scanf reads an arbitrary _Ptr<void> via %p. The right solution here may
    // be to disallow _Ptr<void>
    // (https://github.com/microsoft/checkedc/issues/335). I couldn't find any
    // other way to exploit %p, but that doesn't mean there isn't any.
    _Ptr<void> q = 0;
    sscanf("0x1", "%p", &q);
    _Ptr<char> p = (_Ptr<char>)q;
    (*p)++;
  }
  TEST(scanf_s_overflow) {
    // scanf %s overflows the output buffer. I guess the compiler should require
    // the format string to specify a maximum width and check it against the
    // bounds of the argument?
    char field _Nt_checked[10];
    char input _Nt_checked[1000];
    memset(input, 'x', sizeof input - 1);
    sscanf(input, "%s", field);
  }
  else {
    fprintf(stderr, "Unknown test name\n");
    return 1;
  }

  return 0;
}
```

mgrang:
Thanks @mattmccutchen-cci for identifying these issues. I have filed https://github.com/microsoft/checkedc-clang/issues/1178 to track these.

PR #1176, opened 2021-09-01
Bounds checking for pointer dereferences and array subscripts

kkjeer:
This PR updates bounds validation to handle the following cases:
1. Validate the bounds of a pointer dereference `*p` or array subscript `p[i]`.
2. Account for uses of pointer dereferences and array subscripts in observed bounds

The behavior for synthesizing member expressions whose bounds depend on a member expression that is being modified via an assignment has been extended: we now synthesize member expressions whose bounds depend on an lvalue expression that uses a member expression to update memory via an assignment. For example:

```
struct S {
  _Array_ptr<int> f : count(*ptr_to_len);
  _Array_ptr<int> ptr_to_len : count(10);
};

void f(struct S *s) {
  *s->ptr_to_len = 0;
}
```

The lvalue expression `*s->ptr_to_len` uses the member expression `s->ptr_to_len` to write to memory. We synthesize the member expression `s->f` whose declared bounds `bounds(s->f, s->f + *s->ptr_to_len)` depend on `*s->ptr_to_len`.

There is also a minor fix included in this PR that the pointer bounds checking depends on: in PreorderAST, the canonical form of `e1[e2]` is now `*(e1 + e2 + 0)` rather than `*(e1 + e2)`. This enables bounds validation to treat expressions such as `p[i]` and `*(p + i)` as equivalent, since PreorderAST canonicalizes `*(p + i)` to `*(p + i + 0)`.

### DIFF
--- a/clang/lib/Sema/BoundsUtils.cpp
+++ b/clang/lib/Sema/BoundsUtils.cpp
@@ -242,6 +242,30 @@ namespace {
           return E;
       }
 
+      ExprResult TransformUnaryOperator(UnaryOperator *E) {
+        if (!ExprUtil::IsDereferenceOrSubscript(LValue))
+          return E;
+        if (Lex.CompareExprSemantically(LValue, E)) {
+          if (OriginalValue)
+            return OriginalValue;
+          else
+            return ExprError();
+        } else
+          return E;
+      }
+
+      ExprResult TransformArraySubscriptExpr(ArraySubscriptExpr *E) {
+        if (!ExprUtil::IsDereferenceOrSubscript(LValue))
+          return E;
+        if (Lex.CompareExprSemantically(LValue, E)) {
+          if (OriginalValue)
+            return OriginalValue;
+          else
+            return ExprError();
+        } else
+          return E;
+      }
+
       // Overriding TransformImplicitCastExpr is necessary since TreeTransform
       // does not preserve implicit casts.
       ExprResult TransformImplicitCastExpr(ImplicitCastExpr *E) {
--- a/clang/lib/Sema/BoundsUtils.cpp
+++ b/clang/lib/Sema/BoundsUtils.cpp
@@ -225,10 +225,9 @@ namespace {
         if (Lex.CompareExpr(V, E) == Lexicographic::Result::Equal) {
           if (OriginalValue)
             return OriginalValue;
-          else
-            return ExprError();
-        } else
-          return E;
+          return ExprError();
+        }
+        return E;
       }
 
       ExprResult TransformMemberExpr(MemberExpr *E) {
@@ -238,10 +237,9 @@ namespace {
         if (Lex.CompareExprSemantically(M, E)) {
           if (OriginalValue)
             return OriginalValue;
-          else
-            return ExprError();
-        } else
-          return E;
+          return ExprError();
+        }
+        return E;
       }
 
       ExprResult TransformUnaryOperator(UnaryOperator *E) {
@@ -250,10 +248,9 @@ namespace {
         if (Lex.CompareExprSemantically(LValue, E)) {
           if (OriginalValue)
             return OriginalValue;
-          else
-            return ExprError();
-        } else
-          return E;
+          return ExprError();
+        }
+        return E;
       }
 
       ExprResult TransformArraySubscriptExpr(ArraySubscriptExpr *E) {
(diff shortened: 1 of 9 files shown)


clang/test/CheckedC/inferred-bounds/bounds-context.c
```diff
@@ -1857,29 +1857,80 @@ void inc_dec_bounds4(array_ptr<int> a : bounds(a, a)) { // expected-note {{(expa
...
-// do not result in bounds checking-related warnings or errors
+// Increment/decrement operators on dereference expressions or array subscripts
+// of type _Nt_array_ptr<T> are bounds checked
+// Increment/decrement operators on variables without declared bounds are not
+// bounds checked
 void inc_dec_bounds5(nt_array_ptr<int> *p, array_ptr<int> a) {
-  // Observed bounds context after increment:  { }
-  ++*p;
+  // Observed bounds context after increment:  { *p => bounds(*p - 1, (*p - 1) + 0) }
+  ++*p; // expected-warning {{cannot prove declared bounds for '*p' are valid after increment}} \
```
sulekhark, line 1866: I think this behavior of the compiler is incorrect. Perhaps, at present, we should restrict the bounds validation for `*p` and `p[i]` when the type of `*p` or `p[i]` is a checked pointer type?

kkjeer: In this particular test case, `*p` has type `_Nt_array_ptr<int>`, since `p` has type `_Nt_array_ptr<int> *`.

In general, if `e` is a pointer dereference `*e1` or an array subscript `e1[e2]`, we will only check the bounds of `e` if `e` fulfills one of two conditions:

1. `e` has type `_Ptr<T>`. In this case, the target bounds of `e` are `bounds((_Array_ptr<T>)e, (_Array_ptr<T>)e + 1)`.
2. `e` has type `_Nt_array_ptr<T>`. In this case, the target bounds of `e` are `bounds(e, e + 0)`.

We do not explicitly check the type of `e` in `UpdateAfterAssignment` or `ValidateBoundsContext`. Instead, the `AbstractSet` containing `e` is added to `ObservedBounds` if and only if `e` has target bounds, which will only happen in the two cases described above.

The methods `UnaryOperatorTargetBounds` and `ArraySubscriptExprTargetBounds` in SemaBounds.cpp include comments that state that "Currently, we don't know the target bounds of a pointer stored in a pointer dereference (or returned by a subscripting operation), unless it is a _Ptr type or an _Nt_array_ptr."

sulekhark: Thanks for the clarification (I missed noticing the `*`)!

sulekhark (approved):
LGTM! Thank you!

clang/lib/Sema/BoundsUtils.cpp
```diff
@@ -244,6 +244,30 @@ namespace {
           return E;
       }
 
+      ExprResult TransformUnaryOperator(UnaryOperator *E) {
+        if (!ExprUtil::IsDereferenceOrSubscript(LValue))
+          return E;
+        if (Lex.CompareExprSemantically(LValue, E)) {
+          if (OriginalValue)
+            return OriginalValue;
+          else
```
mgrang, line 253: Nit: `else` is not needed after a return.

clang/lib/Sema/BoundsUtils.cpp
```diff
@@ -244,6 +244,30 @@ namespace {
...
+            return ExprError();
+        } else
```
mgrang, line 255: Same here.

clang/lib/Sema/BoundsUtils.cpp
```diff
@@ -244,6 +244,30 @@ namespace {
...
+          return E;
+      }
+
+      ExprResult TransformArraySubscriptExpr(ArraySubscriptExpr *E) {
+        if (!ExprUtil::IsDereferenceOrSubscript(LValue))
+          return E;
+        if (Lex.CompareExprSemantically(LValue, E)) {
+          if (OriginalValue)
+            return OriginalValue;
+          else
```
mgrang, line 265: Same comments here.

mgrang:
LGTM.

mgrang (approved):

PR #1180, opened 2021-09-03
Separate out windows and non-windows-specific tests for variadic functions

mgrang:
Some -Wformat error messages are different between linux and windows systems.
We separate out the common tests in variadic-functions.c. The windows-specific
tests are in variadic-functions-win.c and the non-windows tests are in
variadic-functions-non-win.c.


sulekhark (approved):
LGTM! Thanks!

kkjeer (approved):
LGTM.

PR #1181, opened 2021-09-07
Member expression bounds checking in multiple assignments

kkjeer:
This PR fixes an issue with bounds checking member expressions across multiple assignments.

Consider an example:

```
struct S {
  int len;
  _Array_ptr<int> p : count(len);
};

void f(struct S *s, _Array_ptr<int> arr : count(3)) {
  s->p = arr, s->len = 3;
}
```

After the assignment `s->p = arr`, the observed bounds of `s->p` are `bounds(arr, arr + 3)`. At the assignment `s->len = 3`, we synthesize the member expression `s->p` whose target bounds `bounds(s->p, s->p + s->len)` use the value of `s->len`. However, since `s->p` already has observed bounds recorded in `State.ObservedBounds`, we do not update the observed bounds of `s->p` to its target bounds. Therefore, the assignment to `s->len` does not modify the observed bounds of `s->p`. At the end of this statement, the observed bounds of `s->p` are `bounds(arr, arr + 3)` and we have equality between `s->p` and `arr` and between `s->len` and `3`. The observed bounds `bounds(arr, arr + 3)` imply the target bounds `bounds(s->p, s->p + s->len)`. No errors or warnings are emitted.

### DIFF
--- a/clang/test/CheckedC/inferred-bounds/bounds-context-members.c
+++ b/clang/test/CheckedC/inferred-bounds/bounds-context-members.c
@@ -8,14 +8,14 @@
 
 struct S {
   int len;
-  array_ptr<int> p : count(len); // expected-note 4 {{(expanded) declared bounds are 'bounds(s->p, s->p + s->len)'}}
+  array_ptr<int> p : count(len); // expected-note 5 {{(expanded) declared bounds are 'bounds(s->p, s->p + s->len)'}}
   int i;
   array_ptr<int> q : count(i); // expected-note 2 {{(expanded) declared bounds are 'bounds(s[3].q, s[3].q + s[3].i)'}} \
                                // expected-note {{(expanded) declared bounds are 'bounds(s[4].q, s[4].q + s[4].i)'}}
   array_ptr<int> r : count(i); // expected-note 2 {{(expanded) declared bounds are 'bounds(s[3].r, s[3].r + s[3].i)'}}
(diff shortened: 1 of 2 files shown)


sulekhark (approved):
LGTM! Thank you!

clang/test/CheckedC/inferred-bounds/bounds-context-members.c
```diff
@@ -360,6 +360,115 @@ void updated_source_bounds3(struct S *s) {
...
+  // CHECK-NEXT:   BinaryOperator {{.*}} '+'
+  // CHECK-NEXT:     ImplicitCastExpr {{.*}} <LValueToRValue>
+  // CHECK-NEXT:       DeclRefExpr {{.*}} 'arr'
+  // CHECK-NEXT:     IntegerLiteral {{.*}} 3
+  // CHECK-NEXT: }
+}
+
+void multiple_assignments2(struct S *s) {
+  // Observed bounds context after statement: { s->a => bounds(s->a - 1, (s->a - 1) + 2) }
+  s->a = s->b, s->a++; // expected-warning {{cannot prove declared bounds for 's->a' are valid after increment}} \
```
kkjeer, line 422: TODO: at the assignment `s->a++`, the observed bounds of the RHS `s->a + 1` of the assignment should be `bounds(s->b, s->b + 2)`, not `bounds(s->a, s->a + 2)`. This can be done in a follow-up PR.

Issue #1178, opened 2021-09-02
Missing checks in calls to variadic functions inside checked scope


mgrang:
I realize I may have waited a little too late to post this, but I tested this PR and found several holes in the new checking.  If you save the code below as `printf-checking-bugs.c` and then run `clang -o printf-checking-bugs printf-checking-bugs.c` and `./printf-checking-bugs percent_n`, etc., you can see the SEGV.  If you don't want to address the problems in this PR, I can copy the examples to a new issue.

```c
#pragma CHECKED_SCOPE on

#include <string.h>
#include <stdio.h>
#include <unistd.h>
#include <fcntl.h>

#define TEST(_name) else if (strcmp(test_name, #_name) == 0)

int main(int argc, _Nt_array_ptr<_Nt_array_ptr<char>> argv : count(argc)) {
  _Nt_array_ptr<char> test_name = argv[1];
  if (!test_name) {
    fprintf(stderr, "No test name specified\n");
    return 1;
  }

  if (0) {}
  TEST(percent_n) {
    // Missing check that %n argument is a _Ptr.
    int arr _Checked[1];
    printf("hello\n%n", arr + 123456789);
  }
  TEST(scanf_scalar) {
    // Missing check that _any_ scalar scanf argument is a _Ptr.
    int arr _Checked[1];
    sscanf("42", "%d", arr + 123456789);
  }
  TEST(printf_s_count) {
    // Missing check that printf %s argument has at least count(0).
    char buf _Nt_checked[1];
    printf("%s", buf + 123456789);
  }
  TEST(scanf_p) {
    // scanf reads an arbitrary _Ptr<void> via %p. The right solution here may
    // be to disallow _Ptr<void>
    // (https://github.com/microsoft/checkedc/issues/335). I couldn't find any
    // other way to exploit %p, but that doesn't mean there isn't any.
    _Ptr<void> q = 0;
    sscanf("0x1", "%p", &q);
    _Ptr<char> p = (_Ptr<char>)q;
    (*p)++;
  }
  TEST(scanf_s_overflow) {
    // scanf %s overflows the output buffer. I guess the compiler should require
    // the format string to specify a maximum width and check it against the
    // bounds of the argument?
    char field _Nt_checked[10];
    char input _Nt_checked[1000];
    memset(input, 'x', sizeof input - 1);
    sscanf(input, "%s", field);
  }
  else {
    fprintf(stderr, "Unknown test name\n");
    return 1;
  }

  return 0;
}
[...]

PR #1182, opened 2021-09-09
Add more safety checks to variadic function calls in checked scope

mgrang:
The checking of variadic functions in checked scope follows these rules:
1. All warnings issued by the -Wformat family of flags are errors in checked
scope.
2. No bounds checking of arguments to variadic functions like printf/scanf,
etc is done.

For printf-like functions:
3. %s is allowed only with arg type _Nt_array_ptr or _Nt_checked.
4. %p is allowed with any arg type.
5. %n is disallowed.
6. For all other format specifiers, only scalar arg types are allowed.

For scanf-like functions:
7. %s is disallowed.
8. All width modifiers to format specifiers are disallowed.
9. %p is disallowed
10. %n is disallowed
11. For all other format specifiers, only _Ptr arg types are allowed.

### DIFF
--- a/clang/lib/Sema/SemaChecking.cpp
+++ b/clang/lib/Sema/SemaChecking.cpp
@@ -7987,7 +7987,7 @@ void CheckFormatHandler::CheckVarargsInCheckedScope(
     //   _Nt_array_ptr<char> p : count(5);
     //   printf("%s", p + 1234);
     // }
-    // Issue https://github.com/microsoft/checkedc-clang/pull/1178 tracks this.
+    // Issue https://github.com/microsoft/checkedc-clang/issues/1178 tracks this.
 
     } else {
       if (ArgTy->isCheckedPointerType()) {
--- a/clang/lib/Sema/SemaChecking.cpp
+++ b/clang/lib/Sema/SemaChecking.cpp
@@ -7581,6 +7581,9 @@ public:
 
   void HandleNullChar(const char *nullCharacter) override;
 
+  // Note: IsFuncScanfLike = true means that the variadic function being called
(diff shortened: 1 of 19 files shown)


mattmccutchen-cci:
It looks like this PR closes off all the safety holes in #1178 except `printf_s_count` (passing an out-of-bounds `_Nt_array_ptr` to `printf` `%s`).  It seems that one inherently requires a bounds check.

sulekhark (approved):
LGTM! Thank you!

clang/lib/Sema/SemaChecking.cpp
```diff
@@ -7974,8 +7977,18 @@ void CheckFormatHandler::CheckVarargsInCheckedScope(
...
+    // specifiers other than %s.
+
+    // TODO: Currently, we do not handle the case where an out-of-bounds
+    // null-terminated array is passed as an argument to %s. The following code
+    // is allowed even though it might be a safety hole.
+    // _Checked {
+    //   _Nt_array_ptr<char> p : count(5);
+    //   printf("%s", p + 1234);
+    // }
+    // Issue https://github.com/microsoft/checkedc-clang/pull/1182 tracks this.
```
mattmccutchen-cci, line 7990: I think you meant:
```suggestion
    // Issue https://github.com/microsoft/checkedc-clang/issues/1178 tracks this.
```
It's unfortunate that there doesn't seem to be an easier way to reuse the normal code path for validating an argument to a `const char * : itype(_Nt_array_ptr<const char>)` parameter, which should do this for free.

kkjeer (approved):
LGTM. Thanks!

PR #1183, opened 2021-09-09
Get rvalue bounds for the value of lvalue expressions from the CheckingState

kkjeer:
This PR modifies the way that rvalue bounds are determined for the value of a member expression, pointer deference, or array subscript to be consistent with the way that rvalue bounds are determined for the value of a variable. The rvalue bounds for the value of a variable, member expression, pointer dereference, or array subscript expression `e` are determined by looking at the `ObservedBounds` map in the checking state. If the `AbstractSet` containing `e` is present in `ObservedBounds`, then those are the bounds for the value of `e`. Otherwise, the bounds for the value of `e` default to the lvalue target bounds of `e` (or the lvalue bounds of `e` if `e` is the subexpression of an `ArrayToPointerDecay` cast.

One consequence of this change is that the order of expressions used to get or create an `AbstractSet` may differ, which may affect the representative expression for the `AbstractSet`. For example, consider the assignment `*p = p[0] + 1`. The rvalue bounds of `p[0]` are determined before modifying the observed bounds of `*p`. This means that `p[0]` is the first expression that is used to create the `AbstractSet` that contains `*p` and `p[0]`, so `p[0]` is the representative expression for this `AbstractSet`.

This PR also includes a minor fix in CanonBounds where `ExtVectorElementExprs` are compared by address. This prevents an `llvm_unreachable` that would otherwise occur when creating an `AbstractSet` for a member expression that contains an `ExtVectorElementExpr`.

### DIFF
--- a/clang/lib/AST/CanonBounds.cpp
+++ b/clang/lib/AST/CanonBounds.cpp
@@ -472,6 +472,7 @@ Result Lexicographic::CompareExpr(const Expr *Arg1, const Expr *Arg2) const {
      // of Lexicographic. These expressions are compared by their addresses.
      // This comparison will be deterministic for one compiler run, but is
      // not guaranteed to be deterministic across compiler runs.
+     case Expr::ExtVectorElementExprClass:
      case Expr::InitListExprClass:
      case Expr::ImplicitValueInitExprClass: return CompareAddress(E1, E2);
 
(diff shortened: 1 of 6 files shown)


sulekhark (approved):
LGTM! Thank you!

mgrang (approved):
LGTM.

PR #1206, opened 2024-09-03
Fix some failing clang tests cases (round 1)

dtarditi:
This change fixes some failing test clang test cases.  It also includes some documentation changes from main related to the checkedc-llvm-project -> checkedc-clang change.

There are 3 types of fixes:

Fixed bad merges:
- AST/ast-dump-attr.m
- AST/ast-dump-c-attr.c
- AST/ast-dump-expr-json.c: Also updated expected output to reflect new run commands.
- AST/sourceranges.cpp
- Driver/openmp-offload.c
- Driver/openmp-offoad-gpu-c.

Array types now have 1 less space: 'int [4]' changes to 'int[4]', for example
- AST/ast-dump-expr.c
- AST/dump-bounds.c
- Checkedc/checked-scope-error-messages.c:

Clang produces additional warnings that need to be suppressed:
- CheckedC/checked-scope/null.c


souragc (approved):
These changes look good. I think we can merge them.

PR #1208, opened 2024-09-30
Fix additional failing test cases

dtarditi:
This change fixes additional failing test cases for the upgrade to clang 17 sources.  It also pulls in some fixes to the main branch for Window CI testing.

The changes include:
1. Use the driver marshalling infrastructure for setting various flags for dumping Checked C internal compiler information.
2. Fix some bad merges of test cases.  The bad merges happened merging the Checked C changes into the clang 17 sources.
3. Fix some test cases that broke because of a change in the formatting of array types.  There is no longer a space between the declarator name the dimension of the array.
4. Clang now produces some additional warnings and errors, some of which were conditions that Checked C was already complaining about.  

This significantly reduce the number of failing test cases for the clang-17 upgrade.  For a Window x64 debug build,`ninja checkedc-clang` produces the following results:
```
  Skipped          :    30
  Unsupported      :   237
  Passed           : 33364
  Expectedly Failed:    39
  Failed           :    68


souragc (approved):
Looks good.

PR #1210, opened 2024-10-03
Fix 3 bad source code merges.

dtarditi:
Fix some incorrect merges that were causing test failures:
- In 2 cases, statements were accidentally deleted.  Restore them.
- In SemaDecl.cpp, we kept some old code that should have been deleted. This caused processing of section pragmas and section information on declarations to be skipped on non-C++ files.

These were identified by looking at diffs with the original clang source code or debugging the handling failing test cases in the compiler.

This fixes 20 failing tests.  With this change, the test results from check-clang are:
```
Testing Time: 6139.36s
  Skipped          :    30
  Unsupported      :   237
  Passed           : 33385
  Expectedly Failed:    39
  Failed           :    48
```

### DIFF
--- a/clang/lib/AST/ItaniumMangle.cpp
+++ b/clang/lib/AST/ItaniumMangle.cpp
@@ -5144,6 +5144,7 @@ recurse:
     mangleType(USN->getTypeSourceInfo()->getType());
 
     Out << "E";
+    break;
   }
   case Expr::PackExprClass:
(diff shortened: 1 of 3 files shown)


souragc (approved):
Looks good.

Machiry (commented):
I don't understand why we are removing this?

PR #1213, opened 2024-12-04
Add new syntax support for generic structs

souragc:


souragc:
checkedc/checkedc#506

